PR #1, opened 2015-09-01
fix indentation

phomes:


bentiss:
works for me.

@phomes, in the future, please try to sign your patches. I know most projects do not require it, but I (we) really like to do it in our projects, especially when the patch will eventually have to go through a public mailing list.

Thanks again.

PR #7, opened 2015-09-13
hidraw: realloc to avoid parsing the report desc twice

phomes:
I mostly did this to avoid the duplicated debug messages.

Did we not use realloc to get the benefits (zeroing / abort on OOM) of zalloc? I could make a zrealloc if needed.

### DIFF
--- a/src/libratbag-hidraw.c
+++ b/src/libratbag-hidraw.c
@@ -73,10 +73,13 @@ ratbag_hidraw_parse_report_descriptor(struct ratbag_device *device)
 			unsigned report_id = 0;
 
 			for (j = 0; j < size; j++) {
+				hidraw->report_ids = realloc(hidraw->report_ids,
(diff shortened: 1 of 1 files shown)


bentiss:
I am not a big fan of this, mainly because having num_reportids realloc is IMO not a very good idea.
We could simply add a parameter to  ratbag_hidraw_parse_report_descriptor() to not output twice the debug messages if this bothers you (the current HID parsing should be fast enough to not be a problem having it repeated twice).

Actually, I think the debug level should be "raw", not "debug", but this code has been quickly written mostly to have a good looking demo next week.

phomes:
No problem. We don't even need an extra parameter as we can just check hidraw->report_ids for null and only log if it is not. Would that be okay?

bentiss:
Yep, works for me.
Feel free to push this, you have now write access to libratbag/libratbag.

PR #13, opened 2015-09-16
Small fixes

cvuchener:
hidpp_request_command always sent short report. I added variable and switch case to use the correct size. I also changed the size for the displayed expected header, mybe there should be a define for that.

I wrote the set current resolution to test it but stopped there as I did not know to use the write_resolution_dpi op. So it is not actually tested.

I think the two other commits are trivial enough.


whot:
decided to merge it, easier for now and the rest looked good anyway. having the abort() would be nice though

PR #21, opened 2015-09-17
Read colors from profiles in hidpp10

phomes:
Not doing anything useful with them yet but this documents 3 of the unknown bytes.
We also skipped one too many profiles.

### DIFF
--- a/src/hidpp10.c
+++ b/src/hidpp10.c
@@ -443,7 +443,7 @@ hidpp10_get_profile(struct hidpp10_device *dev, int8_t number, struct hidpp10_pr
 	struct hidpp10_profile profile;
 
 	/* FIXME: profile offset appears to be 3, 1 and 2 are garbage */
-	number += 3;
+	number += 2;
 
 	log_raw(dev->ratbag_device->ratbag, "Fetching profile %d\n", number);
 
--- a/src/hidpp10.c
+++ b/src/hidpp10.c
@@ -389,7 +389,10 @@ union _hidpp10_button_binding {
 _Static_assert(sizeof(union _hidpp10_button_binding) == 3, "Invalid size");
 
 struct _hidpp10_profile {
-	uint8_t unknown[4];
+	uint8_t red;
+	uint8_t green;
+	uint8_t blue;
+	uint8_t unknown;
 	struct _hidpp10_dpi_mode dpi_modes[PROFILE_NUM_DPI_MODES];
 	uint8_t angle_correction;
 	uint8_t default_dpi_mode;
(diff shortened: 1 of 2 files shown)


phomes:
@cvuchener it's for the G9. The LEDs on it can change color. When I change profile (via a hardware button on the mouse) it will show the configured color.

cvuchener:
I'd like to have some usb traces from starting LGS when the mouse is already plugged with different current profiles. Can you provide that?

phomes:
@cvuchener I would like to but I do not have a windows system where I can install USBPcap. Sorry. What I am doing is changing the colors per profile in LGS and checking with libratbag what changed. The three first bytes are exactly the values I set for the colors. Btw don't expect this to have an effect on most logitech mice as most don't have LED's that can change color.

I am poking my way through all the settings and the profile is definitely different from what you have documented for the g500. E.g. only 3 bytes per DPI mode and many other differences. I will send a patch to your repo with documentation for this profile type.

phomes:
@cvuchener I have created a pull request for your repo with my documentation.

It seems we need two different drivers for this then. @whot would separate versions of the hidd10 files be acceptable or should we add if's?

It seems that I am able to change the active profile from LGS. I will get a trace for you for that in the coming days.

whot:
keep it in the same driver for now with conditions please. hid++ is supposed to be a generic protocol, so there's a high chance that if things don't work the same across devices we're misinterpreting something. And when we figure out what's wrong it's easier to merge two conditions together than merging two separate files into one again.

cvuchener:
The protocol is the same, the data is not. Settings and profile structure will vary between devices.

bentiss:
Knowing how Logitech implemented HID++ 2.0 and from my memories from HID++ 1.0, I really think that there is a generic way of addressing this issue (without relying on the PID).

We also want to eventually make hidppxx.[ch] as a separate library (or set of sources that could be re-used in a different project), so I would rather have the hidppxx files knowing most of the logic behind the protocol, and if quirks are required, I'd like hidppxx knowing them. So that would lead to your first solution.

Hopefully, we should have access to more information soon, but right now, I can not help you much on the actual design until then.

phomes:
Unless LGS relies on knowing each device based on PID (or otherwise), then it would have to query the device when we plug it in. My impression was that @cvuchener already had logged all the traffic that LGS creates during plugin and that we did not find something interesting there.

Something like the feature set from 0x0001 in hidpp20 would be nice - or at least a protocol version that could somehow tell us what kind of layout to expect.

cvuchener:
> Something like the feature set from 0x0001 in hidpp20 would be nice - or at least a protocol version that could somehow tell us what kind of layout to expect.

According to hidpp20 spec, there is no versioning in hidpp10. But you can test features by trying to read them: if you can read 0x57 without error, the device has a LED color setting. But, although the G500 has a LED color setting (both with the command and in the profile), the actual LEDs cannot change color.

If I did not miss anything, after reading the usual USB and HID descriptor, LGS read the profile page (page 2), than check page 1 (only one read command, it does not read the whole page). It then start polling 0x0f, if something is wrong, it rewrites the memory. It also polls 0x63 to get the resolution and update the GUI.

Maybe you should check https://github.com/vsht/g500-control/ and more precisely lines [801](https://github.com/vsht/g500-control/blob/master/g500-control.c#L801) and [803](https://github.com/vsht/g500-control/blob/master/g500-control.c#L803) from g500-control.c. The first write some special data that I don't understand, and the second read what was written. The program replays what a old version of SetPoint did, so it is like reading half a trace.

PR #28, opened 2015-10-09
Use udev rules/hwdb to match mice to drivers and SVGs

phomes:
This adds udev rules/hwdb that sets RATBAG_DRIVER and RATBAG_SVG
properties to the event devices. We then use this property to select
the correct driver instead of checking the hardcoded list in the
drivers. Having this in a hwdb makes it easier for users to test
if drivers match their mouse with a system installed libratbag.

Each driver now has an "id" which is used to match the driver set
in RATBAG_DRIVER. A SVG can optionally be set as well.

All entries in the hwdb are currently without a name as that
information was not available in the hardcoded lists.

Note that this patch breaks the test-driver.


bentiss:
Thanks for the patch. I made a quick fix for make check to be happy so I'll just merge it and apply my patch on top.

PR #29, opened 2015-10-12
extra _safe() helper functions, start of the cleanup attribute

whot:


whot:
withdrawing, this one aborts when we call log_buffer with an empty header, snprintf_safe will write a zero-length string and triggers the abort

bentiss:
just a FYI, I still pulled the first commit which cleans up the README

PR #36, opened 2015-10-13
hidpp10 individual features/enable hid++ reporting cleanup

whot:
Fixes for the 0x00 and 0x01 requests. Doesn't add anything new, it just moves from register-specific #defines to a single enum each with all the values. Nicer for a potential caller.


bentiss:
Works for me (sorry to take so long to merge)

PR #38, opened 2015-10-15
hidpp10 as separate library

whot:
This branch disentangles libratbag from the hidpp10 code so we can use hidpp10 functions on their own. Useful for a bunch of debugging tools or, long-term, for providing things like pairing/unpairing tools, etc.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -47,3 +47,4 @@ test-context
 test-device
 test-*.log
 ratbag-command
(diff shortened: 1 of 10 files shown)


bentiss:
works for me

bentiss:
I just pushed the hidpp20 conversion too so that master is not in a weird state.

PR #39, opened 2015-10-16
Initial support of profiles for HID++ 2.0 devices

bentiss:
This implements a first parsing of the profile that are exported by HID++ 2.0 devices.

Far from being complete, but seems to not fail on the G502 and the G303.

This pull request also contains the same tools to dump the memory of the mouse for hidpp20 devices.


whot:
you'll need to upate the PR - the last two commits seem out of order from what you have in your tree atm (89656bb)

bentiss:
> you'll need to upate the PR - the last two commits seem out of order from what you have in your tree atm

nope. You need to update your local copy :)
I forced push the branch wip/hidpp20, and there is no reasons github would have made a PR on something on my local tree...

Anyway, I'll close the PR because I can not merge it in its current state.

Once I fixed the various points, any strong opinion if I directly push to master?

whot:
nah, feel free to push directly. but swap the last two commits, because 896 uses the function exported later in e5c. should be the other way round

bentiss:
> nah, feel free to push directly. but swap the last two commits, because 896 uses the function exported later in e5c. should be the other way round

That is weird. My local branch and the one on github have the right order, but the pull request is messed up... Anyway, one more github mystery.

PR #40, opened 2015-10-17
hidpp10: convert resolution according to sensor type

cvuchener:
Here is a proposition for handling the different resolution formats. It was tested with G500 and G500s only.

You may want to change the way the device info is retrieved (maybe in the base hidpp_device or in another function).

For invalid resolution values, hidpp10_sensor_to_dpi silently returns 0. Because of disabled dpi modes, it would output a lot of error messages if I logged errors there.

### DIFF
--- a/src/hidpp10.c
+++ b/src/hidpp10.c
@@ -37,6 +37,8 @@
 #include <string.h>
 #include <poll.h>
 #include <unistd.h>
+#include <linux/hidraw.h>
+#include <sys/ioctl.h>
 
 #include "hidpp10.h"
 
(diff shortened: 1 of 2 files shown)


bentiss:
We had a private discussion (sorry for not talking about this earlier) last Wednesday with Peter regarding the retrieval of the DPI values.
The outcome of it was that we would prefer to store this information in the HWDB. The rational being that it's trivial to add something in the hwdb and would allow to support new hidpp10 devices (if there are any left) without recompiling the library.

For some sensors, it's rather easy. We can just put the whole list of accepted values in a udev tag (RATBAG_HIDPP10_DPI_LIST) for instance. Then in drvhidpp10_probe, we retrieve this list, parse it and store it as a table in hidpp10_device. Then, when we have 0x8n, we just take the n-th element in this table. 

By looking at your patch I realize that some more recent sensors will be more of a pain to handle because of the number of values. HIDPP10_SENSOR_S9500 and HIDPP10_SENSOR_S9808 are the ones I am thinking about, and for them, we might want to encode this in the hwdb in a different way. This could be a combination of RATBAG_HIDPP10_DPI_MIN, RATBAG_HIDPP10_DPI_STEP, RATBAG_HIDPP10_DPI_MAX. Or in one udev property: RATBAG_HIDPP10_DPI_PARAMETERS, or even just a magic setting RATBAG_HIDPP10_DPI_TYPE which contains either HIDPP10_SENSOR_S9500 or HIDPP10_SENSOR_S9808.

Any thoughts?

cvuchener:
Because of the variety of resolution encoding, I think RATBAG_HIDPP10_DPI_TYPE would be the best for the driver. But the min/max/step/list would make more sense for the UI. Accepting any value and rounding it seems bad when the step is too wide like with the G5 (or even the G9). Also silently clamping values to the sensor range is a problem.

If the "step" value is for UI only, it may be different from the actual dpi step. The G500 step (400/17 ~ 24 dpi) is not very user-friendly. LGS use a 50 dpi step instead. Rounding with an accuracy of +/- 12 dpi looks fine to me, especially given that the actual sensor resolution change more than that from the type of surface the mouse is used on.

cvuchener:
I wonder if the A6090 is really the correct sensor for the G9. I found it on [wikipedia](https://en.wikipedia.org/wiki/List_of_Logitech_products#Gaming_mice) and [a forum](http://www.overclock.net/t/854100/gaming-mouse-sensor-list). But [the available data sheet](http://media.digikey.com/pdf/Data%20Sheets/Avago%20PDFs/ADNS-6090.pdf) (See register 0x09 "Resolution" on page 28) does not match the resolutions @phomes told me. Also that would be the only non modified Avago sensor used by Logitech (G9x/G500 use S9500 instead of A9500, ...).

whot:
there's two issues here that we need to fix: how to we export this to the next layer (i.e. what's the API that tells the caller which resolutions are supported) and the internal bit: what resolutions does the device actually support. so maybe we should start thinking about the external API first, then worry about the internal bits which we can change at any time anyway?

a general comment: if we have a mix/max with a step size, this should imo be a single property with a specific format, e.g. 50:8200@50 or so, not split across multiple properties. the latter would make the hwdb unwieldly quickly.

bentiss:
> so maybe we should start thinking about the external API first, then worry about the internal bits which we can change at any time anyway?

Though I agree there are indeed two problems, I think @cvuchener tried to solve the internal one in his pull request, and I was only commenting on this. The udev props "RATBAG_*" should not be used in any ways by the clients of the library.

The external API problem should be solved through an API call (ratbag_get_resolution_min|max|steps or any other API), but the internal one (what does 0x84 means for which device) should be IMO solved through a udev prop. We should support both the DPI list (the matching values are not linear in case of the S6006) and also the min/max/step to be able to feed the driver with the correct values while parsing the device.

> a general comment: if we have a mix/max with a step size, this should imo be a single property with a specific format, e.g. 50:8200@50 or so, not split across multiple properties. the latter would make the hwdb unwieldly quickly.

ACK for that

whot:
right, my argument here is that if we figure out how the external API should look like, it may give us a hint on how we want the internal bits hooked up. for example, depending on the external API it may be simpler to ignore the min/max/step option and just have an explicit list. or it may not be :)

bentiss:
Please see pull request #47 for the udev approach.

> if we figure out how the external API should look like, it may give us a hint on how we want the internal bits hooked up

No. We want to be able to parse whatever comes from the device, so we need to be able to understand every possible value. Thus, the udev rules are independent from the external API.

bentiss:
Pull request #47 has been merged, closing this one.

PR #41, opened 2015-10-18
tools: memory offset are in half bytes

phomes:
As I understand the offset for byte i must be given as i/2.
Also bump the limit of the loop so we read the full 512 bytes of
the page.

With this change a dump of the memory of the G9 matches the profile layout we documented in @cvuchener's repo.

### DIFF
--- a/tools/hidpp10-dump-page.c
+++ b/tools/hidpp10-dump-page.c
@@ -35,9 +35,9 @@ dump_page(struct hidpp10_device *dev, size_t page, size_t offset)
 	int rc = 0;
 	uint8_t bytes[16];
 
-	while (offset < 256) {
-		hidpp_log_info(&dev->base, "page 0x%02zx off 0x%02zx: ", page, offset);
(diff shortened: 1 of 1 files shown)


phomes:
I think that it would make sense if offset was divided in hidpp10_read_memory - but I don't see that in the current code. Do I miss something?

whot:
looks like a bug in the current code. But let's keep the hidpp10 API simple and accept a byte offset rather than a 2-byte offset and make hidpp10_read_memory do the right thing (including complaining if offset % 2 != 0). I think that's the most intuitive way for the API

phomes:
should I update the patch to do that and force push the PR branch - or close this and do a new PR?

cvuchener:
Sorry, I must have dreamed it.

whot:
@phomes close and re-push please, that keeps the archive a bit more logical. I think, anyway, no idea what github does when the PR source sha disappars.

for anything that's a trivial fix only (typo, etc.) force-pushing would be fine but here I think we should start a new one.

PR #42, opened 2015-10-18
Fix dumping memory pages

phomes:
Update the hidpp10_read_memory API to take a single byte offset.

Also increase the size of each page dump to the full 512 size.

This PR is a continuation of PR #41


whot:
double-checked with the documentation again and there's nothing about the offset being a multiple of 2. but i guess if the doc implicitly uses "word" instead of "byte" for the offset then it may be correct (or the doc is just wrong ;) And my g500s definitely has a 2 byte offset multiple here.

thanks, merged.

cvuchener:
Although length are in bytes, all offsets address 16 bit words. There is no way to address an unaligned byte. Macro items that may be the destination of a jump need padding because of that.

PR #47, opened 2015-10-23
hidpp10 udev resolutions

bentiss:
This PR enable per device specific dpi mapping.
I was able to test only the parsing of the udev properties, but not the actual use of the table (lack of compatible hidpp10 device currently).

Please test and report your thoughts.

### DIFF
--- a/TODO
+++ b/TODO
@@ -4,6 +4,8 @@ resolution without changing the DPI (looking at you Etekcity and Roccat).
 buttons need a couple of has_capability() functions to check which action
 types and actions are possible
 
+profiles should be allowed to be enabled/disabled
+
 light support:
 - on/off
 - color
--- a/hwdb/70-libratbag-mouse.hwdb
+++ b/hwdb/70-libratbag-mouse.hwdb
@@ -66,10 +66,12 @@ mouse:usb:v046dp1028:name:*:
 # G5
 mouse:usb:v046dpc041:name:*:
  RATBAG_DRIVER=hidpp10
+ RATBAG_HIDPP10_DPI_LIST=400;800;1600;2000
 
 # G5 2007
 mouse:usb:v046dpc049:name:*:
  RATBAG_DRIVER=hidpp10
+ RATBAG_HIDPP10_DPI_LIST=400;800;1600;2000
 
 # G7
 mouse:usb:v046dpc51a:name:*:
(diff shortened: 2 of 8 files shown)


cvuchener:
Not tested yet, but I see a problem with the value 0, it is not an acceptable value. With register 0x63, it means "no change" and thus cannot be written. It can be written in the profile (tested on G500), but the actual resolution is quite high (maybe the maximum). Anyway it should not be considered a valid resolution value.

RATBAG_HIDPP10_DPI for the S9500 mice (G9x/G500/G700?) looks complicated. Why did you choose to start at 12.8 unlike the S9808 mice that start at 0?

cvuchener:
>ratbag info: Device Logitech G500 might have wrong dpi settings. Please add RATBAG_HIDPP10_DPI_LIST or RATBAG_HIDPP10_DPI_LIST to the udev properties.

The error message repeats the same variable name twice.

```
  Profile 0
    Resolutions:
      0: 410x410dpi @ 1000Hz
      1: 810x810dpi @ 1000Hz (default)
      2: 2000x2000dpi @ 1000Hz
      3: 10x10dpi @ 1000Hz
      4: 10x10dpi @ 1000Hz
```

Actual profile is 400/800/2000/disabled/disabled. Your rounding is as good as any, no problem there. But disabled modes are not read correctly.

bentiss:
> value 0, it is not an acceptable value.

I allowed it for S6006, which 0x80 means 400 dpi. If there is a typo here, then we can change it (you are the one with the actual devices, so it's better if you do whatever works for all of them).

> RATBAG_HIDPP10_DPI for the S9500 mice (G9x/G500/G700?) looks complicated. Why did you choose to start at 12.8 unlike the S9808 mice that start at 0?

I made a linear regression between the 2 known values: 0x08 -> 200 dpi, and 0xf3 -> 5700 dpi

> The error message repeats the same variable name twice.

good catch, thanks.

> Actual profile is 400/800/2000/disabled/disabled. Your rounding is as good as any, no problem there. But disabled modes are not read correctly.

We might probable want to round to the 50 closest value. This way we will have 0 for the raw value 0x00.

cvuchener:
> I allowed it for S6006, which 0x80 means 400 dpi.

No problem there. 0x80 is not 0. I meant that `hidpp10_build_dpi_table_from_dpi_info` add 0 in the table. It should start at 1 (or even higher if you want to mimic LGS).

> you are the one with the actual devices

Only G5, G500 and G500s. I am not sure about the G9, there may have been a misunderstanding with @phomes, it would be more coherent for the values to start at 0x80 instead of 0x81. `g_hack.c` (I think it is not online any more, I should re-upload it) uses values from 0x80 to 0x87 for the G9, but the author is not sure about the actual dpi values.

> I made a linear regression between the 2 known values: 0x08 -> 200 dpi, and 0xf3 -> 5700 dpi

Even though 0xf3 is the highest value accepted by LGS, it uses 0xf2 for 5700 dpi (with the 17/400 ratio, it is the closest integer). I propose `0:5700@23.53` (or `0:5720@23.53` for accepting 0xf3). But I cannot really be sure that 0xf2 and 0xf3 are different, the sensor is not accurate enough to make a difference. That depends if you want to mimic LGS or just give a good approximation of the resolution, no need for accuracy then.

> We might probable want to round to the 50 closest value. This way we will have 0 for the raw value 0x00.

I think the disabled mode should be checked before converting the value. Tables built from lists won't have 0.

phomes:
Wrt the G9. 0x81 is the lowest value LGS will set. 0x81 is 200 dpi. 0x82 is 400 dpi. Etc. The max LGS will set is 0x90 which is 3200 dpi. 0x00 means that the dpi option is not used.

cvuchener:
> [0] can be written in the profile (tested on G500), but the actual resolution is quite high (maybe the maximum)
> [...]
> I think the disabled mode should be checked before converting the value.

I did a few more test about the zero values and disabled modes with the G500. It looks like enabled modes are the first modes (mode 0 is *always* enabled) until the first one with a 0 as X resolution. This means that:
- zeroes in mode 0 don't disable it
- zeroes as Y resolution don't disable the mode
- modes with non-zeroes resolution after a disabled mode are ignored

phomes:
Note btw that I did not actually test the patch. To read the profile from the G9 I need to use my branch https://github.com/phomes/libratbag/commits/wip/G9-profile and it has bit rotted.

bentiss:
OK, there were no specific comments in the git commits of this PR, so I just forced pushed after making the following changes:

- rounding is now to the 25th closest. This gives few duplicates in the table, but I am not sure we want to have a 10 precision
- fixed the error message
- fixed the hwdb file with 0:5700@23.53
- do not include 0 in the table of dpi, which means that a raw value of 0 will return a dpi of 0 (disabled).

We do not currently support disabling/enabling profiles (cf discussion with the 0 dpi meaning) so I just added this item in the todo list.

bentiss:
forced push after rebase and merged PR given that nobody complained

PR #48, opened 2015-10-25
Correct index check in ratbag-command

paller:
When initializing N profiles they are indexed 0 to N-1. However,
when setting the active profile to X the validation of X is a
check for X > N. Notice that N is invalid as the maximum index
is N-1.

Change the validation check of the new active profile to X >= N.

Signed-off-by: Palle Ravn <ravnzon@gmail.com>


bentiss:
Thanks!

PR #50, opened 2015-10-25
Add G7 hwdb entry

Tele42:
This is untested, but should match a G5's behavior.


bentiss:
works for me. Thanks!

PR #52, opened 2015-11-12
build: kill -no-install for installed binaries

jengelh:
With -no-install present in ratbag_command_LDFLAGS, the program would
get directly copied from .libs/ratbag-command to /usr/bin, retaining
the undesired rpath tag in the ELF file.

Signed-off-by: Jan Engelhardt <jengelh@inai.de>


whot:
thanks!

PR #54, opened 2015-12-11
Wip/hidpp10 write profiles

bentiss:
This should implement the writing of profiles for hid++ 1.0 devices. At least it works on my G700.

@whot , @cvuchener, @phomes, I'd be glad if you could give it a spin before I merge this for the G500(s). You might want first to stop at d7cc840 which should output the data that are about to be sent, but it won't send it. The next commit actually writes to the device, and the last one fills a missing feature (which closes an issue :-P ).

### DIFF
--- a/src/driver-hidpp10.c
+++ b/src/driver-hidpp10.c
@@ -170,7 +170,52 @@ static int
 hidpp10drv_write_button(struct ratbag_button *button,
 			const struct ratbag_button_action *action)
 {
-	return -ENOTSUP;
+	struct ratbag_device *device = button->profile->device;
+	struct hidpp10drv_data *drv_data = ratbag_get_drv_data(device);
+	struct hidpp10_device *hidpp10 = drv_data->dev;
+	struct hidpp10_profile profile;
+	uint8_t code;
+	int ret;
+
+	if (hidpp10->profile_type == HIDPP10_PROFILE_UNKNOWN)
+		return -ENOTSUP;
+
+	ret = hidpp10_get_profile(hidpp10, button->profile->index, &profile);
+	if (ret)
+		return ret;
+
+	switch (action->type) {
+	case RATBAG_BUTTON_ACTION_TYPE_BUTTON:
+		profile.buttons[button->index].button.type = PROFILE_BUTTON_TYPE_BUTTON;
+		profile.buttons[button->index].button.button = action->action.button;
+		break;
+	case RATBAG_BUTTON_ACTION_TYPE_KEY:
+		code = ratbag_hidraw_get_keyboard_usage_from_keycode(device, action->action.key.key);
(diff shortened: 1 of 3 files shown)


cvuchener:
```
$ sudo ./tools/ratbag-command resolution 2 dpi set 1600 /dev/input/event12
Error: Device 'Logitech G500' has no switchable resolution
```

Maybe because `hidpp10drv_has_capability` returns `(device->num_profiles > 1)`. A device with one profile should have switchable resolution.

bentiss:
> Maybe because hidpp10drv_has_capability returns (device->num_profiles > 1). A device with one profile should have switchable resolution.

oops. Sorry. I forgot that the G500 has only one profile by default. Changed locally on a test on the value of the profile type (hidpp10->profile_type != HIDPP10_PROFILE_UNKNOWN).

> Why force `PROFILE_TYPE_ADDRESS` to use an address from the profile directory? This command is only useful when you want to load a profile that is not in the directory (in RAM for example).

Oops again. I made a wrong assumption while reading the spec. Furthermore this spec mentions that doing a `PROFILE_TYPE_ADDRESS` disables the change profiles via index, which basically means that the profile directory is moot when in this mode.

I'll update locally.

> What i know of LGS G500 metadata:
Start right after the profile (78 bytes after the beginning of the page) with a 5 byte "LGS02" marker.
Then 29 bytes whose I don't know the meaning (they are zeros, maybe an empty profile name).
Then comes 11 strings (the first two buttons can not be assigned macros in LGS and thus they don't have names).
Each string is 17 wide characters (big endian) counting the null terminating character (so maximum length is 16). Despite using wide characters, LGS uses ISO-8859-1 encoding (or something similar), it may depend on my locale.

Great! That is very similar to what I have on the G700 (except the name of the profiles are 23 u16 long).
Thanks also for the tip on the "LGS02" marker.

I haven't checked if the encoding was ISO-8859-1, I just used plain ASCII on the G700 currently.

I think I'll push to master once I have cleaned up the `PROFILE_TYPE_ADDRESS` issue. Then, we can add the retrieval of the names of the G500 too.

cvuchener:
> Great! That is very similar to what I have on the G700 (except the name of the profiles are 23 u16 long).
Thanks also for the tip on the "LGS02" marker.

I miscounted for the unknown data between marker and macro names (I forgot to count a line of 16 bytes, its 45 in total), I edited my comment.

Keep in mind that this data is LGS specific, you should check the marker before reading names. Default profiles (see [floe's memory dump](http://floe.butterbrot.org/external/g700s_dump.zip), there are macros there instead of names) or SetPoint profiles do not have it.

bentiss:
Thanks for the testings. Master is merged and now support write for the profile :)

I created pull request #55 to try to handle the metadata in its own thread.

PR #55, opened 2015-12-14
Wip/hidpp10 metadata

bentiss:
OK, so this should properly parse the metadata for G700 and G500.

Note that if you used SetPoint and set Macros, we might override them until we have proper macro support...

@cvuchener: do you mind giving it a spin on the G500?

### DIFF
--- a/src/hidpp10.c
+++ b/src/hidpp10.c
@@ -571,6 +571,18 @@ union _hidpp10_button_binding {
 } __attribute__((packed));
 _Static_assert(sizeof(union _hidpp10_button_binding) == 3, "Invalid size");
 
+union _hidpp10_profile_metadata {
+	struct {
+		uint8_t marker[5];
+		uint8_t padding[420];
+	} __attribute__((packed)) any;
+	struct {
+		uint8_t marker[5]; /* { 'L', 'G', 'S', '0', '2' } */
+		uint16_t name[23];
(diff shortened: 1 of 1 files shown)


cvuchener:
```
ratbag raw: macro 0 of profile 0 is named: 'MIddle----------'
ratbag raw: macro 1 of profile 0 is named: 'Back'
ratbag raw: macro 2 of profile 0 is named: 'Forward'
ratbag raw: macro 3 of profile 0 is named: 'Cinqui�me-------'
ratbag raw: macro 4 of profile 0 is named: 'Tilt-left'
ratbag raw: macro 5 of profile 0 is named: 'Tilt-right'
ratbag raw: macro 6 of profile 0 is named: 'PLus'
ratbag raw: macro 7 of profile 0 is named: 'Minus'
```
I get that with `--verbose=raw info` and the profile from the other pull-request. Is there anything else to test? `info` does not show the macro yet ("is mapped to 'none'").

bentiss:
This seems like we the tests are good so far.

For the tests, I mostly dump the current profile, overwrite a button setting (which rewrite the profile) and dump the result. Then I compare that the names are still there :)

bentiss:
I went ahead and pushed the branch in master.
I also implemented a first parsing of the macros in both hidpp10 and hidpp20. Still some work to do to support jumps and conditionals.

PR #56, opened 2016-01-07
hidpp10: Parse G9 profile

phomes:
This allows to read but not write profiles from the G9.

The dpi mode sturct is a little different as the mouse only supports
symetrical dpi for x/y. Given the current naming a suffix of _sym
was the best I could come up with.

The G9 profile has some additional unknown bytes compared to the
others. Unfortunately the content of them varies for the 4 first
profiles and the last. So adding a default value for the like with
the G700 is not possible. For this reason I dropped implementing
the write part for now.

Something else to address is that this mouse only has 10 buttons
compared to the usual 13 that we have in PROFILE_NUM_BUTTONS. For
now I just hardcoded it in the profile and hidpp10_get_profile.


bentiss:
> The dpi mode sturct is a little different as the mouse only supports symetrical dpi for x/y. Given the current naming a suffix of _sym was the best I could come up with.

I think it might be better to rename the other with _dual than having this named _sym (or _single).
Anyway, not a big deal.

> The G9 profile has some additional unknown bytes compared to the others. Unfortunately the content of them varies for the 4 first profiles and the last. So adding a default value for the like with the G700 is not possible. For this reason I dropped implementing the write part for now.

Might be easier to do what I did for the G500: just read the current values, and overwrite only what you know. This way, you keep the current data.

> Something else to address is that this mouse only has 10 buttons compared to the usual 13 that we have in PROFILE_NUM_BUTTONS. For now I just hardcoded it in the profile and hidpp10_get_profile.

That's not very clean, but on the other hand... if the profile is that way... Could you do a follow up patch where you create a #define for it instead of having the plain value here and there?

I am pulling this in master. It seems pretty consistent an should not break any other hardware. Please send cleaning follow up patches :)

PR #57, opened 2016-01-12
Add G500 profile to G9x mice

NoXPhasma:


bentiss:
Considering this tested, and so merged.

PR #59, opened 2016-02-10
Remove has_capability driver stub

bentiss:
This driver stub feels wrong. We can infer most of the capabilities by looking into the device. If a device has more than one profile, it is obvious is supports switchable profiles.

Plus this allows to have a common capability check between the drivers.

Submitting as a PR for comments before the merge.

### DIFF
--- a/src/driver-etekcity.c
+++ b/src/driver-etekcity.c
@@ -251,24 +251,6 @@ etekcity_button_action_to_raw(const struct ratbag_button_action *action)
 }
 
 static int
-etekcity_has_capability(const struct ratbag_device *device,
-			enum ratbag_device_capability cap)
-{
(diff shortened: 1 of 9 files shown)


cvuchener:
This causes hidpp10 device with one profile to lose the `profile` capability.

Testing a G500 with one profile:
```
Capabilities: res btn-macros
```

Testing a G500 with two profiles:
```
Capabilities: res profile btn-key btn-macros
```

I expected `profile` to be missing but there seems to be a problem with `btn-key` too.

bentiss:
> This causes hidpp10 device with one profile to lose the profile capability.

Thanks, I pushed a quick fix for that specific case. I did not remember that we only exported the actually enabled profiles, not the available ones.

PR #64, opened 2016-03-02
Create a proper .so for liblur

whot:


whot:
nevermind, I just saw @bentiss already did the sym/pkgconfig file bits

PR #66, opened 2016-03-02
Svg fixes

whot:
This adds a README on the base requirements for the SVG file and adjusts the files to fix into the required canvas size, reduces to the two layers, etc.


bentiss:
merged as of f233b6b

PR #68, opened 2016-03-10
Add a QUERY_CONFIGURATION capability

whot:
Designates a device that can be queried for the current configuration. Some
devices lack this ability so we rely on the caller to upload a known
configuration to the device before we can really proceed.

Set by default, devices without that capability should be the exception, not
the norm.

This patch just adds the capability, actual handling for this needs to be
added when we add the first device without this capability.


bentiss:
works for me

Issue #69, opened 2016-03-10
Wrong button mapping for Logitech G500s


kleinph:
I played around with piper and noticed that some button mappings are wrong.

As you can see in the picture Button 6 and 7 are mapped as Wheel Left and Right and Button 8 and 9 for the Resolution Control. But these mappings are actually the other way round. Also shouldn't the arrows of Button 8 and 9 point left and right?

![bildschirmfoto von 2016-03-10 18-38-24](https://cloud.githubusercontent.com/assets/566413/13679263/2a4e09e4-e6f3-11e5-9440-4114b7be6819.png)

cvuchener:
You are right, and buttons 3 and 4 are inverted too.

whot:
fwiw, that's the old graphic, I applied a bit of polish since. But I didn't change the button mappings (and clearly didn't doublecheck them either). They're just inscape SVGs, can you send us a PR with the labels fixed?

cvuchener:
I sent PR #71, with fixed labels and rotated arrows. I hope I did not break anything.

bentiss:
closing this as fixed as of PR #71.

Please re-open if things are still not correct

PR #71, opened 2016-03-11
data: Fix G500s button labels and arrows

cvuchener:
Changed button labels with their proper numbers and rotated arrows for wheel tilt buttons.


bentiss:
The problem with inkscape is that it can't help itself to not shuffle things around...

For me, the changes looks OK (I don't have the G500s). So I'll apply it to fix bug #69.

PR #74, opened 2016-03-14
Rewrite the macro API 

whot:
Change to have macros as proper objects. This breaks the API but makes macros a lot easier/safer to handle.

Note this is on top of #73, so pulling this one here if you're happy with both avoid merge commit awkwardness.


bentiss:
I tested the Macro revamp on the G502, and it seemed to work fine.
I could not find any nitpicks on the other PR, so:

Taking the 3 pull requests here.

PR #75, opened 2016-03-14
Add Logitech G700s

r3pek:


cvuchener:
Are you sure the DPI setting should not be `RATBAG_HIDPP10_DPI=0:8200@50`? I think the G700s has the same sensor as the G500s (S9808 with 8200 dpi).

r3pek:
@cvuchener 
from what i could gather online they're the exact same mouse (G700{,s}) but I can be wrong.... :-/

PR #77, opened 2016-03-14
configure.ac: udev basedir should be relative to prefix

whot:


bentiss:
This went unnoticed because it's the way Fedora packages things. But you are right.

PR #79, opened 2016-03-16
Fix a couple of error messages

### LINKED ISSUE #78: error message when dealing with devices with multiple event nodes
reproducer here is the G500s, symptom is that this device has two event nodes and I'm getting a "no hidraw device found" on each ratbagd start. ratbag-command list lists both event nodes as supported and both show the same info. but: ratbag_open_hidraw_node() has a check so we don't open the same hidraw node twice for two devices. It returns -ENODEV which triggers the error message.

ratbag-command list papers over this because we close the first device before we look at the second one, thus never triggering that code path.

whot:
Fixes #78 and another error message I get a lot with my half-busted roccat kone

### DIFF
--- a/src/driver-etekcity.c
+++ b/src/driver-etekcity.c
@@ -665,8 +665,8 @@ etekcity_probe(struct ratbag_device *device)
 	if (active_idx < 0) {
 		log_error(device->ratbag,
 			  "Can't talk to the mouse: '%s' (%d)\n",
-			  strerror(-rc),
-			  rc);
+			  strerror(-active_idx),
+			  active_idx);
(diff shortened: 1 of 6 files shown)


bentiss:
Thanks for the PR. Could you elaborate on the "half busted roccat kone"?

whot:
the mouse doesn't always initialise correctly after plugging it in. it seems to happen more often when ratbagd is running but that may be confirmation bias, it's possible that we're triggering some race condition with the kernel driver. or the hardware is generally unhappy from having been dragged across too many countries ;)

PR #82, opened 2016-04-29
Wip/logitech auto detect

whot:
Adds the code required to automatically try hidpp20 and then hidpp10 for any Logitech device. This drops the need for having any supported device in the hwdb though especially for hidpp10 we still need extra tags for profiles, dpi ranges, etc.

But at least this way we bind the driver immediately.

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -19,7 +19,8 @@ libratbag_la_SOURCES =			\
 	libratbag-hidraw.h		\
 	libratbag-private.h		\
 	libratbag-test.c		\
-	libratbag-test.h
+	libratbag-test.h		\
+	usb-ids.h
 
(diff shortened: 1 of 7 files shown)


whot:
fwiw, tested with the m570 and the G500s (both hidpp10) and the t650 and the g303.

bentiss:
I merged the PR, but we have now an issue with the Logitech keyboards that are seen as "supported".

We need to see if requiring the ID_INPUT_MOUSE tag solves the issue without failing for gaming mice as they also export a keyboard as a secondary interface.

whot:
Should be fine, they both revert to the same hidraw node, right? so worst case we don't see one of the event nodes but the other one will still be visible. If anything that makes things better (ratbag-command list currently shows me two devices for the G500s)

PR #94, opened 2016-07-14
hidpp10: fix fetching of serial number

whot:
Was off by one, the message content only includes one byte of extra information
before the serial number. And add a BE 32-bit helper while we're at it.


bentiss:
Looks good to me.
Sorry for taking so long to merge

PR #95, opened 2016-08-30
Added G502 support and small fix

cvuchener:
Added the product ID for the new G502 (with rgb LEDs) in hwdb. I hope I guessed correctly the version from the already present G502 product ID.

The profile format seems to be the one already supported. I will test that more later.

While reading the code I also found a incoherent comparison, I added the obvious bit shift, it looks coherent with the value in my G502's description.


bentiss:
Thanks. You are missing the svg file for the new 502, but I can push a fix for it right after the merge.

PR #97, opened 2016-09-18
Add a configure option for valgrind

skitt:
On some platforms, valgrind is available but dysfunctional; it is
therefore useful to be able to explicitly disable valgrind even if it
is present.

Signed-off-by: Stephen Kitt <steve@sk2.org>

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -121,6 +121,11 @@ AC_ARG_ENABLE(tests,
 	      [build_tests="$enableval"],
 	      [build_tests="auto"])
 
+AC_ARG_ENABLE(valgrind,
+	      AS_HELP_STRING([--enable-valgrind], [Test with valgrind (default=auto)]),
+	      [use_valgrind="$enableval"],
+	      [use_valgrind="auto"])
+
 PKG_CHECK_MODULES(CHECK, [check >= 0.9.10], [HAVE_CHECK="yes"], [HAVE_CHECK="no"])
 
 if test "x$build_tests" = "xauto"; then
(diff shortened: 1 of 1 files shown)


whot:
what platform has a broken valgrind but still needs access to gaming mice?
plus, valgrind is only used for the tests anyway (and they're mostly broken/insufficient right now)

skitt:
On Sun, 18 Sep 2016 17:05:56 -0700, Peter Hutterer <notifications@github.com>
wrote:
> what platform has a broken valgrind

All the MIPSen currently, and ARMv7 (although there's a patch for that) ...

> but still needs access to gaming mice?

... and from a strictly user-based point of view, none of them need access to
gaming mice (one could argue about mips64el but no). I'm looking at this from
a package maintainer's perspective though, in Debian — and we try to build
everything on all our platforms.

> plus, valgrind is only used for the tests anyway (and they're mostly
> broken/insufficient right now)

I know, but I like to run upstream tests during builds, it helps shake bugs
out...

I don't mind carrying this in Debian only, I thought it good form to submit
the patch upstream but I know that means pushing the burden on you. There are
other ways of working around the issue but IMO they make the builds more
annoying (especially for local builds — I don't like having a build which
behaves differently depending on what's installed, and build-conflicts end up
being annoying).

Regards,

Stephen

whot:
hmm. I think it'd be better if you could carry this in debian. One glorious day in the future we'll have a working valgrind on those platforms and everything will be fine, but once we merge this option we'll have to keep it around forever. Given this is a bit of a niche case and we're already badly ETIME on libratbag anyway, i'd prefer not to have this one, sorry.

PR #99, opened 2016-09-30
Add a release script to generate the announce email

bentiss:
Copied from Xorg/util/modular, the license should not be an issue
here.

### DIFF
--- a/release.sh
+++ b/release.sh
@@ -1,0 +1,606 @@
+#!/bin/bash
+#
+#		Generate the announce template
+#
+# Completely copy/paste of Xorg/util/modular/release.sh:
+#
+# Copyright (c) 2009, Oracle and/or its affiliates. All rights reserved.
+#
+# Permission is hereby granted, free of charge, to any person obtaining a
+# copy of this software and associated documentation files (the "Software"),
+# to deal in the Software without restriction, including without limitation
+# the rights to use, copy, modify, merge, publish, distribute, sublicense,
+# and/or sell copies of the Software, and to permit persons to whom the
+# Software is furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice (including the next
+# paragraph) shall be included in all copies or substantial portions of the
+# Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.  IN NO EVENT SHALL
+# THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+# FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+# DEALINGS IN THE SOFTWARE.
+
+export LC_ALL=C
+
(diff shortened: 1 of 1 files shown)


release.sh
```diff
@@ -0,0 +1,660 @@
...
+#------------------------------------------------------------------------------
+#
+generate_announce()
+{
+    cat <<RELEASE
+Subject: [ANNOUNCE] $pkg_name $pkg_version
+To: $list_to
+Cc: $list_cc
+
+Nicknamed "my releaser forgot to named me".
```
whot: typo " to name me"

but I think you can just skip this line altogether

release.sh
```diff
@@ -0,0 +1,660 @@
...
+
+libratbag $tag_name is out.
+
+The list of interesting changes are:
+
+See the full log at the end if you are interested in the details.
+
+Cheers,
+Your releaser
```
whot: also likely to be forgotten, just skip it

release.sh
```diff
@@ -0,0 +1,660 @@
...
+
+    for tarball in $tarbz2 $targz $tarxz; do
+	cat <<RELEASE
+As always, no tar.gz release on a different server, but you can grab
+the tag directly from github:
+
+https://github.com/libratbag/libratbag/archive/$tag_name/$tarball
+MD5:  `$MD5SUM $tarball`
+SHA1: `$SHA1SUM $tarball`
+SHA256: `$SHA256SUM $tarball`
```
whot: this won't work iirc. the timestamps change, so I don't think the tarball has the same md5sum/sha as the one from github. can you verify that? I'd just go with something along the line of

"The libratbag project does not generate tarballs for releases, you can grab one directly from github:"

release.sh
```diff
@@ -0,0 +1,660 @@
...
+	if [ $? -eq 0 ]; then
+	    if [ "x$FORCE" = "xyes" ]; then
+		echo "Warning: overwriting released tarballs due to --force option."
+	    else
+		echo "Error: tarball $tar_name already exists. Use --force to overwrite."
+		cd $top_src
+		return 1
+	    fi
+	fi
+    done
```
whot: if we don't do tarballs, we can skip this

release.sh
```diff
@@ -0,0 +1,660 @@
...
+	# Check if the tag is pointing to the top commit
+	if [ x"$tagged_commit_sha" != x"$remote_top_commit_sha" ]; then
+	    echo "Error: the \"$tag_name\" already exists."
+	    echo "       this tag is not tagging the top commit."
+	    remote_top_commit_descr=`git log --oneline --max-count=1 $remote_top_commit_sha`
+	    echo "       the top commit is: \"$remote_top_commit_descr\""
+	    local_tag_commit_descr=`git log --oneline --max-count=1 $tagged_commit_sha`
+	    echo "       tag \"$tag_name\" is tagging some other commit: \"$local_tag_commit_descr\""
+	    cd $top_src
+	    return 1
```
whot: side-note: I'd like us to check for signed tags here but that's something better added to the xorg script

release.sh
```diff
@@ -0,0 +1,660 @@
...
+		echo "Info: updated jh moduleset: \"$JH_MODULESET\""
+	    else
+		echo "Info: skipping jh moduleset \"$JH_MODULESET\" update in dry-run mode."
+	    fi
+
+	    # $tar* may be unset, so simply loop through all of them and the
+	    # first one that is set updates the module file
+	    break
+	done
+    fi
```
whot: skip?

release.sh
bentiss: Sadly yes, they are timestamped. I kept around the test of downloading the link still, just in case something broke in the future :)

release.sh
bentiss: Yep, not something I'll do today though...

release.sh
bentiss: Why don't you want to keep around jhbuild :)

PR #101, opened 2016-10-11
New profile-oriented API

Lyude:
Pull request for introducing the new profile-oriented API in libratbag. More info in commit descriptions.

### DIFF
--- a/src/libratbag-private.h
+++ b/src/libratbag-private.h
@@ -131,6 +131,18 @@ struct ratbag_driver {
 	void (*remove)(struct ratbag_device *device);
 
 	/**
+	 * Callback called when the driver should write any profiles that
+	 * were modified back to the device.
+	 *
+	 * Both profile and button structs have a dirty variable that can
+	 * be used to tell whether or not they've actually changed since
+	 * the last commit. In order to reduce the amount of time
+	 * committing takes, drivers should use this information to avoid
+	 * writing back profiles and buttons that haven't actually changed.
+	 */
+	int (*commit)(struct ratbag_device *device);
+
+	/**
 	 * Callback called when a read profile is requested by the
 	 * caller of the library.
 	 *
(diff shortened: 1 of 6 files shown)


src/libratbag-private.h
```diff
@@ -131,6 +131,20 @@ struct ratbag_driver {
...
+	 * Callback called when the driver should write any profiles that
+	 * were modified back to the device.
+	 *
+	 * Both profile and button structs have a dirty variable that can
+	 * be used to tell whether or not they've actually changed since
+	 * the last commit. In order to reduce the amount of time
+	 * committing takes, drivers should use this information to avoid
+	 * writing back profiles and buttons that haven't actually changed.
+	 *
+	 * Drivers need not modify dirty, as ratbag will do this for you.
```
whot: skip the dirty bit explanation, it's an implementation detail and since drivers don't need to handle it it shouldn't be mentioned

src/libratbag-private.h
```diff
@@ -131,6 +131,20 @@ struct ratbag_driver {
...
+	 */
+	int (*commit_profiles)(struct ratbag_device *device);
```
whot: just commit here, I'd say

src/libratbag-private.h
```diff
@@ -142,8 +156,7 @@ struct ratbag_driver {
 	void (*read_profile)(struct ratbag_profile *profile, unsigned int index);
 
 	/**
-	 * Here, the driver should actually write the profile to the
-	 * device.
+	 * FIXME
```
whot: FIXME: what?? :)
unless it's obvious, always add what needs fixing

src/libratbag.h
```diff
@@ -540,6 +540,18 @@ ratbag_device_has_capability(const struct ratbag_device *device,
 /**
  * @ingroup device
  *
+ * Write any profile changes to the device. This may take a couple of seconds
+ * to finish due to how slow most gaming mice are.
```
whot: "Write any changes to the device. Depending on the device, this may take up to a couple of seconds."

src/libratbag.c
```diff
@@ -1241,13 +1315,15 @@ LIBRATBAG_EXPORT enum ratbag_error_code
 ratbag_button_set_macro(struct ratbag_button *button,
 			const struct ratbag_button_macro *macro)
 {
-	int rc;
+	if (!ratbag_device_has_capability(button->profile->device,
+					  RATBAG_DEVICE_CAP_BUTTON_MACROS))
+		return RATBAG_ERROR_CAPABILITY;
```
whot: split these ones into a separate patch please

PR #102, opened 2016-10-13
Important iconv fixes + test

Lyude:
Realized that the code I had written for gskill to make sure there was always a profile name wasn't working like I had thought it was… traced down the issue to being that the iconv helper was not working anywhere as well as I thought it was and figured in addition to fixing that it would be a good idea to write some basic tests to make sure it doesn't break in the future.

### DIFF
--- a/test/test-iconv-helper.c
+++ b/test/test-iconv-helper.c
@@ -1,0 +1,178 @@
+/*
+ * Copyright © 2016 Red Hat, Inc.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a
+ * copy of this software and associated documentation files (the "Software"),
+ * to deal in the Software without restriction, including without limitation
+ * the rights to use, copy, modify, merge, publish, distribute, sublicense,
+ * and/or sell copies of the Software, and to permit persons to whom the
+ * Software is furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice (including the next
+ * paragraph) shall be included in all copies or substantial portions of the
+ * Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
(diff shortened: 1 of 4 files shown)


test/test-iconv-helper.c
```diff
@@ -0,0 +1,105 @@
...
+}
+
+int main(void)
+{
+	int nfailed;
+	Suite *s;
+	SRunner *sr;
+
+	s = test_context_suite();
+        sr = srunner_create(s);
```
whot: indentation

test/test-iconv-helper.c
```diff
@@ -0,0 +1,105 @@
+/*
+ * Copyright © 2015 Red Hat, Inc.
```
whot: your calendar is running slow :P

test/test-iconv-helper.c
```diff
@@ -0,0 +1,105 @@
...
+	char output[10] = { 0 };
+	ssize_t rc;
+
+	rc = ratbag_utf8_to_enc(output, sizeof(output), "Enterprise UTF-16LE",
+				"%s", "Foo");
+
+	ck_assert_int_eq(rc, -EINVAL);
+}
+END_TEST
+
```
whot: I'm missing a test here for { 'F', '\0' } (i.e. a single-byte string) and tests for other odd-numbered byte strings

also, what happens when you have multiple consecutive '\0' in the source string?

whot:
so, I clearly don't know what I'm doing and assumed that the first "ok" applied only to that commit rather than the pull request. Please fix up the couple of nitpicks and re-submit, thanks

test/test-iconv-helper.c
Lyude: Sorry for the duplicate, still getting the hang of this thing…
If you're talking about something like

```c
ratbag_utf8_to_enc(output, sizeof(output), "UTF-16LE", "%s", "F\0\0\0oo");
```
nothing really interesting would happen. `vsnprintf()` would see the first `'\0'` and assume the entirety of the string is `"F"`

Lyude:
Are there still any pending review changes for this?

PR #104, opened 2016-10-21
Add driver for the Logitech G300/G300s

phomes:
Supports the common functionality exposed by libratbag:
- selecting profile 1-3 (profiles cannot be disabled)
- selecting resolution 1-4 (can also not be disabled)
- report frequency
- default resolution per profile
- button actions: mouse button numbers or key+modifier
  (modifiers are just a bit field and can easly be set
  in the driver, but not included as libratbag does not
  support it yet)
- LEDs are not yet hooked up but are just a table of 7
  different predifined colors.
- Seting active profile and resolution are both implemented
  but only profile activation is in use. The resolution
  activation code is commented until supported from libratbag
  itself.

The G300 has an unusual way of numbering buttons:
Left:1, Right:2, Middle: 3

The purpose of report type 0xF2 is not known.

### DIFF
--- a/hwdb/70-libratbag-mouse.hwdb
+++ b/hwdb/70-libratbag-mouse.hwdb
@@ -95,6 +95,10 @@ mouse:usb:v046dpc249:name:*:
  RATBAG_HIDPP10_DPI=0:5700@23.53
  RATBAG_HIDPP10_PROFILE=G500
 
+# G300
+mouse:usb:v046dpc246:name:*:
+ RATBAG_DRIVER=logitech_g300
+
 # G500
 mouse:usb:v046dpc068:name:*:
  RATBAG_DRIVER=hidpp10
--- a/src/driver-logitech-g300.c
+++ b/src/driver-logitech-g300.c
@@ -1,0 +1,494 @@
+/*
+ * Copyright © 2016 Thomas Hindoe Paaboel Andersen.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a
+ * copy of this software and associated documentation files (the "Software"),
+ * to deal in the Software without restriction, including without limitation
+ * the rights to use, copy, modify, merge, publish, distribute, sublicense,
+ * and/or sell copies of the Software, and to permit persons to whom the
+ * Software is furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice (including the next
+ * paragraph) shall be included in all copies or substantial portions of the
+ * Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.  IN NO EVENT SHALL
+ * THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+ * FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+ * DEALINGS IN THE SOFTWARE.
+ */
+
+#include "config.h"
+#include <assert.h>
+#include <errno.h>
+#include <libevdev/libevdev.h>
+#include <linux/input.h>
+#include <stdlib.h>
+#include <stdint.h>
+#include <string.h>
+
+#include "libratbag-private.h"
+#include "libratbag-hidraw.h"
+
+#define LOGITECH_G300_PROFILE_MAX			2
+#define LOGITECH_G300_BUTTON_MAX			8
+#define LOGITECH_G300_NUM_DPI				4
+
+#define LOGITECH_G300_REPORT_ID_SET_ACTIVE_PROFILE	0xF0
+#define LOGITECH_G300_REPORT_ID_GET_ACTIVE_PROFILE	0xF1
+#define LOGITECH_G300_REPORT_ID_SET_ACTIVE_RESOLUTION	0xF0
+#define LOGITECH_G300_REPORT_ID_PROFILE_0		0xF3
+#define LOGITECH_G300_REPORT_ID_PROFILE_1		0xF4
+#define LOGITECH_G300_REPORT_ID_PROFILE_2		0xF5
+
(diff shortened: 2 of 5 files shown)


phomes:
Note that the driver is added to the autodetection for logitech mice that support neither HID++ 1 or 2. Specifically it will be used for any mouse with a report ID 0xF1. This might make it useful for e.g. the G600 and other non-HID++ mice. However, it does specifically not work for the G400s that I have. That was the reason I did not give the driver a generic name. Let me know if I should rename the driver. 

I will see what I can do about getting a G600 to test - and perhaps extend the driver to support if relevant.

src/driver-logitech-g300.c
```diff
@@ -0,0 +1,492 @@
...
+		struct ratbag_button_action *action = &button->action;
+		struct logitech_g300_button *raw_button;
+
+		if (!button->dirty)
+			continue;
+
+		raw_button = &report->buttons[button->index];
+
+		raw_button->code = logitech_g300_button_action_to_raw(action);
+		raw_button->modifier = 0x00; //FIXME: add modifier when generally supported
```
bentiss: These would have to go through a macro actually.

src/libratbag.c
```diff
@@ -345,6 +345,9 @@ ratbag_driver_fallback_logitech(struct ratbag_device *device,
 	if (!rc)
 		rc = ratbag_test_driver(device, dev_id, "hidpp10", NULL);
 
+	if (!rc)
+		rc = ratbag_test_driver(device, dev_id, "logitech_g300", NULL);
+
```
bentiss: Not so sure we need to have autodetection if we need to put some profile information in the driver...

src/driver-logitech-g300.c
```diff
@@ -0,0 +1,492 @@
...
+	struct logitech_g300_data *drv_data = NULL;
+	int active_idx;
+
+	rc = ratbag_find_hidraw(device, logitech_g300_test_hidraw);
+	if (rc)
+		goto err;
+
+	drv_data = zalloc(sizeof(*drv_data));
+	ratbag_set_drv_data(device, drv_data);
+
```
bentiss: We should probably force the mouse here to use the on-board memory. Otherwaise we won't be able to configure the mouse.

src/driver-logitech-g300.c
phomes: Comment removed

src/driver-logitech-g300.c
phomes: Not sure I follow. The call to ratbag_device_init_profile will cause each profile to be read from the on-board memory.

src/libratbag.c
phomes: I have removed the autodetection

src/driver-logitech-g300.c
```diff
@@ -0,0 +1,492 @@
...
+	struct ratbag_button_action action;
+};
+
+static struct logitech_g300_button_mapping logitech_g300_button_mapping[] = {
+	/* 0x00 is either key or unassigned. Must be handled separatly  */
+	{ 0x01, BUTTON_ACTION_BUTTON(RATBAG_BUTTON_TYPE_LEFT) },
+	{ 0x02, BUTTON_ACTION_BUTTON(RATBAG_BUTTON_TYPE_RIGHT) },
+	{ 0x03, BUTTON_ACTION_BUTTON(RATBAG_BUTTON_TYPE_MIDDLE) },
+	{ 0x04, BUTTON_ACTION_BUTTON(RATBAG_BUTTON_TYPE_THUMB) },
+	{ 0x05, BUTTON_ACTION_BUTTON(RATBAG_BUTTON_TYPE_THUMB2) },
```
phomes: changed to plain numbers

hwdb/70-libratbag-mouse.hwdb
```diff
@@ -95,6 +95,11 @@ mouse:usb:v046dpc249:name:*:
  RATBAG_HIDPP10_DPI=0:5700@23.53
  RATBAG_HIDPP10_PROFILE=G500
 
+# G300
+mouse:usb:v046dpc246:name:*:
+ RATBAG_DRIVER=logitech_g300
+ RATBAG_HIDPP10_DPI_LIST=125;250;500;1000
```
phomes: oops. Left over from a previous approach. Removed

src/driver-logitech-g300.c
```diff
@@ -0,0 +1,492 @@
...
+	struct logitech_g300_profile_data *pdata;
+	struct logitech_g300_profile_report *profile_report;
+	struct logitech_g300_button *button_report;
+
+	pdata = &drv_data->profile_data[profile->index];
+	profile_report = &pdata->report;
+	button_report = &profile_report->buttons[button->index];
+
+	ratbag_button_enable_action_type(button, RATBAG_BUTTON_ACTION_TYPE_BUTTON);
+	/* Button 0 and 1 can only be mouse clicks */
```
phomes: Right. Removed the check

src/driver-logitech-g300.c
```diff
@@ -0,0 +1,492 @@
...
+	if (ret < 0)
+		return ret;
+
+	if (ret != sizeof(buf))
+		return -EIO;
+
+	switch (buf[1]) {
+	case 0x07: return 0;
+	case 0x02: return 1;
+	case 0x04: return 2;
```
phomes: You are right. It is the color. Which is strange since two profiles can use the same color.

phomes:
I pushed a commit that deals with most comments. I am still thinking about a good name for the driver. It is only relevant for the G300 and G600. I have a few other unsupported logitech mice but they will not be supported by this driver as they do not even use HID :(. What about logitech-hid as a name for this one?

The code to get the current profile now uses the LED color code. This is not ideal as two profiles can use the same color. I don't see any other way to get it though.

src/driver-logitech-g300.c
bentiss: I meant the mouse can be either in the hardware profile mode, or in the software profile mode (where there is no profiles and the key macros are generated by LGS). So here, it would be good to force the hardware profiles.

bentiss:
> I pushed a commit that deals with most comments. I am still thinking about a good name for the driver. It is only relevant for the G300 and G600. I have a few other unsupported logitech mice but they will not be supported by this driver as they do not even use HID :(. What about logitech-hid as a name for this one?

Nah. logitech-hid would also match HID++ which uses HID too. Just keep g300, we'll add support for g600 in it, and see how it goes for other devices if there is a need. People just need to know that G600 and G300 are somewhat equivalent.

> The code to get the current profile now uses the LED color code. This is not ideal as two profiles can use the same color. I don't see any other way to get it though.

I think it's a hardware limitation you can't defeat. IIRC, if you set 2 profiles with the same color in LGS, you are screwed too and you can't access the second profile ;) 

Please squash the update in the initial commit and we'll take this one directly.

phomes:
Thanks. I have squashed the two commits.

PR #112, opened 2016-11-11
Add the ability to disable profiles

Lyude:
This allows us to enable and disable profiles on mice which support it. As well, this includes the required modifications to support this with the G.Skill driver. This is pretty nice since we can get rid of the ugly hack of enabling all profiles that we've had to use in the driver.

### DIFF
--- a/src/driver-gskill.c
+++ b/src/driver-gskill.c
@@ -313,6 +313,14 @@ struct gskill_data {
 	struct gskill_profile_data profile_data[GSKILL_PROFILE_MAX];
 };
 
+static inline struct gskill_profile_data *
+profile_to_pdata(struct ratbag_profile *profile)
+{
+	struct gskill_data *drv_data = profile->device->drv_data;
+
+	return &drv_data->profile_data[profile->index];
+}
+
 static const struct ratbag_button_action *
 gskill_button_function_to_action(enum gskill_button_function_type type)
 {
@@ -1037,9 +1045,7 @@ static void
 gskill_read_resolutions(struct ratbag_profile *profile,
 			struct gskill_profile_report *report)
 {
-	struct gskill_data *drv_data = ratbag_get_drv_data(profile->device);
-	struct gskill_profile_data *pdata =
-		&drv_data->profile_data[profile->index];
+	struct gskill_profile_data *pdata = profile_to_pdata(profile);
 	struct ratbag_resolution *resolution;
 	int dpi_x, dpi_y, hz, i;
 
@@ -1089,7 +1095,7 @@ gskill_read_profile(struct ratbag_profile *profile, unsigned int index)
 {
 	struct ratbag_device *device = profile->device;
 	struct gskill_data *drv_data = ratbag_get_drv_data(device);
-	struct gskill_profile_data *pdata = &drv_data->profile_data[index];
+	struct gskill_profile_data *pdata = profile_to_pdata(profile);
 	struct gskill_profile_report *report = &pdata->report;
 	uint8_t checksum;
 	int rc, retries;
@@ -1144,9 +1150,7 @@ static int
 gskill_update_resolutions(struct ratbag_profile *profile)
 {
 	struct ratbag_device *device = profile->device;
-	struct gskill_data *drv_data = ratbag_get_drv_data(device);
-	struct gskill_profile_data *pdata =
-		&drv_data->profile_data[profile->index];
+	struct gskill_profile_data *pdata = profile_to_pdata(profile);
 	struct gskill_profile_report *report = &pdata->report;
 	int i;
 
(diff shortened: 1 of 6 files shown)


tools/ratbag-command.c
```diff
@@ -101,6 +101,8 @@ usage(void)
 	       "Profile Commands:\n"
 	       "  profile active get		Print the currently active profile\n"
 	       "  profile active set N		Set profile N as to the  active profile\n"
+	       "  profile enabled get           Check if profile is enabled\n"
+	       "  profile enabled set E         Set profile enabled state to E\n"
```
whot: needs to be updated in the manpage too, and I'm really undecided about this here. I'm not a fan of parse_bool down below, we should only accept one option instead of trying to guess what the user meant.

I think it's worth here to add a new parsing command so we can say "profile enable" and "profile disable". It's a bit more work but I think long-term it pays off and it makes the command more natural (and would also work for resolution enable/disable, etc.)

whot: it's worth noting here: ratbag-command is a **debugging tool** for us, not a proper commandline interface that should be used by others. So we have some leeway here, e.g. "profile enabled get" could return multiple values.
I'm open to other ideas though.

Lyude: Sounds fine by me

Lyude:
Ping: updated the pull

tools/ratbag-command.c
```diff
@@ -351,7 +364,8 @@ ratbag_cmd_info(const struct ratbag_cmd *cmd,
 		if (!profile)
 			continue;
 
-		printf("  Profile %d%s\n", i,
+		printf("  Profile %d (%sabled)%s\n", i,
+		       ratbag_profile_is_enabled(profile) ? "en" : "dis",
```
whot: urgh, I really don't like that, it makes it impossible to grep for the message for effectively no benefit. just make it ? "enabled" : "disabled"

tools/ratbag-command.c
```diff
@@ -232,6 +234,17 @@ ratbag_cmd_get_active_resolution(struct ratbag_profile *profile)
 	return NULL;
 }
 
+static int
+parse_bool(const char *in)
```
whot: this one is unused, now, isn't it?

src/libratbag.h
```diff
@@ -606,6 +617,33 @@ ratbag_profile_unref(struct ratbag_profile *profile);
...
+ * @param profile A previously initialized ratbag profile
+ * @param enable Whether to enable or disable the profile
+ *
+ * @return 0 on success or an error code otherwise
+ */
+enum ratbag_error_code
+ratbag_profile_set_enabled(struct ratbag_profile *profile, int enabled);
+
+/**
+ * @ingroup profile
```
whot: I think we have a empty line after the @ingroup, just check the rest please

src/libratbag.h
```diff
@@ -606,6 +616,33 @@ ratbag_profile_unref(struct ratbag_profile *profile);
...
+ * Enable/disable the ratbag profile. For this to work, the device must support
+ * @ref RATBAG_DEVICE_CAP_DISABLE_PROFILE.
+ *
+ * @param profile A previously initialized ratbag profile
+ * @param enable Whether to enable or disable the profile
+ *
+ * @return 0 on success or an error code otherwise
+ */
+enum ratbag_error_code
+ratbag_profile_set_enabled(struct ratbag_profile *profile, int enabled);
```
whot: int enabled -> bool enabled

src/libratbag.h
```diff
@@ -519,6 +520,16 @@ enum ratbag_device_capability {
 	 * libratbag can be used to query the device as normal.
 	 */
 	RATBAG_DEVICE_CAP_QUERY_CONFIGURATION,
+
+	/**
+	 * The device has the capability to disable and enable profiles. It
+	 * should be noted that while profiles are not immediately deleted
```
whot: s/It should be noted that//

src/libratbag.h
```diff
@@ -519,6 +520,16 @@ enum ratbag_device_capability {
...
+	 * after being disabled, it is not guaranteed that the device will
+	 * remember any disabled profiles the next time ratbag runs. As well,
```
whot: s/As well/Furthermore/

deferring to you as non-ESL here though ;)

src/libratbag.c
```diff
@@ -714,12 +715,31 @@ ratbag_device_get_profile(struct ratbag_device *device, unsigned int index)
 	return NULL;
 }
 
+LIBRATBAG_EXPORT enum ratbag_error_code
+ratbag_profile_set_enabled(struct ratbag_profile *profile, int enabled)
```
whot: int->bool

src/driver-gskill.c
```diff
@@ -499,6 +500,9 @@ gskill_get_profile_count(struct ratbag_device *device)
 		return rc;
 	}
 
+	log_debug(device->ratbag,
+		  "Profile count: %d\n", buf[3]);
```
whot: this looks like it should fit on one line (same with a few other log_debugs introduced here)

Lyude:
Updated again, sorry for all the accidentally leftover junk!

PR #113, opened 2016-11-25
G900, g502 Modify LEDs settings

amfern:
can get LED's current settings from profile.
stores LED's mode, color, and effect settings to profile

### DIFF
--- a/src/driver-hidpp20.c
+++ b/src/driver-hidpp20.c
@@ -51,7 +51,8 @@
 #define HIDPP_CAP_BUTTON_KEY_1b04			(1 << 2)
 #define HIDPP_CAP_BATTERY_LEVEL_1000			(1 << 3)
 #define HIDPP_CAP_KBD_REPROGRAMMABLE_KEYS_1b00		(1 << 4)
-#define HIDPP_CAP_ONBOARD_PROFILES_8100			(1 << 5)
+#define HIDPP_CAP_COLOR_LED_EFFECTS_8070			(1 << 5)
+#define HIDPP_CAP_ONBOARD_PROFILES_8100			(1 << 6)
 
 struct hidpp20drv_data {
 	struct hidpp20_device *dev;
@@ -61,6 +62,7 @@ struct hidpp20drv_data {
 	unsigned num_controls;
 	struct hidpp20_control_id *controls;
 	struct hidpp20_profiles *profiles;
+	struct hidpp20_color_led_zone_info *led_infos;
 
 	unsigned int num_profiles;
 	unsigned int num_resolutions;
@@ -226,6 +228,7 @@ hidpp20drv_read_button_8100(struct ratbag_button *button)
 		break;
 	default:
 		button->action.type = RATBAG_BUTTON_ACTION_TYPE_UNKNOWN;
+		break;
 	}
 
 	ratbag_button_enable_action_type(button, RATBAG_BUTTON_ACTION_TYPE_BUTTON);
@@ -239,6 +242,44 @@ hidpp20drv_read_button(struct ratbag_button *button)
 {
 	hidpp20drv_read_button_1b04(button);
 	hidpp20drv_read_button_8100(button);
+}
+
+static void
+hidpp20drv_read_led(struct ratbag_led *led)
+{
+	struct ratbag_device *device = led->profile->device;
+	struct hidpp20drv_data *drv_data = ratbag_get_drv_data(device);
+	struct hidpp20_profile *profile;
+	struct hidpp20_color_led_zone_info *led_info;
+	struct hidpp20_led *h_led;
+
+	if (!(drv_data->capabilities & HIDPP_CAP_ONBOARD_PROFILES_8100))
+		return;
+
+	led_info = &drv_data->led_infos[led->index];
+	profile = &drv_data->profiles->profiles[led->profile->index];
+	h_led = &profile->leds[led->index];
+
+	switch (h_led->mode) {
+	case HIDPP20_LED_ON:
+		led->mode = RATBAG_LED_ON;
+		break;
+	case HIDPP20_LED_CYCLE:
+		led->mode = RATBAG_LED_CYCLE;
+		break;
+	case HIDPP20_LED_BREATHING:
+		led->mode = RATBAG_LED_BREATHING;
+		break;
+	default:
+		led->mode = RATBAG_LED_OFF;
+		break;
+	}
+	led->type = hidpp20_8070_get_location_mapping(led_info->location);
+	led->color.red = h_led->color.red;
+	led->color.green = h_led->color.green;
+	led->color.blue = h_led->color.blue;
+	led->hz = h_led->rate;
+	led->brightness = h_led->brightness * 255.0 / 100;
 }
 
 static int
@@ -342,6 +383,52 @@ hidpp20drv_write_button(struct ratbag_button *button,
 		return hidpp20drv_write_button_1b04(button, action);
 
 	return -ENOTSUP;
+}
+
+static int
+hidpp20drv_write_led(struct ratbag_led *led,
+		     enum ratbag_led_mode mode,
+		     struct ratbag_color color, unsigned int hz,
+		     unsigned int brightness)
+{
+	struct ratbag_profile *profile = led->profile;
+	struct ratbag_device *device = profile->device;
+	struct hidpp20drv_data *drv_data = ratbag_get_drv_data(device);
+	struct hidpp20_profile *h_profile;
+	struct hidpp20_led *h_led;
+
+	if (!(drv_data->capabilities & HIDPP_CAP_COLOR_LED_EFFECTS_8070))
+		return -ENOTSUP;
+
+	h_profile = &drv_data->profiles->profiles[profile->index];
+
+	h_led = &(h_profile->leds[led->index]);
+
+	if (!h_led)
+		return -EINVAL;
+
+	switch (mode) {
+	case RATBAG_LED_ON:
+		h_led->mode = HIDPP20_LED_ON;
+		break;
+	case RATBAG_LED_CYCLE:
+		h_led->mode = HIDPP20_LED_CYCLE;
+		break;
+	case RATBAG_LED_BREATHING:
+		h_led->mode = HIDPP20_LED_BREATHING;
+		break;
+	default:
+		h_led->mode = HIDPP20_LED_OFF;
+		break;
+	}
+	h_led->color.red = color.red;
+	h_led->color.green = color.green;
+	h_led->color.blue = color.blue;
+	h_led->rate = hz;
+	h_led->brightness = brightness / 255.0 * 100;
+
+	return hidpp20_onboard_profiles_write(drv_data->dev, profile->index,
+					      drv_data->profiles);
 }
 
 static int
@@ -571,6 +658,29 @@ hidpp20drv_read_kbd_reprogrammable_key(struct ratbag_device *device)
 }
 
 static int
+hidpp20drv_read_color_leds(struct ratbag_device *device)
+{
+	struct hidpp20drv_data *drv_data = ratbag_get_drv_data(device);
+	int rc;
+
+	if (!(drv_data->capabilities & HIDPP_CAP_COLOR_LED_EFFECTS_8070))
+		return 0;
+
+	free(drv_data->led_infos);
+	drv_data->led_infos = NULL;
+	drv_data->num_leds = 0;
+
+	rc = hidpp20_color_led_effects_get_zone_infos(drv_data->dev, &drv_data->led_infos);
+
+	if (rc > 0) {
+		drv_data->num_leds = rc;
+		rc = 0;
+	}
+
+	return rc;
+}
+
+static int
 hidpp20drv_read_onboard_profile(struct ratbag_device *device, unsigned index)
 {
 	struct hidpp20drv_data *drv_data = ratbag_get_drv_data(device);
@@ -719,6 +829,14 @@ hidpp20drv_init_feature(struct ratbag_device *device, uint16_t feature)
 	}
 	case HIDPP_PAGE_COLOR_LED_EFFECTS: {
 		log_debug(ratbag, "device has color effects\n");
+		drv_data->capabilities |= HIDPP_CAP_COLOR_LED_EFFECTS_8070;
+
+		/* we read the profile once to get the correct number of
+		 * supported leds. */
+		if (hidpp20drv_read_color_leds(device))
+			return 0;
+
+		ratbag_device_set_capability(device, RATBAG_DEVICE_CAP_LED);
 		device->num_leds = drv_data->num_leds;
 		break;
 	}
@@ -889,6 +1007,8 @@ struct ratbag_driver hidpp20_driver = {
 	.write_profile = hidpp20drv_write_profile,
 	.set_active_profile = hidpp20drv_set_current_profile,
 	.read_button = hidpp20drv_read_button,
+	.read_led = hidpp20drv_read_led,
 	.write_button = hidpp20drv_write_button,
 	.write_resolution_dpi = hidpp20drv_write_resolution_dpi,
+	.write_led = hidpp20drv_write_led,
 };


src/hidpp20.h
```diff
@@ -317,6 +324,46 @@ union hidpp20_button_binding {
...
+	HIDPP20_LED_LOGO = 0,
+	HIDPP20_LED_SIDE
+};
+
+struct hidpp20_internal_led {
+	uint8_t mode;                  /* 00 - off; 01 - on; 03 - cycle; 0a - breath */
+	struct hidpp20_color color;
+	union {
+		struct {
+			uint8_t free[2];
```
whot: uint16_t please. Avoid pad arrays unless where there are no good datatypes. Even for the free[3] below please use a uint8_t and a uint16_t. Makes the size more obvious.

you could be using _Static_assert for these too if you want

src/hidpp20.h
```diff
@@ -317,6 +324,46 @@ union hidpp20_button_binding {
...
+
+// TODO: put the leds into array and access them through this enum
+enum hidpp20_led_type {
+	HIDPP20_LED_UNKNOWN = -1,
+	HIDPP20_LED_LOGO = 0,
+	HIDPP20_LED_SIDE
+};
+
+struct hidpp20_internal_led {
+	uint8_t mode;                  /* 00 - off; 01 - on; 03 - cycle; 0a - breath */
```
whot: see also the later comment: use a hidpp20 specific enum for the modes and convert from the public API ones. that way we don't leak implementation into the API

src/libratbag-private.h
```diff
@@ -215,6 +224,16 @@ struct ratbag_resolution{
 	uint32_t capabilities;
 };
 
+struct ratbag_led {
+	struct ratbag_profile *profile;
+	enum ratbag_led_type type;
+	enum ratbag_led_mode mode;  /**< off; on; cycle; breath  */
+	struct ratbag_color color; /**< led color  */
```
whot: heh, at least these two comments shouldn't be needed. Mode is explained by the enum and color is, well color :)

src/libratbag-private.h
```diff
@@ -215,6 +224,16 @@ struct ratbag_resolution{
...
+	unsigned int rate;	/**< rate of action 100 - 20000 */
+	unsigned int bright;	/**< brightness of the LED 1 - 100 */
```
whot: 0 isn't allowed?

I'm in two minds here: 0..255 would be a more machine natural range and matches some of the LED capabilities closer. 0..100% is a more human-natural range but has lower granularity. and if we use 0..100, it might be worth adding a typedef int percent_t to make the code more obvious about data ranges.

src/libratbag-private.h
```diff
@@ -229,6 +248,8 @@ struct ratbag_profile {
 		struct ratbag_resolution *modes;
 		unsigned int num_modes;
 	} resolution;
+	struct ratbag_led *logo_led;
+	struct ratbag_led *side_led;
```
whot: This related to what @bentiss said in the bug: it looks like this is a bit too specific and would have to be changed later when we have devices with more/other than logo/side LEDs. I'm ok with it for now since this backend implementation only and can be changed anytime. But I wonder: should we make this an array or linked list from the start?

src/libratbag.c
```diff
@@ -611,6 +611,10 @@ ratbag_create_profile(struct ratbag_device *device,
 					   sizeof(*profile->resolution.modes));
 	profile->resolution.num_modes = num_resolutions;
 	profile->is_enabled = true;
+	profile->logo_led = zalloc(sizeof(*profile->logo_led));
+	profile->side_led = zalloc(sizeof(*profile->side_led));
+  profile->logo_led->profile = profile;
+  profile->side_led->profile = profile;
```
whot: sooner or later we'll need an allocator function for LEDs, might as well get started with it :) see e.g. ratbag_create_button, ratbag_create_profile, etc. for help

src/libratbag.c
```diff
@@ -677,6 +681,9 @@ ratbag_profile_destroy(struct ratbag_profile *profile)
 
 	free(profile->resolution.modes);
 
+	free(profile->logo_led);
+	free(profile->side_led);
+
```
whot: make the profile ref-countable please and use unref instead of a direct free.

src/libratbag.c
```diff
@@ -1273,6 +1299,93 @@ ratbag_button_unref(struct ratbag_button *button)
...
+	/* 	return RATBAG_ERROR_CAPABILITY; */
+
+	led->color = color;
+	led->dirty = true;
+	led->profile->dirty = true;
+	return RATBAG_SUCCESS;
+}
+
+LIBRATBAG_EXPORT enum ratbag_error_code
+ratbag_led_set_rate(struct ratbag_led *led, unsigned int rate) {
```
whot: 'unsigned int hz' please

src/libratbag.c
```diff
@@ -1273,6 +1299,93 @@ ratbag_button_unref(struct ratbag_button *button)
...
+	struct ratbag_device *device = profile->device;
+
+	switch (type) {
+	case RATBAG_LED_TYPE_UNKNOWN:
+		return NULL;
+	case RATBAG_LED_TYPE_LOGO:
+		return profile->logo_led;
+	case RATBAG_LED_TYPE_SIDE:
+		return profile->side_led;
+	}
```
whot: yeah, looking at this now after the comment above: we can just make this an array and loop through it until we find the type.

src/libratbag.h
```diff
@@ -1125,6 +1141,12 @@ enum ratbag_button_type {
 	RATBAG_BUTTON_TYPE_PROFILE_DOWN,
 };
 
+struct ratbag_color {
+	unsigned short red;
+	unsigned short green;
+	unsigned short blue;
+};
```
whot: if that is an open struct (which I'm not opposed to) you can just define it in the declaration in the hunk above.

I wonder: why did you pick unsigned short? unsigned int seems the more natural choice here (or even uint32_t if we start switching to stdint for the API, which we should)

src/libratbag.h
```diff
@@ -1300,6 +1322,139 @@ enum ratbag_button_action_special
 ratbag_button_get_special(struct ratbag_button *button);
 
 /**
+ * @ingroup led
+ *
+ * LED mods.
```
whot: typo, but this would also need a bit more of an explanation for each mode. on/off is obvious, but cycle for example isn't clear.

src/libratbag.h
```diff
@@ -1300,6 +1322,139 @@ enum ratbag_button_action_special
...
+ */
+enum ratbag_led_mode {
+	RATBAG_LED_OFF = 0,
+	RATBAG_LED_ON = 1,
+	RATBAG_LED_CYCLE = 03,
+	RATBAG_LED_BREATH = 10
+};
+/**
```
whot: let's make this a normal sequential enum and not leak implementation detail. Means we need to convert types somewhere, but that's a mere switch statement - which we'd need in every driver that doesn't use hidpp20 constants anyway

src/libratbag.h
```diff
@@ -1300,6 +1322,139 @@ enum ratbag_button_action_special
...
+ * @see ratbag_led_set_bright
+ */
+int ratbag_led_get_bright(struct ratbag_led *led);
+/**
+ * @ingroup led
+ *
+ * this function returns the LED object
+ *
+ * @return The LED rate
+ *
```
whot: copy/paste error

src/libratbag.h
```diff
@@ -1300,6 +1322,139 @@ enum ratbag_button_action_special
...
+/**
+ * @ingroup led
+ *
+ * this function returns the LED effect rate.
+ *
+ * @return The LED rate
+ *
+ * @see ratbag_led_set_rate
+ */
+int ratbag_led_get_rate(struct ratbag_led *led);
```
whot: newline after int (same below)

src/libratbag.h
```diff
@@ -1300,6 +1322,139 @@ enum ratbag_button_action_special
...
+ *
+ * @see ratbag_led_get_rate
+ */
+enum ratbag_error_code
+ratbag_led_set_rate(struct ratbag_led *led, unsigned int rate);
+
+/**
+ * @ingroup led
+ *
+ * this function sets the LED bright.
```
whot: s/bright/brightness/ everywhere please. But one comment on the API documentation: it should be precise and describe what happens. Just saying "sets the LED brightness" isn't enough, there should be comments about allowed ranges, what is expected to happen on the device, etc. Please address this for all the new functions.

tools/ratbag-command.c
```diff
@@ -1435,6 +1455,292 @@ static const struct ratbag_cmd cmd_button = {
...
+	char* str, *endptr;
+	int bright, b;
+	int rc;
+
+	if (argc < 1)
+		return ERR_USAGE;
+
+	led = options->led;
+	str = argv[0];
+	bright = strtoul(str, &endptr, 10);
```
whot: If there's no real error checking, we should just used atoi, or copy the safe_atoi wrapper from libinput (which I wrote, so feel free to copy it :)

(that also applies to a bunch of other strtol already there)

tools/ratbag-command.c
```diff
@@ -1435,6 +1455,292 @@ static const struct ratbag_cmd cmd_button = {
...
+
+	if (argc < 1)
+		return ERR_USAGE;
+
+  led = options->led;
+	str = argv[0];
+
+	struct ratbag_color color;
+	char *p;
+	char red[2], green[2], blue[2];
```
whot: please don't mix declarations and code, it's permitted in C99 but we don't use it

tools/ratbag-command.c
```diff
@@ -1435,6 +1455,292 @@ static const struct ratbag_cmd cmd_button = {
...
+
+	memcpy(&red, &str[0], 2);
+	memcpy(&green, &str[2], 2);
+	memcpy(&blue, &str[4], 2);
```
whot: whoah, no. please use sscanf  here which makes the whole lot a lot easier. especially given that you have an out-of-bounds writ in the line directly below :)

tools/ratbag-command.c
```diff
@@ -1435,6 +1455,292 @@ static const struct ratbag_cmd cmd_button = {
...
+	enum ratbag_led_mode mode;
+
+	if (!strcmp(str, "off"))
+		mode = RATBAG_LED_OFF;
+	if (!strcmp(str, "on"))
+		mode = RATBAG_LED_ON;
+	if (!strcmp(str, "cycle"))
+		mode = RATBAG_LED_CYCLE;
+	if (!strcmp(str, "breath"))
+		mode = RATBAG_LED_BREATH;
```
whot: use streq here please

tools/ratbag-command.c
```diff
@@ -1435,6 +1455,292 @@ static const struct ratbag_cmd cmd_button = {
...
+
+	/* set default rate and brightness */
+	if (mode == RATBAG_LED_CYCLE || mode == RATBAG_LED_BREATH) {
+		ratbag_led_set_rate(led, 20);
+		ratbag_led_set_bright(led, 100);
+	}
+
+  /* set default color */
+	if (mode == RATBAG_LED_ON) {
```
whot: should be else if here or, preferably, a switch statement with fallthrough

tools/shared.c
```diff
@@ -48,6 +48,23 @@ udev_device_from_path(struct udev *udev, const char *path)
 	return udev_device;
 }
 
+const char *
+led_mode_to_str(enum ratbag_led_mode mode) {
+	const char *str = "UNKNOWN";
+	switch (mode) {
+	case RATBAG_LED_OFF:	str = "off"; break;
+	case RATBAG_LED_ON:	str = "on"; break;
```
whot: one line for this is ok, but please use the same for all entries

src/hidpp20.c
```diff
@@ -1667,8 +1663,8 @@ union hidpp20_internal_profile {
 			char txt[16 * 3];
 			uint8_t raw[16 * 3];
 		} name;
-		uint8_t logo_effect[11]; /* G303 only */
-		uint8_t side_effects[11]; /* G303 only */
+		struct hidpp20_internal_led logo_led; /* G303, g502, g900 only */
+		struct hidpp20_internal_led side_led; /* G303, g502, g900 only */
```
whot: these are currently unused, so if you split out the structs into a separate patch we could merge this directly without depending on the rest of the implementation. i.e. make a patch that adds the hidpp20_internal_led and the hidpp20 color and replace it here in the struct.

src/hidpp20.h
```diff
@@ -317,6 +324,46 @@ union hidpp20_button_binding {
...
+};
+_Static_assert(sizeof(struct hidpp20_internal_led) == 11, "Invalid size");
+
+struct hidpp20_led {
+	enum hidpp20_led_mode mode;
+	struct hidpp20_color color;
+	uint16_t rate;
+	uint8_t bright;
+} __attribute__((packed));
+
```
whot: generally, wherever you have a packed struct you should add a static assert to make sure it works as expected. the only reason to pack a struct is for it to go on the wire, so better to be doubly safe here.

I"m wondering why this one is packed though, it seems like a normal struct?

src/libratbag.h
```diff
@@ -1300,6 +1322,139 @@ enum ratbag_button_action_special
...
+ *
+ * this function returns the LED object
+ *
+ * @return The LED rate
+ *
+ * @see ratbag_led_set_rate
+ */
+struct ratbag_led*
+ratbag_profile_get_led(struct ratbag_profile *profile,
+                       enum ratbag_led_type type);
```
whot: see also my main comment: please change this to get the led by index and add a led_get_type() function.

src/driver-hidpp20.c
```diff
@@ -343,6 +343,41 @@ hidpp20drv_write_button(struct ratbag_button *button,
 	return -ENOTSUP;
 }
 
+static int hidpp20drv_write_led(struct ratbag_led *led,
+    enum ratbag_led_mode mode,
+    struct ratbag_color color, unsigned int rate,
+    unsigned int bright) {
```
amfern: do you use some formater?
maybe some style specification or name for this style?

src/driver-hidpp20.c
```diff
@@ -343,6 +343,41 @@ hidpp20drv_write_button(struct ratbag_button *button,
...
+	struct ratbag_profile *profile = led->profile;
+	struct ratbag_device *device = profile->device;
+	struct hidpp20drv_data *drv_data = ratbag_get_drv_data(device);
+	struct hidpp20_profile *h_profile;
+	struct hidpp20_led *h_led;
+
+	/* if (!(drv_data->capabilities & HIDPP_CAP_BUTTON_KEY_1b04)) */
+	/* 	return -ENTSUP; */
```
amfern: I haven't figured out what capabilities indicates support for LED, if any?
I will remove this comment but maybe you have suggestions where to look for LED capabilities?

bentiss: probably checking for the feature 0x8070 for color LED effects, and feature 0x1300 for non RGB leds.

src/driver-hidpp20.c
whot: I have a .vimdir file containing 'set noexpandtab shiftwidth=8 cinoptions=:0,+0,(0', that should get you most of the way here (if you're using vim, anyway)

src/hidpp20.h
```diff
@@ -317,6 +324,46 @@ union hidpp20_button_binding {
...
+	HIDPP20_LED_SIDE
+};
+
+struct hidpp20_internal_led {
+	uint8_t mode;                  /* 00 - off; 01 - on; 03 - cycle; 0a - breath */
+	struct hidpp20_color color;
+	union {
+		struct {
+			uint8_t free[2];
+			uint16_t rate;          /* 100 - 20000 */
```
cvuchener: I think it is milliseconds.

src/driver-hidpp20.c
whot: fwiw, I have an old branch wip/hidpp20-led-support here:
https://github.com/whot/libratbag/tree/wip/hidpp20-led-support
that may contain some other useful bits, though last I remember it wasn't working and then I got side-tracked.

amfern: logitech drivers sends between 16000 - 20000 values, 16000 being 1 sec cycle and 20000 being 20 sec cycle,
i have no idea what kind of unit it is.

src/hidpp20.h
amfern: logitech drivers sends between 16000 - 20000 values, 16000 being 1 sec cycle and 20000 being 20 sec cycle,
i have no idea what kind of unit it is.

src/libratbag-private.h
amfern: the brightness at the lowest level is 0 - 100, so its a percentage. so 0..255 won't fit here

src/libratbag.h
amfern: because short is all we need for 0..255.

src/hidpp20.c
amfern: created a separate pull request for the internal structures #114

src/libratbag.h
whot: remember, this is the public API and it's an open-coded struct so we can't hide it behind APIs. So short needs to be all we need for... forever? if we really only care about 0-255, then unsigned char would be even better, makes it more obvious that we're only using one byte.

bentiss: I'd rather go for unsigned int, as @whot, or even a double (normalized between 0 and 1.0). The reason being that in the same way we are seeing 10-bits per color channel capable monitor, I wouldn't be surprised if some HW manufacturer decide to have more color capable LEDs.

src/driver-hidpp20.c
```diff
@@ -620,6 +728,17 @@ hidpp20drv_read_profile_8100(struct ratbag_profile *profile, unsigned int index)
...
+	for (i = 0; i < profile->resolution.num_modes; i++) {
+		res = ratbag_resolution_init(profile, i,
+					     p->dpi[i], p->dpi[i], p->report_rate);
+
+		if (profile->is_active && res->dpi_x == dpi)
+			res->is_active = true;
+
+		if (i == p->default_dpi)
+			res->is_default = true;
+	}
```
whot: this looks like a copy/paste issue?

src/driver-hidpp20.c
```diff
@@ -344,6 +384,51 @@ hidpp20drv_write_button(struct ratbag_button *button,
...
+		h_led->mode = HIDPP20_LED_ON;
+		break;
+	case RATBAG_LED_CYCLE:
+		h_led->mode = HIDPP20_LED_CYCLE;
+		break;
+	case RATBAG_LED_BREATHING:
+		h_led->mode = HIDPP20_LED_BREATHING;
+		break;
+	default:
+		h_led->mode = HIDPP20_LED_OFF;
```
whot: add a break here too please (and all other default cases)

src/driver-hidpp20.c
```diff
@@ -240,6 +243,43 @@ hidpp20drv_read_button(struct ratbag_button *button)
...
+		led->mode = RATBAG_LED_ON;
+		break;
+	case HIDPP20_LED_CYCLE:
+		led->mode = RATBAG_LED_CYCLE;
+		break;
+	case HIDPP20_LED_BREATHING:
+		led->mode = RATBAG_LED_BREATHING;
+		break;
+	default:
+		led->mode = RATBAG_LED_OFF;
```
whot: add a break here too please

amfern:
@whot thanks for the hard work reviewing my commits, this is the last one for supporting my mices :)

src/driver-hidpp20.c
```diff
@@ -719,6 +829,14 @@ hidpp20drv_init_feature(struct ratbag_device *device, uint16_t feature)
 	}
 	case HIDPP_PAGE_COLOR_LED_EFFECTS: {
 		log_debug(ratbag, "device has color effects\n");
+		drv_data->capabilities |= HIDPP_CAP_COLOR_LED_EFFECTS_8070;
+
+		/* we read the profile once to get the correct number of
+		 * supported leds. */
+		if (hidpp20drv_read_color_leds(device))
+      return 0;
+
```
whot: fix the indentation here please. Wish there was a way to just do that online and merge it, but if there is I don't know how :)

PR #114, opened 2016-12-01
hidpp20: define internal LED structure

amfern:


src/hidpp20.h
```diff
@@ -317,6 +317,34 @@ union hidpp20_button_binding {
 } __attribute__((packed));
 _Static_assert(sizeof(union hidpp20_button_binding) == 4, "Invalid size");
 
+struct hidpp20_color {
+	uint8_t red;
+	uint8_t green;
+	uint8_t blue;
+};
```
whot: I think this one needs to be packed too? might not matter if we're only using it inside another packed struct, I'm not sure

src/hidpp20.h
```diff
@@ -317,6 +317,34 @@ union hidpp20_button_binding {
...
+			uint16_t rate;  /* 16000 - 20000Hz */
+			uint8_t brightness; /* 0 - 100 percent */
+			uint16_t pad_2;
+		} __attribute__((packed)) cycle;
+		struct hidpp20_led_breath {
+			uint16_t rate;
+			uint8_t pad_1;
+			uint8_t brightness;
+			uint16_t pad_2;
+			uint8_t pad_3;
```
whot: swap those two around please, for padding you should always go up, i.e. use uint8 to get to 16 bit, then uint16 to get to 32 bit, etc. It probably doesn't matter (and I've been tainted by the X protocol where we can't rely on packed) but it's still good form.

amfern:
@whot, I've changed the commit as you requested, can you review again?

PR #116, opened 2016-12-15
Internally export some HID++ constants for being able to reuse them in different projects

bentiss:
This is mostly just HIDPP_ prefixing and shuffling around some defines.

Looking at the fwupd code, it seems necessary to also export the HID++1.0 registers and HID++2.0 feature pages, but I not entirely sure about it given that now we have 3 places to put the protocol data: hidppX0.c, hidppX0.h, hidpp-constants.h.

But OTOH, I don't think we should put the individual commands in hidpp-constants.h.

Thoughts?


whot:
ok, the code itself looks good, but I haven't merged it, I'm just not that happy about the hidpp-constants.h itself. If we have a header that keeps both, just name it hidpp.h and move everything there. Otherwise, if we can separate the two enough, make a hidpp10.h and hidpp20.h (and move the internal ones) but tbh I think a single header is good enough here, so this just comes down to a rename now. I'd like the actual constants prefixed with hidpp10 and hidpp20 to avoid accidentally mixing them, with hidpp constants being truly generic across both versions. That's annoying to do and to type, but I think long term it should pay off. hopefully, maybe.

whot:
benjamin, what's the status of this? See my last comment

bentiss:
>benjamin, what's the status of this? See my last comment

I think I haven't pushed forward given that fwupd already took the files it needed from this PR. So it kind of fell into the nice to have requests.

whot:
let's close this for now until we have a specific request for it. If it happens as part of some other work, good, but no need to keep a PR open here.

PR #120, opened 2017-01-13
G700s special actions

cvuchener:
@11111010000 found more special actions code for the G700s with a different value for "resolution cycle up".

The first patch add resolution and profile cycle down definitions in libratbag and the second uses these definitions to fix and complete the special action list for HID++1.0.


whot:
hmm, does it matter that we have cycle up being cycle down? I wonder whether there's a need to define both (long-term answer is probably yes, but still). Do people/devices map buttons as cycle in two different directions?

cvuchener:
If you prefer, I could fix the values without adding the cycle down actions.

I am not sure about its usefulness, but since there was "cycle up", it felt right to add "cycle down". If you don't care about the direction, maybe it should simply be called "cycle".

While we are discussing this, I should add there is also a special action to switch to a specific profile by its index. It is useful on the G500 where it can replace the cycle action that does not work. I am not sure how it could be added because of the profile index (maybe by hard-coding the first n profiles, or simulating the profile cycle with it).

For more information, see [the updated documentation](https://github.com/cvuchener/g500/blob/master/doc/profile.md#0x83--special-g500-function).

PR #125, opened 2017-02-08
Teach ratbag-command the "ratchet mode switch" special action name

mxmgh:
This doesn't effect functionality (since assigning this action to buttons seems to do nothing on the MX Master at present), but the ratbag-command program should at least know the action's name.


whot:
the logitech page seems to call this button the "mode shift" button. maybe we should stick with that nomenclature?

http://support.logitech.com/en_nz/article/Configure-the-MX-Master-mouse-with-Logitech-Options?product=a0qi0000006Njj9AAC

bentiss:
I am not a big fan of a simple "mode shift". In the MX Master case, it's rather obvious what it is, but I am not sure everybody who don't own this particular mouse will understand that this is a reserved one for the ratchet feature. Plus on the drawing on the page you linked, the button is called "manual mode switch".

So I'd say if we want to expose it, stick to "ratchet mode switch", as in the PR.

whot:
oki, in that case - merged! thanks

PR #131, opened 2017-03-02
add SVG Logitech G303 data

alcomposer:
An SVG of the Logitech G303 mouse, can also pass as the G302 as very little difference between models

### DIFF
--- a/data/logitech-g303.svg
+++ b/data/logitech-g303.svg
@@ -1,0 +1,578 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<!-- Created with Inkscape (http://www.inkscape.org/) -->
+
+<svg
+   xmlns:dc="http://purl.org/dc/elements/1.1/"
+   xmlns:cc="http://creativecommons.org/ns#"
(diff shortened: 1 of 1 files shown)


alcomposer:
Ok, just realised that button 3 & 4 are lable wrong way around.

Will fix soon.

alcomposer:
Not on my computer right now, and can't install Git, can't this happen auto now? 

http://stackoverflow.com/questions/35948131/github-squash-commits-from-web-interface-on-pull-request-after-review-comments

Otherwise I can later tonight.

alcomposer:
Hey @bentiss, let me know if the squash is ok.

bentiss:
I don't think the squash you did is correct. You should have had only one commit on top of the common branch, not 2. I'll try the squash and merge feature of github on my side.

alcomposer:
Yeah, I think I did it right the first time, but then messed up something... Git is newish to me... :-(

bentiss:
No worries. We all need to learn the process at some point. The good thing is that the squash and merge feature (that was disable before by me) seems to do the job properly. So we are good now :)

I disabled it to prevent aggressive squash and merges, but it seems that in such cases it's a good feature.

PR #144, opened 2017-04-18
Switch to meson as build system, replace autotools

### LINKED ISSUE #142: Switch to meson as build system
The meson build system [1] is gaining a fair bit of traction, parts of GNOME are being switched over and we are currently discussing it for X.Org and Wayland. It'd remove automake, the big benefit is a great speedup in configure + build time (quick tests here are 20s vs 4s for a build from a fresh repo).

What's better though is that it's a lot more understandable, your average meson config makes a lot more sense than your average autotools setup.

I have a test branch that adds meson, if we're happy with this I think we should go ahead here. 

[1] https://github.com/mesonbuild/meson/

whot:
Fixes #142


whot:
unsurprisingly, the build fails because it's still using make which won't work anymore.

meson.build
```diff
@@ -0,0 +1,381 @@
...
+svg_files = [
+	'data/etekcity.svg',
+	'data/logitech-g300.svg',
+	'data/logitech-g303.svg',
+	'data/logitech-g500s.svg',
+	'data/logitech-g502.svg',
+	'data/logitech-g700.svg',
+	'data/logitech-g900.svg',
+	'data/logitech-mx_master.svg',
+	'data/logitech-mx-master.svg',
```
bentiss: This extra not upstream logitech-mx-master.svg prevents meson to succeed

README.md
```diff
@@ -67,6 +67,27 @@ Source
...
+
+libratbag uses the meson build system, see http://mesonbuild.com which in
+turn uses ninja to invoke the compiler. From a fresh git checkout, run the
+following commands to init the repository:
+
+    meson builddir --prefix=/usr/
+
+And to build or re-build after code-changes, run:
+
+    ninja -C builddir
```
bentiss: Fedora 25 here: s/ninja/ninja-build/
Weird :-/

README.md
```diff
@@ -67,6 +67,27 @@ Source
...
+    sudo ninja -C build install
```
bentiss: shouldn't this be s/build/builddir/

whot:
All three issues updated, I still have the semaphore bit to do though. (that requires pulling meson from pip or somesuch...)

bentiss:
One other thing that will require updating is the release.sh script.

whot:
Bit of trial and error, but after updating the build script to install meson from pip and ninja from git, it got there. Only change to the meson.build file is the switch to C99, not an unreasonable one in 2017 :)

That exposes a lot of warnings in the semaphore build (but a lot less locally, I wonder if that's different ninja versions?) but they can be fixed separately, imo.

whot:
ok, fixed by changing to gnu99 instead of c99. i'm going to bed now :)

PR #146, opened 2017-04-19
Fix warning: ‘x’ may be used uninitialized in this function

bentiss:
Few defects detected by compiling the rpms


src/libratbag-util.h
```diff
@@ -82,22 +82,24 @@ int list_empty(const struct list *list);
...
+	if (udev && *udev)
 		udev_unref(*udev);
 }
 
 static inline void
 cleanup_udev_device_unref(struct udev_device **udev_device) {
-	if (*udev_device)
+	if (udev_device && *udev_device)
 		udev_device_unref(*udev_device);
 }
```
whot: tbh, for the cleanup functions I'd rather have an assert() than a null-check, there is no code path where we can pass null here

bentiss: Yeah, I now realized that. I have dropped the libratbag-util.h changes, and added an other fix for hidpp20-dump-pages

bentiss:
Forced the merge, the code is now pretty small and without issues

PR #150, opened 2017-04-24
Update gitignore for meson

whot:
Drop all automake detritus, leave only the default build directory meson
suggests


bentiss:
Actually, I pulled this one too fast. The recommended build directory in the README is not `build`, but `builddir`...

PR #156, opened 2017-05-05
cleanup bits, and --version support

whot:
I'm rather partial about the git version number and instead switching to a plain "0.7.0" which can be parsed easily. Most of the ppl that need the git version would be able to just run git log anyway, so...


whot:
done, dropped the gitversion and changed the --version output to just the version number

PR #159, opened 2017-05-23
hwdb: add G403 entry

gicmo:
The G403 comes in a (optional) wireless version too; the dongle
that come with it has the product id c539, which is already
handled and also works (by the G900) entry.

Signed-off-by: Christian Kellner <christian@kellner.me>

### DIFF
--- a/hwdb/70-libratbag-mouse.hwdb
+++ b/hwdb/70-libratbag-mouse.hwdb
@@ -100,6 +100,11 @@ mouse:usb:v046dpc246:name:*:
  RATBAG_DRIVER=logitech_g300
  RATBAG_SVG=logitech-g300.svg
 
+# G403 over USB
+mouse:usb:v046dpc082:name:*:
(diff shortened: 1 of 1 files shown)


bentiss:
Couple of issues:
- we don't need to add the USB version given it's autodetected (unless you have a nice SVG to attach to it)
- we have an issue regarding the SVG for the wireless version: both mice (g900, g403) are now sharing the same drawing :(

gicmo:
@bentiss was planning on trying to come up with a picture; not sure how nice it will end up being ;)

Hjdskes:
There is an image for the G403 in #182.

gicmo:
Let me change the commit to reference the picture from #182 ;)

whot:
Note to self: review this again when #182 is merged and we have the svg

bentiss:
There is actually not much preventing us from pulling this one out, so do it :)

PR #161, opened 2017-05-23
tools: rename led mode from breathing to breath

gicmo:
The help calls the mode "breath" and the mode will also be reported
as such (via led_mode_to_str).

If you prefer it the other way around (rename other locations to "breathing"), let me know.

### DIFF
--- a/tools/ratbag-command.c
+++ b/tools/ratbag-command.c
@@ -157,7 +157,7 @@ usage(void)
 	       "  led N {COMMAND}\n"
 	       "                      get                                print the LED properties\n"
 	       "                      set {COMMAND}                      sets LED properties\n"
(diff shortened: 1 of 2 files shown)


whot:
tbh, I'd rather go the other way round and s/breath/breathing/. Reason is that "breathing" is fairly well understood (in, out, in, out, at least usually ;) whereas "breath" is fairly ambiguous - it doesn't describe a continuous activity. And it has bad connotations (e.g. bad breath). Mind swapping this around wherever it says breath instead of breathing?

gicmo:
@whot sure no problem; makes more sense for me too actually -- I was taking the docs as the reference point.

whot:
our docs can be wrong, don't hesitate to fix them were it makes sense

gicmo:
Branch/PR updated, btw; forgot to mention.

PR #162, opened 2017-05-24
Color led

bentiss:
Some various improvements of the LEDs.
Tested on the G900 and G403.


src/hidpp20.h
```diff
@@ -389,26 +382,46 @@ enum hidpp20_led_type {
...
+			uint8_t intensity; /* 1 - 100 percent, 0 means 100 */
 		} __attribute__((packed)) cycle;
 		struct hidpp20_led_breath {
-			uint16_t rate;
-			uint8_t pad_1;
-			uint8_t brightness;
-			uint8_t pad_2;
-			uint16_t pad_3;
+			struct hidpp20_color color;
+			uint16_t speed_or_period;
```
cvuchener: Same as cycle, this value is the period in milliseconds.

bentiss: Actually, the documentation says that it's either the period (in ms, as you say), or the "speed". The speed is a device-specific matching between the value and the actual period. I guess Ilia had such a device which explain why the default period in ratbag-command when you set the mode to cycle is 20.

We know if the device uses period or speed by looking at the output of getZoneEffectInfo().

For Color Cycling:
- if 0x0002 is set -> speed is used (so 8-bits)
- if 0x8000 is set -> the period is used.

For breathing (0x0a):
- 0x0002 -> speed
- 0x8000 -> period
(so same than cycle)

Thanks for the comment, I'll update the PR soonish.

cvuchener: Thanks for the clarification. I guess getZoneEffectInfo is function 2 and is not currently implemented. It looks coherent with what you said.

bentiss: oops, yes, it's not implemented :)

And yes, getZoneEffectInfo is function 2.

bentiss:
Amended the documentation commit, and rebased on top of master + #166

bentiss:
Rebased on top of master to remove the conflicts

PR #163, opened 2017-05-24
hidpp20drv: fix segfault with G502

bentiss:
When using the G502 without RGB LEDs, the feature 8070 is not here.
And this segfaults :)


amfern:
hehe, my bad.
btw where do you get g502 without leds?

bentiss:
> btw where do you get g502 without leds?

The first available G502 had a LED, not a RGB one. The second generation that came out recently supports the RGB protocol.

Thanks for the quick review

PR #168, opened 2017-05-28
Manpages

skitt:
This pull-request adds a lur-command manpage and fixes a few issues with the ratbag-command manpage.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -256,6 +256,14 @@ executable('lur-command',
 	install : true,
 )
 
+man_lur_command = configure_file (
+	input: 'tools/lur-command.man',
+	output: 'lur-command.1',
+	configuration: man_config,
+	install : true,
+	install_dir : join_paths(get_option('mandir'), 'man1')
+)
+
 #### hwdb ####
 udev_dir = get_option('udev-dir')
 if udev_dir == ''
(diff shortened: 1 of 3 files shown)


tools/lur-command.man
```diff
@@ -0,0 +1,49 @@
+.TH lur\-command 1 "Oct 26, 2016" lur\-command
```
whot: pls make this the same as the one from ratbag-command (adjust that one in a separate commit, where neceessary).

Specifically: instead of a hardcoded date please use @version@ and hook it into the build system so that gets handled correctly.

skitt: I’ve adjusted the lur-command manpage and added it to the build. I think the ratbag-command manpage is OK there; I could add ratbag-command again at the end but that doesn’t add much information (perhaps “libratbag” would be better as the “source manual” entry...).

meson.build
```diff
@@ -256,6 +256,14 @@ executable('lur-command',
 	install : true,
 )
 
+man_ratbag_command = configure_file (
```
Hjdskes: Probably a nitpick but shouldn't this be `man_lur_command`?

skitt: Oh dear, so it should, thanks for spotting that.

PR #182, opened 2017-06-13
Use toned-down device SVGs

Hjdskes:
During a discussion I had with [Jakub Steiner](http://jimmac.musichall.cz/) he suggested we use toned-down device SVGs instead. Both @whot and I like this approach, so let's discuss this with the rest! This is a first device he did for us, which is the G403 that I own so I can use it while working on Piper.

The mockups aren't entirely up-to-date (will fix that tonight or tomorrow, most likely) but [this](https://hjdskes.github.io/img/blog/gsoc-part-4/mousemap_toned_down.png) is where we're going (that's still proof of concept code that I'll extend in the next few days). As you can see, we will need to retrieve some coordinates from the SVG to place the labels. This is why I also updated the guidelines. I also made the canvas size smaller (simply took what Jakub gave me, it looks good) which noticeably increases performance when opening and resizing Piper.

### DIFF
--- a/data/gnome/README.md
+++ b/data/gnome/README.md
@@ -1,0 +1,58 @@
+libratbag gnome theme
+====================
+
+Requirements
+------------
+
+See the Logitech G403 for an example of a device with a small amount of buttons,
+and the Logitech G700 for an example of a device with a large amount of buttons.
+
+- Canvas size should be between 400x400 and 500x500 pixels.
+- Three layers in the final SVG:
+  1. a lower layer named "Device" with the device itself. Each button on the
+  device should have an id `buttonX`, with `X` being the number of the button
+  (so for button 0 the id would be `button0`). Similarly, each LED on the device
+  should have an id `ledX`.
+  2. a middle layer named "Buttons" with the button leaders (see below for
+  leaders).
+  3. an upper layer named "LEDs" with the LED leaders.
+- A leader line is a path that extends from the button or LED to the left or the
+  right of the device (see below). Each leader line requires the following:
+  - It should start with a 7x7 square placed on or close to the button or LED
+    that it maps with.
+  - From this square, a path should extend left or right (see below).
+  - Each path should end with a 1x1 pixel with identifier `buttonX-leader` (or
+    `ledX-leader`), where `X` is the number of the button (or LED) with
+    which the leader maps. For button 0, this would be `button0-leader`.
+  - All these elements should be grouped and given the identifier `buttonX-path`
+    (or `ledX-path` for LEDs)
+- Leader lines should have a vertical spacing of at least 40 pixels. When there
+  are several leader lines above and below each other, make the spacing between
+  them equal.
+- If the device's scroll wheel supports horizontal tilting, add two small arrows
+  left and right of the scroll wheel with the respective button identifiers (see
+  the Logitech G700 for an example). Do not cut the scroll wheel in half
+  vertically to map these buttons.
+- If there aren't too many buttons, preferably make the leaders point to the
+  right with the device itself placed on the left. If the buttons would extend
+  below or above the device, make some point to the left instead with the device
+  itself centered in the middle. In this case, half of the leaders should extend
+  to the left and the other half to the right.
+- When a leader points to the right, its 1x1 pixel should have a style property
+  `text-align:start`. When a leader points to the left, its 1x1 pixel should
+  have a style property `text-align:end`.
+- The canvas should be resized so that there is a 20px gap between the device
+  and the edge of the canvas and no gap between the 1x1 pixels and the canvas.
+
+Technique
+---------
+
+The simplest approach is to find a photo of the device and import it into
+inkscape. Put it on the lowest layer, create a new layer "Device" above it
+and start tracing the outlines and edges of the device. Fill in the shapes
+and your device should resemble the underlying photo. Delete the photo
+layer, add leaders in their respective layers and you're done.
+
+Make sure the image looks ''toned-down'' and not realistic. Do not use dark or
+bright colors.
+
--- a/data/gnome/logitech-g403.svg
+++ b/data/gnome/logitech-g403.svg
@@ -1,0 +1,380 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<!-- Created with Inkscape (http://www.inkscape.org/) -->
+
+<svg
+   xmlns:dc="http://purl.org/dc/elements/1.1/"
+   xmlns:cc="http://creativecommons.org/ns#"
+   xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#"
+   xmlns:svg="http://www.w3.org/2000/svg"
+   xmlns="http://www.w3.org/2000/svg"
+   xmlns:sodipodi="http://sodipodi.sourceforge.net/DTD/sodipodi-0.dtd"
+   xmlns:inkscape="http://www.inkscape.org/namespaces/inkscape"
+   width="409.2084"
+   height="411.271"
+   viewBox="0 0 409.2084 411.27099"
+   version="1.1"
+   id="svg8"
+   inkscape:version="0.92.1 r"
+   sodipodi:docname="logitech-g403.svg"
+   inkscape:export-filename="/home/jimmac/logitech-g403.png"
+   inkscape:export-xdpi="96"
+   inkscape:export-ydpi="96">
+  <defs
+     id="defs2" />
+  <sodipodi:namedview
+     id="base"
+     pagecolor="#ffffff"
+     bordercolor="#666666"
+     borderopacity="1.0"
+     inkscape:pageopacity="0.0"
+     inkscape:pageshadow="2"
+     inkscape:zoom="1.4142135"
+     inkscape:cx="223.25242"
+     inkscape:cy="156.77583"
+     inkscape:document-units="px"
+     inkscape:current-layer="LEDs"
+     inkscape:document-rotation="0"
+     showgrid="false"
+     units="px"
+     borderlayer="true"
+     inkscape:showpageshadow="false"
+     inkscape:pagecheckerboard="false"
+     showguides="true"
+     inkscape:window-width="1366"
+     inkscape:window-height="704"
+     inkscape:window-x="0"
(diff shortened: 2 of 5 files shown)


Hjdskes:
CC @raghukamath @whot @bentiss. Please CC more people that should be involved.

Hjdskes:
Also, Jakub offered to do three more devices for us. Are there any devices that are particularly important to get done?

gicmo:
@Hjdskes the image looks very nice, I like it a lot!

gicmo:
I have updated #159 to include the image.

bentiss:
Just a small comment on the SVG itself (which I like very much).

I am a little bit concerned the " G" is too close to the real Logitech logo. IIRC we put as a requirement to not have any copyrighted material, and even if the shape of the mouse has a lot of chances of being copyrighted, having the actual logo is a little bit more worrying.

Please change this into something else (circle maybe), and we can merge the PR.

Hjdskes:
Good point, I'll pass that on to Jakub.

Hjdskes:
Jakub also mentioned (after reading the requirements for the SVGs) that using photos to start isn't something to encourage. He said that it's best to let designers to this work, and he'll try to help out as much as he can.

We also discussed prelighting/highlighting the button in the SVG when the respective button is hovered/pressed in Piper (no capture mode yet to do this directly from the device). I'll see if I can make this work, from the SVG side it just requires that the buttons have unique identifiers next to the leaders so we can inject CSS into the SVG programmatically.

whot:
fwiw, I've asked @jimmac in email for the G502, G700 and the roccat kone xtd, all three have have special features that require some design input, once we have those we can probably get the others based on those as template.

raghukamath:
This is nice, I like it. 
In the [mockup](https://hjdskes.github.io/img/blog/gsoc-part-4/mousemap_toned_down.png) you linked, the text appears white, I believe it is a bit low on contrast and can be more readable is it is a bit darker in color.

whot:
@raghukamath fwiw, that's not the final text, we're trying to do away with the button labels altogether. So that white text won't stay there.

Hjdskes:
Look what we've got: https://github.com/gnome-design-team/gnome-mockups/tree/master/libratbag/assets

I'll update the SVG instructions to account for left and right leaders, and I'll start importing those SVGs as well.

bentiss:
> Look what we've got: https://github.com/gnome-design-team/gnome-mockups/tree/master/libratbag/assets

Nice!
But there are few errors on those SVGs (not easy to find out when you don't own the mice):
- G403: nothing to report
- G502:
  - there is a missing "sniper button" below the 2 left thumb buttons (see https://github.com/libratbag/libratbag/blob/master/data/logitech-g502.svg button 5)
  - the button between the wheel and the one in the middle (button 8) doesn't report anything, it's a mechanical switch
  - the wheel has both vertical and horizontal scrolling capabilities, horizontal "buttons" can be configured (but not the vertical ones)
- G700:
  - there are 2 missing "thumb buttons" below the 2 left thumb buttons (see https://github.com/libratbag/libratbag/blob/master/data/logitech-g700.svg button 3 4 5 and 6)
  - the button between the wheel and the one in the middle (button 10) doesn't report anything, it's a mechanical switch
  - the wheel has both vertical and horizontal scrolling capabilities, horizontal "buttons" can be configured (but not the vertical ones)
  - there is no LED on the mouse

Both the G502 and G700 have their button leaders on each side, so I guess this requirement should go away, or we should add a "preferably" in front of the paragraph.

Hjdskes:
> Both the G502 and G700 have their button leaders on each side, so I guess this requirement should go away, or we should add a "preferably" in front of the paragraph.

Yes, this is precisely what I meant when I said I'll update the SVG instructions :)

Hjdskes:
Devices left to tone down:

- [ ] etekcity
- [ ] logitech-g300
- [ ] logitech-g303
- [ ] logitech-g500s
- [ ] logitech-g502 (awaiting update to fix issues raised in https://github.com/libratbag/libratbag/pull/182#issuecomment-308467084)
- [ ] logitech-g700 (awaiting update to fix issues raised in https://github.com/libratbag/libratbag/pull/182#issuecomment-308467084)
- [ ] logitech-g9
- [ ] logitech-g900
- [ ] logitech-mx_master
- [ ] roccat-kone-xtd

jimmac:
Hopefully I've addressed the issues raised about the model g700 and g502. I've also added the scrollwheel vertical directions as assignable events as per Jente's request.

Hjdskes:
Very cool, thanks!

> I've also added the scrollwheel vertical directions as assignable events as per Jente's request.

These seem to be missing from the G403 still (not sure if those are still planned, apologies if they are)

jimmac:
Sorry Sparkleshare was stuck pushing the local changes. Even the 403 should have the scrollwheel fixed now.

Hjdskes:
@bentiss Are you aware of any devices that do not have a static button assigned to switching resolutions? With "static" I mean that it's not exposed to the outside and hence libratbag cannot remap it, as on e.g. the G403 and G500.

This is relevant for fa8194ae6dacb5d0011e2b6527cbf9908be059f1. For Piper's resolutions page, I need to highlight which button(s) is(are) assigned to switch resolutions. For the G403 and G500 this is, as mentioned, not exposed through libratbag and hence we have to set this statically in the SVG. Of course, this will fail to work for devices where this *is* remappable, if there are any?

Hjdskes:
(Replying here to keep everything in this thread)

> Not sure about the G500, but the resolution switch button (under the wheel) on the G403 is definitively addressable and can be remapped.

Hm... not sure where I got the idea that this is static then.. @whot thought his G500 was static as well. In that case, we'll probably have to deal with both static and not static buttons. This will in both cases require the fourth layer but without the label in the SVG. I'll have to figure out how to draw a label for static buttons, but that should be do-able. That solves this issue, then.

> What could be interesting is to mention that any other drawings need to be drawn in one or more separate layers, that may (or not) be drawn by the UI.

Sure, I can add this. It shouldn't be in the way for Piper I think.

bentiss:
> @bentiss Are you aware of any devices that do not have a static button assigned to switching resolutions? With "static" I mean that it's not exposed to the outside and hence libratbag cannot remap it, as on e.g. the G403 and G500.

After reading this, I believe you might be mixing up the mice models and buttons:
- the G700, G900 and the G502 in this PR have all a mechanical switch below the wheel that enable/disable the ratchet mode on the wheel. When ratchet is disabled, the wheel is in the "free wheel" mode, where it just doesn't have any click (and a lot of inertia).
- all buttons on the Logitech devices I have are addressable and can be remapped in libratbag.

So IMO, the ratchet mode buttons that are just mechanical should just be grayed out in the SVG and nothing should mention it.

Also, there is one thing we need to do on the SVG (no need to ask jimmac for that), is that all the Logitech devices can't remap the vertical wheel events. So there are a couple of leaders/buttons on the drawings that need to be removed (and we need to rearrange the other leaders).

cvuchener:
Every electrical button can be remapped on the G500/G500s. The mechanical switch for ratchet mode is the one under the wheel. Also, as bentiss said, vertical wheel does not count as buttons, but horizontal wheel does.

Hjdskes:
I just updated all three toned down SVGs we currently have according to the new guidelines. The only device that isn't entirely done is the G700; my Inkscape skills aren't good enough to separate a button0 and a button1 from the rest of the device such that these can be given their identifiers.

cvuchener:
I made an [attempt at a G500/G500s SVG](https://github.com/cvuchener/libratbag/blob/toned-down-svgs/data/logitech-g500.svg) copying the G700 style. I have a few remarks regarding the instructions:

- There is no size for the image but margins and spacing are given in pixels. Different images will be scaled differently resulting in different margins, spacing, border size, ...
- The two main buttons on the G500 and G700 do not have separate pieces, so there is no corresponding object. Mouse wheel tilt is not drawn either. Is the "buttonX" id mandatory for every button? What should be done for this kind of buttons.

Regarding the G500, I have drawn the LEDs but not the leaders. These LEDs cannot be configured with effects, but they can be switched on and off depending on the resolution (and profile). Should they be ignored, or should there be 4 LED leaders?

whot:
@cvuchener I think for the G500s we should just ignore the LEDs, they're such a minor thing that the effort of exposing the configuration is probably not worth it.

as for 
> The two main buttons on the G500 and G700 do not have separate pieces, so there is no corresponding object.

Do you refer to it being the same piece of plastic? For the highlight, just make something up then I'd say.

Hjdskes:
> I made an [attempt at a G500/G500s SVG](https://github.com/cvuchener/libratbag/blob/toned-down-svgs/data/logitech-g500.svg) copying the G700 style. I have a few remarks regarding the instructions:

Looks good!

> - There is no size for the image but margins and spacing are given in pixels. Different images will be scaled differently resulting in different margins, spacing, border size, ...

Right, good point. Up until now we haven't cared about size too much, since it's unlikely that a user will have two devices to notice the window resizing when (s)he switches device in Piper. Let's add a minimum of 400\*400 and a maximum of 500\*500.

> - The two main buttons on the G500 and G700 do not have separate pieces, so there is no corresponding object. Mouse wheel tilt is not drawn either. Is the "buttonX" id mandatory for every button? What should be done for this kind of buttons.

Mouse wheel tilt should be drawn in the latest G502 and G700; it's the little arrows pointing left and right. Copy those and place the marker squares somewhere close to it, and give the arrows the button ids.

The button id is mandatory insofar that without it, the button cannot be highlighted (like [this](https://hjdskes.github.io/img/blog/gsoc-part-6/mousemap.webm)). I agree with @whot in that you can just create an object that overlaps with part of the button (e.g. the top) so it's clear what element is being configured. This also has to be done for the G700 SVG still.

Hjdskes:
This should be merged after #212. With the new theme support in libratbag, I suppose we can go ahead and merge the three devices that currently have an SVG. If so, I'll move the list from https://github.com/libratbag/libratbag/pull/182#issuecomment-308504676 to an issue instead.

Do note that the G700 still has no buttons in the SVG for button0 and button1 (my Inkscape isn't good enough to fix that)

bentiss:
@Hjdskes, can you quickly rebase the PR on top of master? It seems to conflict.
Once rebased, feel free to merge the PR.

Issue #180, opened 2017-06-12
Export the min and max DPI resolutions


Hjdskes:
This is required to implement the configuration in Piper. The exact design of this isn't yet decided, but whether we go with a slider or a spinbutton we'll need to know the min and max values.

bentiss:
Should be good now

PR #183, opened 2017-06-22
Add API to get the min/max resolution values for a device

whot:
Implemented for HID++ only, not for the other devices - wasn't obvious from the code how it works at first.

https://github.com/libratbag/libratbag/issues/180


Hjdskes:
I haven't tested it, but it looks like it does what Piper needs it to do.

whot:
can I get a quick review and ack for this one please? Thanks

Issue #179, opened 2017-06-12
merge ratabgd into libratbag


bentiss:
Currently, the only user of libratbag is ratbagd (minus ratbag-command).
Given libratbag requires root access to the hidraw nodes, and given that the devices that do not store their config in the device itself (some mice with CAP_QUERY_CONFIGURATION and the G910 keyboard for instance), we might as well force anybody else to use ratbagd as this is the cleaner way to access devices.

This should also ensures ratbag-ctl gets promoted to be the actual CLI interface for libratbag (or ratbag-command can be converted to use dbus).

whot:
Let me expand on that:
* libratbag was originally *intended* to be a generic device access library
* it requires root, so it needs a wrapper to be useful on a normal setup
* at that point the wrapper becomes the generic library applications use, not libratbag itself
* the interface the wrapper uses (dbus, in the case of ratbagd) becomes the real API, so having separate APIs in libratbag becomes cumbersome for little gain
* and if we expect to always ship a wrapper process, we might as well accept that and start using it to our advantage (software profiles, etc.)

Everybody wins, except people with a terminal allergy to dbus :)

whot:
Ok, given the fairly low outrage here I'll probably go ahead with this very soon

whot:
Closed va #184

PR #184, opened 2017-06-29
Merge ratbagd into libratbag

whot:
This merges ratbagd into libratbag, keeping the histories of both projects intact (hence the rather large PR). The meson.build files are merged and a few things are shuffled around to look like one normal project directory. The build artifacts are still the same, i.e. we still build libratbag and ratbagd.

https://github.com/libratbag/libratbag/issues/179
https://github.com/libratbag/ratbagd/issues/31


whot:
That build failure is caused by a missing libsystemd-devel that is now required because ratbagd relies on it. We can't really fix that, semaphore only provides Ubuntu 14.04 and that doesn't have systemd or the version we need. Only option would be to build systemd from scratch but that's bound to have a bunch of other dependency issues.

Might be time to ditch semaphore and find something newer.

PR #186, opened 2017-06-29
README cleanups

whot:
A bit of tidying up the README.md after the ratbagd merge.


whot:
Updated to also remove the semaphore build script, previous PR only had the README.md change

bentiss:
FWIW, I also deleted the webhook. So if we ever want to re-enable Semaphore in the future, we'll have to re-add such a hook.

Issue #188, opened 2017-06-29
systemd unit file is installed in /usr/lib64/


whot:
Fedora uses /usr/lib64 as libdir and we use libdir as default for the systemd-unit-dir. Our systemd service files end up in /usr/lib64/systemd/system where there are happily ignored.

We could get the full path from the systemd.pc file ('systemdsystemunitdir') but that would break the expected behaviour of installing into a custom prefix. It's probably best to just hack around this particular behaviour and move on.

Hjdskes:
Fixed in https://github.com/libratbag/libratbag/pull/189.

PR #189, opened 2017-06-29
meson: hack around Fedora's systemd libdir confusion

whot:
Fedora uses /usr/lib64 as libdir but systemd uses /usr/lib for its files. We
used libdir to install the unit files, which then end up in
/usr/lib64/systemd/systemd where it will be ignored. There's no good way to
avoid this beyond detecting lib64 for this use-case and hacking around it.

https://github.com/libratbag/libratbag/issues/188

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -45,7 +45,8 @@ config_h.set_quoted('LIBRATBAG_DATA_DIR', libratbag_data_dir)
 pkgconfig = import('pkgconfig')
 dep_udev = dependency('libudev')
 dep_libevdev = dependency('libevdev')
-dep_systemd = dependency('libsystemd', version : '>=227')
+dep_libsystemd = dependency('libsystemd', version : '>=227')
+dep_systemd = dependency('systemd')
 dep_lm = cc.find_library('m')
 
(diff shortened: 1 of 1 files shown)


bentiss:
Is it possible in meson to check if `/etc/redhat-release` is there, and if so, adapt the path accordingly?

whot:
Ok, new approach. This one detects a /usr prefix (i.e. a system installation) and then mangles things as need be. That should cover the main use-cases of normal install vs custom prefix install.

bentiss:
Actually, looking at the cicrlce ci build on Ubuntu: https://circleci.com/gh/libratbag/libratbag/13 the path appeared to be changed too. Is that expected?

bentiss:
Replying to self: looks like the `system` folder is `/lib/systemd/system`, not `/usr/lib/systemd/system`, so it seems to be doing the right thing.
Now I wonder if the udev/hwdb rules are correctly placed there :/

whot:
isn't /lib just a symlink to /usr/lib? So depending on how the path is handled, it may show up as /lib sometimes, /usr/lib other times

PR #192, opened 2017-07-03
circle.yml: add install step

bentiss:
It doesn't cost much to actually install the files, so we can check where they are installed.

Also add mesonconf before building so we can see in the logs the various configure environment.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -6,7 +6,12 @@ libratbag_references:
       name: Build and test
       command: |
         meson build --prefix=/usr
+        mesonconf build
         ninja -C build test
+  install: &install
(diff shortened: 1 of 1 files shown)


whot:
maybe add a uninstall too and check if there's any files left?

bentiss:
> maybe add a uninstall too and check if there's any files left?

Good point. I tried to run the daemon and see if ratbagctl was picking up the service, but unfortunatelly, circle-ci doesn't allow systemd to run within the container.

So after installing the files, it might be interesting to check if they are installed in the distribution canonical place, and then remove them. Maybe not today though :)

PR #194, opened 2017-07-03
circle.yml: uninstall and checks if there are any leftovers

bentiss:
Try to uninstall and see if there are leftovers.

Currently the tests fail as `/usr/share/ratbagd` is not uninstalled.
I couldn't find out what magic is required for it to be removed by meson, so submitting the PR in the hope someone tells me what to fix in `meson.build`.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -12,6 +12,19 @@ libratbag_references:
     run:
       name: Installing
       command: ninja -C build install
+  check_uninstall: &check_uninstall
+    run:
+      name: Checking if any files are left after uninstall
+      command: |
+        PREFIX=/root/test_install
+        meson build_install --prefix=$PREFIX
+        ninja -C build_install install
+        ninja -C build_install uninstall
+        if [ -d $PREFIX ]
+        then
+          tree $PREFIX
+          exit 1
+        fi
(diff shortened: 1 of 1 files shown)


whot:
that seems overly complicated :) how about installing into a custom prefix, uninstalling and then making sure nothing but directories remain? Otherwise we may miss hwdb files, etc.

bentiss:
> that seems overly complicated :)

.....Maybe....

> how about installing into a custom prefix, uninstalling and then making sure nothing but directories remain? Otherwise we may miss hwdb files, etc.

That would work too :)

Anyway, any ideas on how to make meson/ninja remove the shared directory in the end?

whot:
The directories being left over after uninstall is a meson bug, see https://github.com/mesonbuild/meson/issues/2032

bentiss:
OK, pushed some (temporary) changes to make the tests happy: I'll update the script to pull directly from meson/master once mesonbuild/meson#2032 is fixed

bentiss:
> not super happy to pull this from a 'random' branch, but it'll do for now.

We could also have waited a day or two before merging. Anyway.
Feel free to ping me when the meson bug is fixed upstream.

PR #195, opened 2017-07-04
fix meson complaints

whot:
meson git master added warnings for the places where unsupported keywords are in use. And we were using them (outrageous, I know!), so here's a PR to remove the warning. Because there's already enough shouting going on in the world, we don't need more from meson.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -22,7 +22,7 @@ fedora_settings: &fedora_settings
   steps:
     - run:
         name: Install prerequisites
-        command: dnf install -y git gcc gcc-c++ meson check-devel libudev-devel libevdev-devel doxygen graphviz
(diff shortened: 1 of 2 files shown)


bentiss:
Care to add valgrind in circle.yml in both fedora and ubuntu setups so that the CI still works???

Thanks :)

whot:
done. tests are all happy now

bentiss:
Quick question (merging this anyway), don't we want to trigger the valgrind test in the CI too? Or is tat only doable with actual hardware?

bentiss:
Aargh, I tried using the "fix conflicts" feature, and it seems it made some bullshit in the PR. I'll manually apply it.

whot:
For the valgrind test, we'd need a separate command: `mesontest --setup=valgrind` to be run in the build directory. It does the same as the normal test, so if the normal test works, valgrind should too.

PR #200, opened 2017-07-04
ratbagd: differentiate SetResolution from SetResolutionXY

bentiss:
The code does not fallback on ratbag_resolution_set_dpi(), so we probably need a custom API for it.

This is where dragons start. This breaks current API/ABI, no ideas if this is the right way of doing this.


whot:
I intentionally left this out because IMO it's not needed. for libratbag yes, because it writes to the hardware. But for ratbagd - any caller can just call SetResolution(500, 500) if they need to.

bentiss:
OK, then we need at least to amend ratbagd so it calls the correct libratbag function given the libratbag function doesn't fall back on the other one if the capability is not there.

whot:
Fixed via 251b6023230b59abffe

Issue #205, opened 2017-07-05
ActiveResolutionChanged should only emit when the active resolution changed.


Hjdskes:
Right now, it's emitted for every change: https://github.com/libratbag/libratbag/blob/master/ratbagd/ratbagd-resolution.c#L98

whot:
yep, definitely a bug. We don't have a call to change the active resolution in ratbagd, so this signal is a bit superfluous.

Hjdskes:
Fixed in #208.

PR #208, opened 2017-07-05
ratbagd: drop the ActiveResolutionChanged signal generation

whot:
Leave the signal in the DBus API for now so we can hook it up later, but
stop the signal being generated. Changing the dpi does not change the active
resolution.

https://github.com/libratbag/libratbag/issues/205


whot:
update: added a fix to generate the signals on the right interface name

Issue #204, opened 2017-07-05
Add support for theming to ratbagd


whot:
This is required for #182, especially to merge the new SVGs while they're still WIP without breaking everything else.

Proposal is to add three new APIs to ratbagd:
- `GetSvgBaseDir()` -> /usr/share/libratbag
- `GetThemes()` -> [ 'default', 'gnome' ]
- `GetSvg()` -> 'logitech-g500s.svg'

And the caller then assembles the path to the right file. Alternatively, an option could be to supply the theme as argument to `GetSvg('gnome')` and get the full path.

bentiss:
> Alternatively, an option could be to supply the theme as argument to `GetSvg('gnome')` and get the full path.

I think I like this one better.

Hjdskes:
^ Me too

Hjdskes:
Done through https://github.com/libratbag/libratbag/pull/209, but note https://github.com/libratbag/libratbag/pull/212 as a followup.

PR #209, opened 2017-07-06
Add theming support to ratbagd

whot:
Two new DBus interfaces:
org.freedesktop.ratbagd1.Manager.Themes -> string array
org.freedesktop.ratbagd1.Device.GetSvg('theme') -> string

The Themes property is a static list of available themes. These are currently
hardcoded in ratbagd, but the theme 'default' is guaranteed to exist.

The GetSvg() method takes a theme name and returns the full file path to the
themed SVG.

Longer-term this will deprecate the Svg and SvgPath properties in the Device
but for now we leave them there.

https://github.com/libratbag/libratbag/issues/204

### DIFF
--- a/DBusInterface.md
+++ b/DBusInterface.md
@@ -14,6 +14,8 @@ Interfaces:
 The **org.freedesktop.ratbag1.Manager** interface provides:
 - Properties:
   - Devices -> array of object paths with interface Device
+  - Themes -> array of strings with theme names. The theme 'default' is
+              guaranteed to be available.
 - Signals:
   - DeviceNew -> new device available, carries object path
   - DeviceRemoved -> device removed, carries object path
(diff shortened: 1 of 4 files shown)


DBusInterface.md
```diff
@@ -30,6 +32,9 @@ The **org.freedesktop.ratbag1.Device** interface provides:
 - Methods:
   - GetProfileByIndex(uint) -> returns the object path for the given index
   - Commit() -> commit the changes to the device
+  - GetSvg(string) -> returns the svg for the theme given, must be one of
```
Hjdskes: Maybe this should say that it returns an absolute path to the SVG, as string.

Also, since `ratbagd_device_get_theme_svg` can return an empty string (through [`goto out`](https://github.com/libratbag/libratbag/pull/209/files#diff-de4a098bea35a92dcc7b13868920af30R209)), maybe you should mention this as well.

bentiss: > Also, since `ratbagd_device_get_theme_svg` can return an empty string (through `goto out`), maybe you should mention this as well.

Well, we should probably mention that there is no guarantees that the given path actually exists. `ratbagd` doesn't check if there is a file at this path, so the caller needs anyway to check the returned value (and fall back to something else).

whot: Fixed locally, but fwiw I have some plans to extend the dbus documentation with better markup anyway.

whot:
rebased on master, only change is the documentation fix

PR #212, opened 2017-07-07
data: move default SVGs under `default/` subdirectory..

Hjdskes:
.. and install them in the appropriate directory after #209.

In preparing to rebase #182, I noticed that the SVGs hadn't yet been moved to a theme directory. This PR fixes that; I'll rebase #182 on top of it.


whot:
I don't think there's a 'requirement' for it as such, but it'll end up being better organised and be more obvious to the casual viewer that the svgs are categorised

PR #214, opened 2017-07-08
data: add toned-down SVG for Logitech G500/G500s

cvuchener:
Still WIP, do not merge yet. But I open the pull request for discussion.

The `text-align` properties are not set on the leaders. I did not find how to add that on rectangles in Inkscape.

I named the layers but that set the `inkscape:label` attribute not the `id` one. I guess the `id` is needed although the readme is not clear about that.

For the main buttons, I added open paths (copied from the light gray area) so there is no border on the bottom. The buttons should be there although not visible.

Since the G500 and the G500s are exactly the same when removing details. They should use the same file. Because of the themes I cannot set the same svg file from the hwdb, so I added a symlink. What is the preferred way to achieve that? meson installs a copy of the file instead of the symlink.

@Hjdskes, I wanted to test your version of piper, but the ledsandbuttons branch is broken, because of lack of theme support, I guess. The ratbagd-themes branch does not seem to work (crashes with `bAttributeError: 'RatbagdDevice' object has no attribute 'svg_path'`).

### DIFF
--- a/data/gnome/logitech-g500.svg
+++ b/data/gnome/logitech-g500.svg
@@ -1,0 +1,484 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<!-- Created with Inkscape (http://www.inkscape.org/) -->
+
+<svg
+   xmlns:dc="http://purl.org/dc/elements/1.1/"
+   xmlns:cc="http://creativecommons.org/ns#"
+   xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#"
+   xmlns:svg="http://www.w3.org/2000/svg"
+   xmlns="http://www.w3.org/2000/svg"
+   xmlns:sodipodi="http://sodipodi.sourceforge.net/DTD/sodipodi-0.dtd"
+   xmlns:inkscape="http://www.inkscape.org/namespaces/inkscape"
+   width="461.86667"
+   height="437.33334"
+   id="svg2"
+   version="1.1"
+   inkscape:version="0.92+devel unknown"
+   sodipodi:docname="logitech-g500.svg"
+   viewBox="0 0 433 410">
+  <style
+     id="style5346" />
+  <defs
+     id="defs4" />
+  <sodipodi:namedview
+     id="base"
+     pagecolor="#ffffff"
+     bordercolor="#666666"
+     borderopacity="1.0"
+     inkscape:pageopacity="0.0"
+     inkscape:pageshadow="2"
+     inkscape:zoom="4.0000004"
+     inkscape:cx="184.69456"
+     inkscape:cy="292.7786"
+     inkscape:document-units="px"
+     inkscape:current-layer="Device"
+     showgrid="false"
+     showguides="true"
+     inkscape:guide-bbox="true"
+     inkscape:window-width="1853"
+     inkscape:window-height="1016"
+     inkscape:window-x="1507"
+     inkscape:window-y="27"
+     inkscape:window-maximized="1"
+     units="px"
+     fit-margin-top="20"
+     fit-margin-left="-40"
+     fit-margin-right="-40"
+     fit-margin-bottom="20"
+     inkscape:document-rotation="0">
+    <sodipodi:guide
+       orientation="-0.98480775,-0.17364818"
+       position="118.1781,220.98096"
+       id="guide4656"
+       inkscape:locked="false" />
+    <sodipodi:guide
+       orientation="-0.98480775,-0.17364818"
+       position="124.47548,146.18232"
+       id="guide4658"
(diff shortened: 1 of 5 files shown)


Hjdskes:
Try the `resolutions` branch, that one should work, the others are outdated and only there because there is some stuff in there I'm going to use later.

cvuchener:
I got the "Device SVG is incompatble", so I fixed the layer ids. But buttons cannot be tested in this version. I had to change the layers to group to edit the id and then back to layer.

bentiss:
> I had to change the layers to group to edit the id and then back to layer.

I am not sure how you edited the id, but the easiest is simply to open the XML editor (Ctrl-Shift-X) and edit the SVG elements directly. You can also afdd any XML property in it to add the text-align property.

Not sure if it helps, but at least I said what I wanted to say :)

cvuchener:
Thanks for the tip. It is better than the object window for doing this. It is also easier for checking that every id is there. I fixed a few errors I missed.

It remains only to test the buttons, but that cannot be done with the current piper. Changing the background colors for these objects in Inkscape looks fine. I guess you will have to do the same to see what they look like.

I changed what I wanted to, I think it is ready for review.

Hjdskes:
I can maybe test it sometime this weekend, otherwise it'll be Monday. And yea, I also use the XML editor to set the ids and the text align attribute.

Hjdskes:
Oke, back at my laptop now so I can give a proper response. It looks great! Here's a picture:

screenshot from 2017-07-08 21-42-01

A few things I noticed:

1. The arrows next to the scroll wheel that indicate tilting are not recognized as buttons. Please treat the arrows as buttons, with the identifier and paths and everything.
2. There seem to be two buttons to the left of button 1, but they aren't mapped to anything in Piper (see the screenshot). Is this intentional or an oversight?

> The text-align properties are not set on the leaders. I did not find how to add that on rectangles in Inkscape.

This is already mentioned but indeed the SVG editor is the best approach here. I'll update the requirements so that they mention this Monday.

> I named the layers but that set the `inkscape:label` attribute not the id one. I guess the id is needed although the readme is not clear about that.

Good point; I'll ammend the README.

> For the main buttons, I added open paths (copied from the light gray area) so there is no border on the bottom. The buttons should be there although not visible.

Excellent solution, I really like how it looks (see screenshot in this comment). Is this something you can also fix for the already merged logitech-g700?

> Since the G500 and the G500s are exactly the same when removing details. They should use the same file. Because of the themes I cannot set the same svg file from the hwdb, so I added a symlink. What is the preferred way to achieve that? meson installs a copy of the file instead of the symlink.

This is something for @whot or @bentiss to answer, but I suppose a copy isn't too bad since it's a small file.

cvuchener:
> [picture]

I wanted the button to be below the the DPI buttons (the two small buttons on the left) and the dark part around them. But now I see that, even if you modified the application to not draw it on top, the left part of the border would be hidden. I will need to go around.

> A few things I noticed:

They should be there, I could have made some mistake but is weird that it would be exactly with the last four buttons. Isn't it a problem from your prototype that would make it stop after the first six buttons (like using it with a 6 button mouse instead of using a real G500).

> Is this something you can also fix for the already merged logitech-g700?

I will have a look after this is done.

Hjdskes:
> Isn't it a problem from your prototype that would make it stop after the first six buttons (like using it with a 6 button mouse instead of using a real G500).

Derp, that should be it. Note to self: don't test after a long day.

cvuchener:
I changed the left button so it does not go over the DPI buttons.

cvuchener:
I did the G700, tell me if you are happy with the buttons.

Hjdskes:
Both are excellent, thanks! There's a little border still "sticking out" on the left side when hovering button 0 on the G500, it doesn't bother me personally but I'll wait with merging in case you want to fix it:

screenshot from 2017-07-10 08-41-30

bentiss:
Thanks both of you for this excellent work.
@Hjdskes the drawing raised a question in my mind. Are the labels on the widgets going to stay as the button _name_ or is it going to show up the button _action_?

bentiss:
> As I said, both look good to me. Maybe @whot or @bentiss wants to comment on the symlink thing before I merge it, though.

I personally don't care much if there is a duplicate of the file. The SVG are small (the one I have from the beginning of this PR is 20K), and if we really want a symlink in the end, we can always add this later through a bug/PR.

Hjdskes:
> @Hjdskes the drawing raised a question in my mind. Are the labels on the widgets going to stay as the button name or is it going to show up the button action?

It's going to show the assigned action, the name right now (or rather the index) is just a placeholder.

cvuchener:
The border should be fixed.

Hjdskes:
Great, thanks!

PR #217, opened 2017-07-11
Fix HID++2.0/1b04 divert and remapping

cvuchener:
Partially fixes issue #124.

Remapped buttons were diverted at the same time, resulting in a disabled button (diverted buttons don't send their remapped event).

This fixes both remapping the button and changing the divert state correctly.


whot:
Not 100% on the first commit, I'll let @bentiss decide on that. The second one is correct, thanks.

PR #220, opened 2017-07-12
ratbagd: fix ratbagd_device_new() error path

bentiss:
ratbagd_device_new() increments the ref count on the ratbag device, but forgets to unref it in case of an error.
Instead of doing that, just steal the reference (i.e. do nothing), and let the caller unref the device only when an error occurs.

### DIFF
--- a/ratbagd/ratbagd-device.c
+++ b/ratbagd/ratbagd-device.c
@@ -385,7 +385,8 @@ int ratbagd_device_new(struct ratbagd_device **out,
 
 	device->ctx = ctx;
 	rbnode_init(&device->node);
-	device->lib_device = ratbag_device_ref(lib_device);
+	/* note: we steal the reference here */
+	device->lib_device = lib_device;
 
 	device->name = strdup(name);
 	if (!device->name)
--- a/ratbagd/ratbagd.c
+++ b/ratbagd/ratbagd.c
@@ -313,11 +313,8 @@ static int ratbagd_reset_test_device(sd_bus_message *m,
 	device = ratbag_device_new_test_device(ctx->lib_ctx, &ratbagd_test_device_descr);
 
 	r = ratbagd_device_new(&ratbagd_test_device, ctx, "test_device", device);
-
-	/* the ratbagd_device takes its own reference, drop ours */
-	ratbag_device_unref(device);
-
 	if (r < 0) {
+		ratbag_device_unref(device);
 		log_error("Cannot track test device\n");
 		return r;
 	}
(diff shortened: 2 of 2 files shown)


ratbagd/ratbagd-device.c
```diff
@@ -385,7 +385,8 @@ int ratbagd_device_new(struct ratbagd_device **out,
 
 	device->ctx = ctx;
 	rbnode_init(&device->node);
-	device->lib_device = ratbag_device_ref(lib_device);
+	/* note: we steal the reference here */
+	device->lib_device = lib_device;
```
whot: no, sorry. never steal references, the whole point of having ref/unref is that you don't have to worry about who has the pointer, you always get a reference and unref it when you're done with it.

this can be fixed properly with the goto error: approach that calls unref if device->lib_device is not NULL

bentiss:
Alright, then after a few thoughts, either we allow to have a ugly error path, or we need to convert libratbag to also have the cleanups functions so we have a simple error path.

I would rather have the _cleanup_ decorator in libratbag, but this will require more work before we can merge this.

whot:
libratbag's `ratbag_device_new_from_udev_device()` already works like it should? You either get a functional device or an error and no device.

But yeah, using `cleanup` is fine with me otherwise but last time I checked there wasn't much that would lend itself to cleanup.

whot:
Fixed with the last commit in https://github.com/libratbag/libratbag/pull/225, I think :)

bentiss:
Are you sure about that?

Actually I might have overlook this bug because ratbagd_device_free calls
unref on lib_device, which was my point.

I do not see how having an explicit call to ratbagd_device_free is better
than using the _cleanup_ function.

bentiss:
> The only failure path in there now calls `ratbagd_device_free` and that unrefs the lib device. So we behave correctly now - on success we have a valid ref in in the struct, on failure we remove that ref before returning.

Just to be sure I am understanding correctly how _cleanup_ works. In the current code (without your PR), when the function exits, we free the pointer no matter what?

whot:
cleanup inserts the code when the variable goes out of scope (e.g. when returning from the function). It's unconditional, that's why we have to set device to NULL before we return to make sure we don't clean it up on return. You can't return a variable that's marked cleanup either because it will be freed and the return value is invalid.

bentiss:
Then I don't get the point of the last commit of PR #225. I'll answer in the PR directly

whot:
Ok, much ado about nothing, I think we can close this one now :)

PR #225, opened 2017-07-13
Macro cleanup

whot:
Also fixes the issue in https://github.com/libratbag/libratbag/pull/220


whot:
yeah, good point, not sure why I thought that was necessary. I guess I got confused by #220 :)
Anyway, rebased, commit removed, pushed.

bentiss:
Could you force push the rebased changed, please? It looks like we still have the original code in the PR.

whot:
whoops, pushed to the wrong remote. updated now, I'll push this later if you don't merge it before

PR #228, opened 2017-07-14
Add logitech G303 SVG for gnome theme

whot:


Hjdskes:
This looks like it should work, except for these two points:

1. The canvas isn't resized according to the instructions: *"The canvas should be resized so that there is a 20px gap between the device and the edge of the canvas and no gap between the 1x1 pixels and the canvas."*
2. The SVG reports having two LEDs, LED0 being the logo and LED1 what seems to be in the scroll wheel (like my G403, see the highlight in the screenshot). The leader for LED1 appears next to that of LED0, however:

screenshot from 2017-07-14 10-16-29

whot:
Force-pushed changes:
- minor movement of the lower LED markers
- resized canvas for 20px space around device

I double checked the leader lines for the LEDs and they are named correctly. What did you mean by 
> The leader for LED1 appears next to that of LED0, however:

Hjdskes:
> I double checked the leader lines for the LEDs and they are named correctly. What did you mean by

>> The leader for LED1 appears next to that of LED0, however:

As you can see in the [screenshot](https://github.com/libratbag/libratbag/pull/228#issuecomment-315300777), the leader line of LED1 originates almost next to the leader line of LED0, while the LED in the scroll wheel gets highlighted. So, either:

1. LED1 is actually in the scrollwheel (like in my G403) and the leader needs to be moved up to originate close to the scroll wheel, or
2. LED1 is actually where the leader line currently originates, in which case the LED path should be removed from the scrollwheel and an LED should be drawn where the leader is currently.

Hjdskes:
@whot please tag me when you fixed the thicker border and the z-ordering, then I'll merge this.

whot:
@Hjdskes ok, I think I got everything this time, please pull if you're happy with it

Issue #203, opened 2017-07-05
Make ratbagctl take the same arguments as ratbag-command


whot:
In the medium-term plan to replace/remove ratbag-command, ratbagctl needs to do the exact same thing, albeit by going through ratbagd. This means the commandline interface should be the same.

Hjdskes:
Ratbagctl's bindings are not up to date, just FYI.

whot:
Benjamin has the ratbagd.py sync patch for that in https://github.com/libratbag/libratbag/pull/207

bentiss:
This has been almost done in PR #218.

We are missing a few commands and proper keycode mappings but we should be close to be able to close this one.

bentiss:
Feature complete:
https://226-41394739-gh.circle-artifacts.com/0/root/libratbag/build/meson-logs/testlog.txt
https://circleci.com/gh/libratbag/libratbag/226#artifacts/containers/0

\o/

PR #229, opened 2017-07-14
Drop libratbag.so as separate library

whot:
This removes libratbag as library, so the only product of libratbag is ratbagd and its dbus interface.

I'm giving it the WIP label, we should make sure #203  is fixed before we merge this


whot:
I think we're close enough with ratbagctl that we can consider #203 complete enough. The rest should just be treated as bugs.

PR #234, opened 2017-07-18
tools: copy ratbagd.py from piper and integrate it in ratbagctl

bentiss:
We are using a custom python script to generate the final ratbagctl so we can strip out the headers.

Now merging and testing ratbagd.py in piper will be easier than ever.


meson.build
```diff
@@ -515,12 +515,19 @@ install_data('dbus/org.freedesktop.ratbag1.conf',
...
+
+custom_target('ratbagctl',
+  output : 'ratbagctl',
+  input : 'tools/ratbagctl.in',
+  build_by_default : true,
+  command : [py3, join_paths(meson.source_root(), 'tools', 'merge_ratbagd.py'),
+				'@INPUT@',
+				join_paths(meson.source_root(), 'tools', 'ratbagd.py'),
+				'--output', '@OUTPUT@',
+				'--version', meson.project_version()])
```
whot: you're missing the `install: true` and `install_dir: get_option('bindir')` options here

tools/merge_ratbagd.py
```diff
@@ -0,0 +1,68 @@
...
+
+import argparse
+import os
+import stat
+import sys
+
+
+def print_ratbagctl(ratbagctl_path, ratbagd_path, version_string):
+    with open(ratbagctl_path, 'r') as ratbagctl:
+        with open(ratbagd_path, 'r') as ratbagd:
```
whot: I just found out you can do: `with open() as a, open() as b:` if you have multiple variables. Saves nesting and would be appropriate here

PR #238, opened 2017-07-20
enable/disable profiles

bentiss:
enable and disable profiles in ratbagd, ratbagctl and ratbag-command.

The hidpp10 implementation was not done with that in mind, which explains the large commits for them.
The result is now less communications with the HID++ 1.0 devices as we are caching the data inside libratbag.

### DIFF
--- a/hwdb/70-libratbag-mouse.hwdb
+++ b/hwdb/70-libratbag-mouse.hwdb
@@ -69,31 +69,41 @@ mouse:usb:v046dpc041:name:*:
  RATBAG_HIDPP10_DPI_LIST=400;800;1600;2000
 
 # G5 2007
+# http://support.logitech.com/en_us/product/g5-laser-mouse-product/specs
 mouse:usb:v046dpc049:name:*:
  RATBAG_DRIVER=hidpp10
  RATBAG_HIDPP10_DPI_LIST=400;800;1600;2000
+ RATBAG_HIDPP10_PROFILE_COUNT=1
 
 # G7
+# http://support.logitech.com/en_us/product/g7-laser-cordless-mouse/specs
 mouse:usb:v046dpc51a:name:*:
  RATBAG_DRIVER=hidpp10
+ RATBAG_HIDPP10_PROFILE_COUNT=1
 
 # G9
+# http://support.logitech.com/en_us/product/g9-laser-mouse/specs
 mouse:usb:v046dpc048:name:*:
  RATBAG_DRIVER=hidpp10
  RATBAG_HIDPP10_DPI_LIST=0;200;400;600;800;1000;1200;1400;1600;1800;2000;2200;2400;2600;2800;3000;3200
  RATBAG_HIDPP10_PROFILE=G9
+ RATBAG_HIDPP10_PROFILE_COUNT=5
 
 # G9x [Original]
+# http://support.logitech.com/en_us/product/g9x-laser-mouse/specs
 mouse:usb:v046dpc066:name:*:
  RATBAG_DRIVER=hidpp10
  RATBAG_HIDPP10_DPI=0:5700@23.53
  RATBAG_HIDPP10_PROFILE=G500
+ RATBAG_HIDPP10_PROFILE_COUNT=5
 
 # G9x [Call of Duty MW3 Edition]
+# http://support.logitech.com/en_us/product/laser-mouse-g9x/specs
 mouse:usb:v046dpc249:name:*:
  RATBAG_DRIVER=hidpp10
  RATBAG_HIDPP10_DPI=0:5700@23.53
  RATBAG_HIDPP10_PROFILE=G500
+ RATBAG_HIDPP10_PROFILE_COUNT=5
 
 # G300
 mouse:usb:v046dpc246:name:*:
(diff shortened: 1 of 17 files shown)


ratbagd/ratbagd-profile.c
```diff
@@ -325,6 +331,43 @@ static int ratbagd_profile_set_active(sd_bus_message *m,
...
+	int r;
+
+	r = sd_bus_message_read(m, "");
+	if (r < 0)
+		return r;
+
+	(void) sd_bus_emit_signal(sd_bus_message_get_bus(m),
+				  RATBAGD_OBJ_ROOT,
+				  RATBAGD_NAME_ROOT ".Profile",
+				  "EnabledProfileChanged",
```
Hjdskes: Just checking: is the intended usage here to check for `Enabled` in the callback of this signal? That'd be another round trip over DBus; perhaps we can emit a `ProfileEnabled`/`ProfileDisabled` signal instead.

tools/ratbagd.py
```diff
@@ -374,6 +383,14 @@ def set_active(self):
         """Set this profile to be the active profile."""
         return self._dbus_call("SetActive", "")
 
+    def enable(self):
+        """Enable this profile."""
+        return self._dbus_call("Enable", "")
+
+    def disable(self):
+        """Enable this profile."""
```
Hjdskes: Nitpick: should be "Disable" here

ratbagd/ratbagd-profile.c
```diff
@@ -346,17 +389,35 @@ static int ratbagd_profile_get_resolution_by_index(sd_bus_message *m,
...
 const sd_bus_vtable ratbagd_profile_vtable[] = {
 	SD_BUS_VTABLE_START(0),
+	SD_BUS_PROPERTY("Enabled", "b", ratbagd_profile_is_enabled, 0, 0),
 	SD_BUS_PROPERTY("Index", "u", NULL, offsetof(struct ratbagd_profile, index), SD_BUS_VTABLE_PROPERTY_CONST),
 	SD_BUS_PROPERTY("Resolutions", "ao", ratbagd_profile_get_resolutions, 0, 0),
 	SD_BUS_PROPERTY("Buttons", "ao", ratbagd_profile_get_buttons, 0, 0),
 	SD_BUS_PROPERTY("Leds", "ao", ratbagd_profile_get_leds, 0, 0),
 	SD_BUS_PROPERTY("ActiveResolution", "u", ratbagd_profile_get_active_resolution, 0, 0),
 	SD_BUS_PROPERTY("DefaultResolution", "u", ratbagd_profile_get_default_resolution, 0, 0),
+	SD_BUS_METHOD("Enable", "", "u", ratbagd_profile_enable, SD_BUS_VTABLE_UNPRIVILEGED),
```
Hjdskes: Why not `SetEnabled(bool)` like we have for `SetActive`?

bentiss:
> The changes made to ratbagd and the DBus interface look good to me, just have some questions regarding the intended use and some documentation fixes.

Just a quick FYI, @whot is currently looking at the dbus API and ideally we would suppress methods and only have properties. I did not think much about the API right now, I was just trying to have something that just works :)

I'll amend the changes, but beware that the dbus API will likely change in the (near) future.

ratbagd/ratbagd-profile.c
bentiss: Not a strong opinion. I can emit both if you want.

ratbagd/ratbagd-profile.c
bentiss: Because... lazy me?

tools/ratbagd.py
bentiss: good catch

Hjdskes:
I'm aware of this and am looking forward to it :)

whot:
that first commit (add missing Commit()) shows a slight problem we have with ratbagd. ratbagd returns the state of the device before the commit, so in the extreme case we can have a client change that state but never commits it to the device. The next client will then get a state that does not reflect the device. I wonder if we can do client tracking in ratbagd to reset the device on client disconnect.

hwdb/70-libratbag-mouse.hwdb
```diff
@@ -69,31 +69,39 @@ mouse:usb:v046dpc041:name:*:
  RATBAG_HIDPP10_DPI_LIST=400;800;1600;2000
 
 # G5 2007
+# http://support.logitech.com/en_us/product/g5-laser-mouse-product/specs
 mouse:usb:v046dpc049:name:*:
  RATBAG_DRIVER=hidpp10
  RATBAG_HIDPP10_DPI_LIST=400;800;1600;2000
```
whot: is there something missing? If not, the commit msg should point out that those two are unknown, otherwise it looks like a copy/paste error.

hwdb/70-libratbag-mouse.hwdb
```diff
@@ -69,31 +69,39 @@ mouse:usb:v046dpc041:name:*:
...
 
 # G7
+# http://support.logitech.com/en_us/product/g7-laser-cordless-mouse/specs
 mouse:usb:v046dpc51a:name:*:
  RATBAG_DRIVER=hidpp10
```
whot: as above

bentiss:
> that first commit (add missing Commit()) shows a slight problem we have with ratbagd. ratbagd returns the state of the device before the commit, so in the extreme case we can have a client change that state but never commits it to the device. The next client will then get a state that does not reflect the device. I wonder if we can do client tracking in ratbagd to reset the device on client disconnect.

Nah. I think we should consider the daemon state as being the correct one. If a client _forgets_ to call commit(), we should not have a different state for other clients. This will be race prone and we might end up having too much code for no benefit.
Anyway, if one client forgets to commit, the next one will more likely commit given that we expect clients to actually change the device.

> looks good, but i'd like this to be rebased on top of #242 please, sorry.

No worries. When I saw your local brnach, I already made up my mind that I would have to rewrite it.
I could also have dropped the ratbagd part, but that would just differ the commit in a separate PR, so no gain.

hwdb/70-libratbag-mouse.hwdb
bentiss: IIRC, they are not unknown, they are just using the default value 1.
I guess we can still override this here.

bentiss:
> looks good, but i'd like this to be rebased on top of #242 please, sorry.

Done. Squashed the 2 comments and also changed the `Enabled` property to be read-write.

PR #241, opened 2017-07-24
Test suite cleanups

whot:
Hook up ratbagctl.test to the test suite and make sure it passes both normal and valgrind runs


whot:
Ok, had to push a few more, we can't run ratbagctl in CircleCI because we don't have dbus. Still, imo it should run on a local ninja test

PR #242, opened 2017-07-24
DBus interface revamp

whot:
This is on top of #241 and should only be merged when that one is merged.

When I started tidying up the DBusInterface.md document I noticed a fair few inconsistencies with the API. A few basically superfluous calls and some imbalanced property handling (having a 'foo' readonly property but a SetFoo() method - instead we can just make the property read-write).

Note that his is not complete but I'm about to head out for GUADEC, so better to get this in now. The bits missing are changing the Button interface from using strings to using the libratbag enums and the LED type string to an enum too.

ratbagd.py will need to be synced to piper after merging this but no client changes are needed, this is all handled within the bindings.

### DIFF
--- a/ratbagd/ratbagd-resolution.c
+++ b/ratbagd/ratbagd-resolution.c
@@ -38,6 +38,7 @@
 #include "libratbag-util.h"
 
 struct ratbagd_resolution {
+	struct ratbagd_profile *profile;
 	struct ratbag_resolution *lib_resolution;
 	unsigned int index;
 	char *path;
@@ -105,6 +106,24 @@ static int ratbagd_resolution_set_resolution(sd_bus_message *m,
 	return sd_bus_reply_method_return(m, "u", r);
 }
 
+static int ratbagd_resolution_default_signal_cb(sd_bus *bus,
+						struct ratbagd_resolution *resolution)
+{
+	struct ratbag_resolution *lib_resolution = resolution->lib_resolution;
+
+	/* FIXME: we should cache is default and only send the signal for
+	 * those resolutions where it changed */
+
+	(void) sd_bus_emit_signal(bus,
+				  resolution->path,
+				  RATBAGD_NAME_ROOT ".Resolution",
+				  "IsDefault",
+				  "b",
+				  ratbag_resolution_is_default(lib_resolution));
+
+	return 0;
+}
+
 static int ratbagd_resolution_set_default(sd_bus_message *m,
 					  void *userdata,
 					  sd_bus_error *error)
@@ -116,12 +135,9 @@ static int ratbagd_resolution_set_default(sd_bus_message *m,
 	if (r < 0)
 		return r;
 
-	(void) sd_bus_emit_signal(sd_bus_message_get_bus(m),
-				  resolution->path,
-				  RATBAGD_NAME_ROOT ".Resolution",
-				  "DefaultResolutionChanged",
-				  "u",
-				  resolution->index);
+	ratbagd_for_each_resolution_signal(sd_bus_message_get_bus(m),
+					   resolution->profile,
+					   ratbagd_resolution_default_signal_cb);
 
 	return sd_bus_reply_method_return(m, "u", r);
 }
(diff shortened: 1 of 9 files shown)


ratbagd/ratbagd-resolution.c
```diff
@@ -250,16 +298,19 @@ const sd_bus_vtable ratbagd_resolution_vtable[] = {
...
+	SD_BUS_WRITABLE_PROPERTY("ReportRate", "u",
+				 ratbagd_resolution_get_rate,
+				 ratbagd_resolution_set_report_rate, 0,
+				 SD_BUS_VTABLE_UNPRIVILEGED|SD_BUS_VTABLE_PROPERTY_EMITS_CHANGE),
 	SD_BUS_PROPERTY("Maximum", "u", ratbagd_resolution_get_maximum, 0, SD_BUS_VTABLE_PROPERTY_CONST),
 	SD_BUS_PROPERTY("Minimum", "u", ratbagd_resolution_get_minimum, 0, SD_BUS_VTABLE_PROPERTY_CONST),
-	SD_BUS_METHOD("SetReportRate", "u", "u", ratbagd_resolution_set_report_rate, SD_BUS_VTABLE_UNPRIVILEGED),
-	SD_BUS_METHOD("SetResolution", "uu", "u", ratbagd_resolution_set_resolution, SD_BUS_VTABLE_UNPRIVILEGED),
 	SD_BUS_METHOD("SetDefault", "", "u", ratbagd_resolution_set_default, SD_BUS_VTABLE_UNPRIVILEGED),
-	SD_BUS_SIGNAL("ActiveResolutionChanged", "u", 0),
```
Hjdskes: Again: you forgot to remove this signal and the one below it from the bindings.

ratbagd/ratbagd.c
```diff
@@ -350,10 +349,8 @@ static inline void ratbagd_init_test_device(struct ratbagd *ctx) {}
 
 static const sd_bus_vtable ratbagd_vtable[] = {
 	SD_BUS_VTABLE_START(0),
-	SD_BUS_PROPERTY("Devices", "ao", ratbagd_get_devices, 0, 0),
+	SD_BUS_PROPERTY("Devices", "ao", ratbagd_get_devices, 0, SD_BUS_VTABLE_PROPERTY_EMITS_CHANGE),
 	SD_BUS_PROPERTY("Themes", "as", ratbagd_get_themes, 0, 0),
-	SD_BUS_SIGNAL("DeviceNew", "o", 0),
```
Hjdskes: Again: you forgot to remove this signal and the one below it from the bindings

Hjdskes:
(Grr.. commenting on individual commits makes them show up as outdated)

tools/ratbagd.py
```diff
@@ -151,7 +152,30 @@ def _get_dbus_property(self, property):
 
     def _set_dbus_property(self, property, type, value):
         # Sets a cached property on the bus.
+
+        # Take our real value and wrap it into a variant. To call
+        # org.freedesktop.DBus.Properties.Set we need to wrap that again
+        # into a (ssv), where v is our value's variant.
+        # args to .Set are "interface name", "function name",  value-variant
```
whot: it took me long enough to figure out how to send this, so I'd rather have the comment there to clarify what the code does ;)

tools/ratbagd.py
```diff
@@ -111,7 +111,11 @@ def __init__(self, interface, object_path):
         GObject.GObject.__init__(self)
 
         if _RatbagdDBus._dbus is None:
-            _RatbagdDBus._dbus = Gio.bus_get_sync(Gio.BusType.SYSTEM, None)
+            try:
+                _RatbagdDBus._dbus = Gio.bus_get_sync(Gio.BusType.SYSTEM, None)
+            except GLib.Error:
+                raise RatbagdDBusUnavailable()
+
             if _RatbagdDBus._dbus is None:
```
whot: fwiw, the Is None thing was part of the other PR and is already merged, so I'll keep this local and push it with some other PR

Hjdskes: I know it's not part of this PR; I was trying to say that according to the documentation checking for `None` **and** catching the exception isn't necessary.

whot:
Does it matter that we change the interface? ratbagd had a note from day one that this interface will change and we're not finished with the changes. the version number is also a bit of a hint :)

Anyway, addressed all o @Hjdskes comments and squashed them in, thanks for the reviews

bentiss:
Then we are good to merge :)

PR #244, opened 2017-07-24
Make RatbagdButton properties writable

Hjdskes:
@whot wanted to wait until the conversion to enums was done[¹](https://github.com/libratbag/libratbag/pull/242#issue-244965513):

> Note that his is not complete but I'm about to head out for GUADEC, so better to get this in now. The bits missing are changing the Button interface from using strings to using the libratbag enums and the LED type string to an enum too.

I needed this quickly to be able to test some things in https://github.com/libratbag/piper/pull/47, and AFAIK changing to enums is just as easy now as it was before making these properties writable. Feel free to close this and cherry-pick these commits or use them in another way if I'm wrong :)

### DIFF
--- a/DBusInterface.md
+++ b/DBusInterface.md
@@ -214,9 +214,11 @@ Index of the button
 
 String describing the button type
 #### `ButtonMapping`
-- type: `u`, read-only, mutable
+- type: `u`, read-write, mutable
 
-uint of the current button mapping (if mapping to button)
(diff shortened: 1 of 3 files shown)


whot:
yeah, I agree with @bentiss. Properties should IMO only be for straightforward things without side effects. Anything that has more side-effects should be a method. Code looks good, but I think it's better to keep the methods here, sorry.

PR #248, opened 2017-07-25
Add support for macros in ratbagd

whot:
Disclaimer: I just finished this and I'm running out of time before leaving. The test suite passes, but there may be bugs left, sorry.

Main item here is support for macros in ratbagd. I'm not 100% sure whether that'll be the final API, but the approach here is an array of tuples of `(type, value)` for each element in the macro. Macro names are picked by ratbagd. There's the theoretical case of exposing macros as separate objects on dbus but I don't think that's worth doing. DBus would allow for `SetMacro(s, a(uu))` where we can set the string as well, may be worth doing, maybe not.

Meanwhile, this will be at least useful, if it works ;)

This pull also removes the remaining string button properties and changes them to enums. The LED type is still left and should be fixed too.

### DIFF
--- a/DBusInterface.md
+++ b/DBusInterface.md
@@ -219,9 +219,11 @@ org.freedesktop.ratbag1.Button
 
 Index of the button
 #### `Type`
-- type: `s`, read-only, constant
+- type: `u`, read-only, constant
 
-String describing the button type
+Enum describing the button physical type. This type is unrelated to the
+logical button mapping and serves to easily identify the button on the
+device.
 #### `ButtonMapping`
 - type: `u`, read-only, mutable
 
--- a/DBusInterface.md
+++ b/DBusInterface.md
@@ -236,15 +236,14 @@ String of the current special mapping (if mapped to special)
 Array of uints, first entry is the keycode, other entries, if any, are
 modifiers (if mapped to key)
 #### `ActionType`
-- type: `s`, read-only, mutable
+- type: `u`, read-only, mutable
 
-String describing the action type of the button ("none", "button", "key",
-"special", "macro", "unknown"). This decides which Mapping  property has a
-value
-
+An enum describing the action type of the button, see
+ratbag\_button\_get\_action\_type for the list of enums.
+This decides which Mapping  property has a value.
 #### `ActionTypes`
-- type: `as`, read-only, constant
-Array of strings, possible values for ActionType
+- type: `au`, read-only, constant
+Array of enum ratbag\_button\_action\_type, possible values for ActionType
 
 ### Methods:
 #### `SetButtonMapping(u) → ()`
--- a/ratbagd/ratbagd-button.c
+++ b/ratbagd/ratbagd-button.c
@@ -53,90 +53,11 @@ static int ratbagd_button_get_type(sd_bus *bus,
 				   sd_bus_error *error)
 {
 	struct ratbagd_button *button = userdata;
-	const char *type = NULL;
 	enum ratbag_button_type t;
 
 	t = ratbag_button_get_type(button->lib_button);
 
-	switch(t) {
-		default:
-			log_error("Unknown button type %d\n", t);
-			/* fallthrough */
-		case RATBAG_BUTTON_TYPE_UNKNOWN:
-			type = "unknown";
-			break;
-		case RATBAG_BUTTON_TYPE_LEFT:
-			type = "left";
-			break;
-		case RATBAG_BUTTON_TYPE_MIDDLE:
-			type = "middle";
-			break;
-		case RATBAG_BUTTON_TYPE_RIGHT:
-			type = "right";
-			break;
-		case RATBAG_BUTTON_TYPE_THUMB:
-			type = "thumb";
-			break;
-		case RATBAG_BUTTON_TYPE_THUMB2:
-			type = "thumb2";
-			break;
-		case RATBAG_BUTTON_TYPE_THUMB3:
-			type = "thumb3";
-			break;
-		case RATBAG_BUTTON_TYPE_THUMB4:
-			type = "thumb4";
-			break;
-		case RATBAG_BUTTON_TYPE_WHEEL_LEFT:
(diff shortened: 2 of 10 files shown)


whot:
Also, the `WIP - do not merge` label is only because it should be tested more. If you find it works fine anyway, feel free to merge.

DBusInterface.md
```diff
@@ -236,15 +236,14 @@ String of the current special mapping (if mapped to special)
...
+- type: `u`, read-only, mutable
 
+An enum describing the action type of the button, see
+ratbag\_button\_get\_action\_type for the list of enums.
+This decides which Mapping  property has a value.
 #### `ActionTypes`
-- type: `as`, read-only, constant
-Array of strings, possible values for ActionType
+- type: `au`, read-only, constant
+Array of enum ratbag\_button\_action\_type, possible values for ActionType
```
Hjdskes: Might want to amend here that this is for the current device and not just the enum values.

DBusInterface.md
```diff
@@ -236,15 +236,14 @@ String of the current special mapping (if mapped to special)
...
-- type: `s`, read-only, mutable
-
-String describing the action type of the button ("none", "button", "key",
-"special", "macro", "unknown"). This decides which Mapping  property has a
-value
+- type: `u`, read-only, mutable
 
+An enum describing the action type of the button, see
+ratbag\_button\_get\_action\_type for the list of enums.
+This decides which Mapping  property has a value.
```
Hjdskes: Not sure if you intentionally left out the * for *Mapping

ratbagd/ratbagd-button.c
```diff
@@ -454,6 +454,90 @@ static int ratbagd_button_get_action_types(sd_bus *bus,
...
+
+		r = sd_bus_message_append(reply, "(uu)", type, value);
+		if (r < 0)
+			return r;
+	}
+
+out:
+	return sd_bus_message_close_container(reply);
+}
+static int ratbagd_button_set_macro(sd_bus_message *m,
```
Hjdskes: Intentionally no empty line here?

ratbagd/ratbagd-button.c
```diff
@@ -454,6 +454,90 @@ static int ratbagd_button_get_action_types(sd_bus *bus,
...
+			value = ratbag_button_macro_get_event_key(macro, idx);
+			break;
+		case RATBAG_MACRO_EVENT_WAIT:
+			value = ratbag_button_macro_get_event_timeout(macro, idx);
+			break;
+		}
+
+		r = sd_bus_message_append(reply, "(uu)", type, value);
+		if (r < 0)
+			return r;
```
Hjdskes: What happens if you return here without closing the container?

ratbagd/ratbagd-button.c
```diff
@@ -454,6 +454,90 @@ static int ratbagd_button_get_action_types(sd_bus *bus,
...
+	unsigned int type, value;
+	int r, idx = 0;
+	_cleanup_(ratbag_button_macro_unrefp) struct ratbag_button_macro *macro = NULL;
+
+	r = sd_bus_message_enter_container(m, 'a', "(uu)");
+	if (r < 0)
+		return r;
+
+	macro = ratbag_button_macro_new("macro");
+	while ((r = sd_bus_message_read(m, "(uu)", &type, &value)) > 0) {
```
Hjdskes: What about `r == 0`, which [may happen](https://github.com/systemd/systemd/blob/dd8352659c9428b196706d04399eec106a8917ed/src/libsystemd/sd-bus/bus-message.c#L3390)?

tools/ratbagd.py
```diff
@@ -467,6 +472,25 @@ def mapping(self, button):
...
+    @macro.setter
+    def macro(self, macro):
+        """Set the macro to the given macro. Note that the type must be one of
+        RatbagdButton.KEY_PRESS or RatbagdButton.KEY_RELEASE and the keycodes
+        must be as specified in linux/input.h.
+
+        @param macro A list of (type, keycode) tuples to form the new macro.
+        """
+        ret = self._dbus_call("SetMacro", "a(uu)", macro)
+        self._set_dbus_property("Macro", "a(uu)", macro, readwrite=False)
```
Hjdskes: If I'm not mistaken this should also be added to the mapping setters, as yesterday I'd been getting a bunch of [PropertyReadOnly](https://github.com/libratbag/piper/pull/47#issuecomment-317416953) exceptions.

tools/ratbagd.py
```diff
@@ -461,6 +498,25 @@ def mapping(self, button):
         return ret
 
     @GObject.Property
+    def macro(self):
+        """A list of (type, keycode) tuples that form the currently set macro,
+        where type is either RatbagdButton.KEY_PRESS or RatbagdButton.KEY_RELEASE
```
bentiss: nitpick, there should also be mention of RatbagdButton.MACRO_WAIT

tools/ratbagd.py
```diff
@@ -461,6 +498,25 @@ def mapping(self, button):
...
+        and the keycode is as specified in linux/input.h."""
+        return self._get_dbus_property("Macro")
+
+    @macro.setter
+    def macro(self, macro):
+        """Set the macro to the given macro. Note that the type must be one of
+        RatbagdButton.KEY_PRESS or RatbagdButton.KEY_RELEASE and the keycodes
```
bentiss: ditto :)

ratbagd/ratbagd-button.c
bentiss: Looking at https://github.com/systemd/systemd/blob/master/src/run/run.c -> it seems that it is a pretty common way of exiting in case of errors...

tools/ratbagd.py
bentiss: good point. I'll add those in a separate commit.

bentiss:
I have used my super powers and pushed the PR. Closing now.

PR #249, opened 2017-07-25
implement resolution active set

### LINKED ISSUE #80: Add an API to switch between resolutions
Currently, libratbag only allows to change the value of a resolution within a profile. When the profile supports several resolutions that can be toggled, libratbag doesn't export the required API to change between those resolutions.

bentiss:
Fixes #80

Though it's only in ratbagd/ratbagctl, we need some work in the drivers (but I like to remove the skipped tests)

### DIFF
--- a/ratbagd/ratbagd-resolution.c
+++ b/ratbagd/ratbagd-resolution.c
@@ -79,6 +79,42 @@ ratbagd_resolution_set_report_rate(sd_bus *bus,
 	return r;
 }
 
+static int ratbagd_resolution_active_signal_cb(sd_bus *bus,
+						struct ratbagd_resolution *resolution)
+{
+	struct ratbag_resolution *lib_resolution = resolution->lib_resolution;
+
+	/* FIXME: we should cache is_active and only send the signal for
+	 * those resolutions where it changed */
+
+	(void) sd_bus_emit_signal(bus,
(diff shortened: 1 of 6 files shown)


ratbagd/ratbagd-resolution.c
```diff
@@ -79,6 +79,42 @@ ratbagd_resolution_set_report_rate(sd_bus *bus,
 	return r;
 }
 
+static int ratbagd_resolution_active_signal_cb(sd_bus *bus,
+						struct ratbagd_resolution *resolution)
+{
+	struct ratbag_resolution *lib_resolution = resolution->lib_resolution;
+
+	/* FIXME: we should cache is default and only send the signal for
```
Hjdskes: s/default/active/

tools/ratbagd.py
```diff
@@ -426,6 +426,10 @@ def set_default(self):
         """Set this resolution to be the default."""
         return self._dbus_call("SetDefault", "")
 
+    def set_active(self):
+        """Set this profile to be the active profile."""
```
Hjdskes: s/profile/resolution/

bentiss:
That's a lot of copy/paste for a 2 functions change :/

Thanks for the review. Amended and force pushed the patch.

PR #251, opened 2017-07-26
Start dbus in the container for being able to run the test suite

bentiss:
We don't need a full systemd service, just dbus-daemon to be started.


bentiss:
Actually, there is no need for a review on this one, let's have this right away so the tests are run from CircleCI

PR #253, opened 2017-07-26
Enumify the remaining string properties in ratbagd

### LINKED ISSUE #252: SetSpecialMapping in ratbagd should accept an enum instead of a string
I suppose this one was simply glanced over, but this function too should take an `enum ratbag_button_action_special` instead of a string. Right now Piper will crash over this (in https://github.com/libratbag/piper/pull/47), because it already assumes the enum while the bindings and ratbagd still want the string.

Hjdskes:
This fixes #252 and also makes the LED type return an enum value.

### DIFF
--- a/tools/ratbagd.py
+++ b/tools/ratbagd.py
@@ -592,6 +592,10 @@ class RatbagdButton(_RatbagdDBus):
 class RatbagdLed(_RatbagdDBus):
     """Represents a ratbagd led."""
 
+    TYPE_UNKNOWN = -1
+    TYPE_LOGO = 0
+    TYPE_SIDE = 1
+
     MODE_OFF = 0
     MODE_ON = 1
     MODE_CYCLE = 2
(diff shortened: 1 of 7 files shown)


bentiss:
Aren't you missing the ratbagd.py changes?
Also I'd like to have the ratbagctl / ratbagctl.test changes in the same PR, but no worries if you left these ones out.

Hjdskes:
> Aren't you missing the ratbagd.py changes?

Whoops, will add them in a bit. I'll have a look at ratbagctl too.

Hjdskes:
Just amended my commits with the updates to ratbagd.py, ratbagctl and the tests. The LED tests are still failing but I'm not sure why, I intend to look into it more but I have to go soon so feel free to take over if you want.

tools/ratbagd.py
```diff
@@ -440,6 +440,26 @@ class RatbagdButton(_RatbagdDBus):
     ACTION_TYPE_KEY = 3
     ACTION_TYPE_MACRO = 4
 
+    ACTION_SPECIAL_UNKNOWN = (1 << 30),
```
bentiss: I'll amend this locally and push, but you forgot the comma at the end of the line, which is a bug you did not see because driver-test.c was not picking up specials from the config input.

PR #260, opened 2017-07-29
make sure the dbus daemon is available before starting the test suite

bentiss:
On CircleCI, the builds fail one over four times because we are launching
the testsuite while ratbagd.devel is still not ready.

Do a basic polling after we launch the daemon so that we are aware when
the ratbagd daemon is ready.


bentiss:
Got the ACk from @whot sitting right to me.

PR #263, opened 2017-07-30
RatbagdProfile: emit ActiveProfileChanged signal on the device

Hjdskes:
This allows Piper to listen for profile changes using a clean approach based on this one signal.


whot:
sorry, misunderstood your (private) email. Not a fan of having this in the dbus interface, please make this a signal within the bindings themselves, the DBus interface has `IsActive` for the profile, that should be enough

Hjdskes:
Ah, apologies. I should've been more clear about that. I'll change it there, closing.

PR #264, opened 2017-07-31
Emit PropertyChanged signals for AcitonType when setting button mappings

Hjdskes:
Internally, setting mappings changes the action type but this isn't reflected on DBus because ratbagd doesn't emit the PropertyChanged signal for them.


Hjdskes:
Changes to the bindings will be done in Piper (in the buttons PR), which I'll sync with libratbag once the buttons PR is merged.

PR #273, opened 2017-08-02
doc: drop documentation

whot:
Our DBus interface is the only API now and libratbag is just the internal
library. Let's not build doxygen for it, there's no need.


Hjdskes:
Should we then also remove https://libratbag.github.io/libratbag/ ?

whot:
> Should we then also remove https://libratbag.github.io/libratbag/ ?

yep, good point, thanks. all done

PR #284, opened 2017-08-16
RatbagdResolution: emit properties changed from signal callbacks

Hjdskes:
This allows clients to respond to changes in the active resolution. It's not the cleanest approach, but it works. To clean it up, we could cache the changed resolutions and make sure to only emit the signal on those, just as the FIXME already notes.

References https://github.com/libratbag/piper/pull/132.

### DIFF
--- a/ratbagd/ratbagd-resolution.c
+++ b/ratbagd/ratbagd-resolution.c
@@ -94,6 +94,12 @@ static int ratbagd_resolution_active_signal_cb(sd_bus *bus,
 				  "b",
 				  ratbag_resolution_is_active(lib_resolution));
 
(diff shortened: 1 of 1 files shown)


Hjdskes:
I'm attempting to do the same for `ratbagd_profile_set_active`, but I'm starting to get some [`No active profile. Please report this bug to the libratbag developers`](https://github.com/libratbag/piper/blob/master/piper/ratbagd.py#L327) from Piper with that change... haven't figured out why yet. It's getting late so I'll continue tomorrow; feel free to merge this already, I can just open a new PR once I get that one working too.

PR #290, opened 2017-08-21
RatbagdButton: make mapping properties writable

Hjdskes:
This brings RatbagdButton in line with the other Ratbagd* objects, and
more importantly solves https://github.com/libratbag/piper/issues/140.


Hjdskes:
After this and https://github.com/libratbag/piper/pull/148 are merged, I will sync the bindings between libratbag and Piper.

PR #296, opened 2017-08-24
Add async commit support

### LINKED ISSUE #269: Implement async commit in ratbagd
We get timeouts on commit and we should be using async anyway. So client calls commit which finishes immediately and then gets a "yay, I'm done, let's go party" dbus signal when the commit is complete.

whot:
Defer the actual commit so we can return immediately. This fixes the various
timeout issues we've seen on some devices where a commit can take several
seconds.

This commit merely adds the async hook. If the commit itself fails we
should re-sync from libratbag and update all the properties. This is currently
missing.

Fixes #269

Note: this does **NOT** include the ratbagd.py changes because we currently have some issues with syncing that from piper (see #295). There's a piper-commit pending for adding it there, it will be synced back once that's merged.

### DIFF
--- a/DBusInterface.md
+++ b/DBusInterface.md
@@ -89,7 +89,9 @@ Provides the list of profile paths for all profiles on this device, see
 #### `Commit() → ()`
 
 Commits the changes to the device. Changes to the device are batched; they
-are not written to the hardware until `Commit()` is invoked.
+are not written to the hardware until `Commit()` is invoked. This method is
+**asynchronous** and always succeeds. When the commit has been completed, a
+**Committed** signal is emitted.
 
 #### `GetSvg(s) → (s)`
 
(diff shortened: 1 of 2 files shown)


bentiss:
Damned, I put a comment in #296, but it looks like my firefox screwed up.

I don't think we need that, for the simple reason that it is a client side issue IMO. The problem lies in the fact that piper is always using `proxy.call_sync()` while here, for `commit()` it should use the `proxy.call()` which is async. You'll get the benefit of both this PR and the next one (#297) without having to touch ratbagd.

Nacking for now to prevent anybody else to ack and merge this.

Hjdskes:
I agree with @bentiss; if we can solve this by only changing the bindings that's the preferred solution. I'm writing my final report for GSoC at the moment so I'm a little occupied here but I'll do some testing when that's done.

whot:
closing, will reopen if we can't work around it

PR #302, opened 2017-08-28
Switch from a HWDB to an ini file device database

### LINKED ISSUE #268: Drop udev hwdb, switch to .ini files
After switching to the commit-style API the biggest issue we have now is that we don't get errors until we write out the profile. So we need to be able to catch any errors before querying the device and we need to figure out what capabilities devices have so that the clients can know what to set.

Solution: a hardware database. But since we'll now need a bunch of extra info the udev hwdb is getting overloaded and won't fit the bill anymore. So let's switch to a libwacom-style database with a bunch of .ini files where we can add the info we need over time.

the udev hwdb currently is in use only because it was easy to add with a few lines of code. There's no other specific reason for it. And, coincidentally, dropping it may make it possible to run ratbagd in a container/flatpak/etc.

whot:
Much more flexible and easier to update too. Especially since we re-read the files on every device connect which makes adding new devices even easier (drop the file, plug in the device, done).

This is still relatively early stages, but I find myself skimming over everything when I look at it again...

Fixes #268


Hjdskes:
Does this also mean we can start looking at flatpaks again?

whot:
not really, we still need ratbagd to discover the devices and that's handled via udev. No easy way to get rid of those bits, afaik

Issue #300, opened 2017-08-27
No image for device


kudlatyamroth:
as suggested by @Hjdskes in https://github.com/libratbag/piper/issues/161
i opened this bug.

When i launch piper i get error with message that i don't have image for my device
"Device has no image or its path is invalid"

My device is Logitech G903 (same as G900)
My system is Arch Linux

piper version: 0.2.900-3
libratbag version: 0.9.900-2

udevadm output:

> [root@arch kudlaty]# udevadm info /sys/class/input/event2
> P: /devices/pci0000:00/0000:00:1d.0/usb2/2-1/2-1.5/2-1.5.2/2-1.5.2:1.1/0003:046D:C539.0004/input/input19/event2
> N: input/event2
> S: input/by-id/usb-Logitech_USB_Receiver-if01-event-mouse
> S: input/by-path/pci-0000:00:1d.0-usb-0:1.5.2:1.1-event-mouse
> E: DEVLINKS=/dev/input/by-path/pci-0000:00:1d.0-usb-0:1.5.2:1.1-event-mouse /dev/input/by-id/usb-Logitech_USB_Receiver-if01-event-mouse
> E: DEVNAME=/dev/input/event2
> E: DEVPATH=/devices/pci0000:00/0000:00:1d.0/usb2/2-1/2-1.5/2-1.5.2/2-1.5.2:1.1/0003:046D:C539.0004/input/input19/event2
> E: ID_BUS=usb
> E: ID_INPUT=1
> E: ID_INPUT_KEY=1
> E: ID_INPUT_MOUSE=1
> E: ID_MODEL=USB_Receiver
> E: ID_MODEL_ENC=USB\x20Receiver
> E: ID_MODEL_ID=c539
> E: ID_PATH=pci-0000:00:1d.0-usb-0:1.5.2:1.1
> E: ID_PATH_TAG=pci-0000_00_1d_0-usb-0_1_5_2_1_1
> E: ID_REVISION=3904
> E: ID_SERIAL=Logitech_USB_Receiver
> E: ID_TYPE=hid
> E: ID_USB_DRIVER=usbhid
> E: ID_USB_INTERFACES=:030101:030102:030000:
> E: ID_USB_INTERFACE_NUM=01
> E: ID_VENDOR=Logitech
> E: ID_VENDOR_ENC=Logitech
> E: ID_VENDOR_ID=046d
[...]

Issue #213, opened 2017-07-07
Devices missing for the GNOME theme


Hjdskes:
Devices converted:

- [ ] etekcity
- [x] logitech-g300
- [x] logitech-g303
- [x] logitech-g403
- [x] logitech-g500
- [x] logitech-g500s
- [x] logitech-g502
- [x] logitech-g700
- [x] logitech-g9
- [x] logitech-g900
- [x] logitech-mx_master
- [x] roccat-kone-xtd

Hjdskes:
@jimmac Feel free to do more, requirements are in [the GNOME guidelines](https://github.com/libratbag/libratbag/blob/master/data/gnome/README.md).

whot:
https://github.com/libratbag/libratbag/pull/228 has the g303

Hjdskes:
Roccat Kone in #294.

phomes:
I am working on G300 in https://github.com/phomes/libratbag/tree/WIP-G300-svg

phomes:
G9 is completed

whot:
I think we can just remove the default theme and symlink it to the gnome theme. We only promise the "default" theme exists, not how it looks like.

And this is git, we can git rm the svgs and resurrect them if we ever need them again. tbh, I have no idea if anyone is using the default theme anyway.

whot:
Closing this. SVGs have moved to Piper anyway so any missing SVGs should be handled there.

PR #303, opened 2017-08-28
Add Logitech G900 SVG for gnome theme

cvuchener:
Related issues: #300, #213.

@kudlatyamroth or @kaymio please test this.

### DIFF
--- a/data/gnome/logitech-g900.svg
+++ b/data/gnome/logitech-g900.svg
@@ -1,0 +1,554 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<!-- Created with Inkscape (http://www.inkscape.org/) -->
+
+<svg
+   xmlns:dc="http://purl.org/dc/elements/1.1/"
+   xmlns:cc="http://creativecommons.org/ns#"
+   xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#"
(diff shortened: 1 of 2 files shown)


bentiss:
Thanks for the SVG.

Couple of issues:
- you are missing:
```
diff --git a/meson.build b/meson.build
index c4f104b..4bb8755 100644
--- a/meson.build
+++ b/meson.build
@@ -280,6 +280,7 @@ gnome_svg_files = [
        'data/gnome/logitech-g500s.svg',
        'data/gnome/logitech-g502.svg',
        'data/gnome/logitech-g700.svg',
+       'data/gnome/logitech-g900.svg',
        'data/gnome/roccat-kone-xtd.svg',
 ]
 install_data(gnome_svg_files,
```
- all the texts that are supposed to be shown on right are aligned from the left (so they are overlapping with the leader instead of moving out from it)

bentiss:
And some more information:
- you need to use `style="text-align:start...` instead of `style="text-align:end...`
- the LED are inverted, and you can not really configure the LED that show the DPI/profile (right under the resolution up/down buttons). So I'd remove the second LED and keep only the logo one as LED-0

cvuchener:
I knew I forgot something. Is it good now?

cvuchener:
I missed your message about the LEDs. I changed that too, now.

PR #304, opened 2017-08-28
Make valgrind an optional dependency

sanyarajan:
Since valgrind is only used for tests it can made an optional dependency if no tests are going to be run. i.e. when it is being installed.
See https://aur.archlinux.org/packages/piper/ and https://aur.archlinux.org/packages/libratbag/ for examples of users who have reported issues.


Hjdskes:
To build without tests use `-Denable-tests=false` on the meson command. This then doesn't require valgrind as a dependency.

sanyarajan:
Cool, that works. Thank you.

PR #306, opened 2017-08-29
data: add a fallback.svg file

whot:
This SVG is generic enough to cover 7-button mice with up to 3 LEDs that are
supported but don't have an SVG. The SVG style is intentionally NOT a generic
mouse to make it obvious that something isn't right here. If we use a generic
mouse, people may not put the effort into a proper SVG style because the
generic mouse is good enough.

The fallback is currently only used for test devices

### DIFF
--- a/data/gnome/fallback.svg
+++ b/data/gnome/fallback.svg
@@ -1,0 +1,506 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<!-- Created with Inkscape (http://www.inkscape.org/) -->
+
+<svg
+   xmlns:dc="http://purl.org/dc/elements/1.1/"
+   xmlns:cc="http://creativecommons.org/ns#"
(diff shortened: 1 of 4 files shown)


Hjdskes:
Hah, I love it. I'm merging this now, we can discuss the default SVG in #307.

whot:
Note: updated with the commit to always supply the fallback.svg from ratbagd if libratbag returns NULL.

bentiss:
Looks like you forgot to remove the fallback on the test-device (at least, this is what written on the commit)

whot:
oops, fixed and force-pushed. that was an old commit message, removing the fallback in the test device gives us an error every time on startup which isn't pretty

Hjdskes:
@bentiss things look good to me, leaving this one up to you to merge.

whot:
ask for forgiveness, not for permission. so I'm merging this now :)

PR #308, opened 2017-08-29
Add an SVG parsing tool

whot:


data/gnome/check-svg.py
```diff
@@ -0,0 +1,99 @@
...
+
+logging.basicConfig(level=logging.DEBUG)
+logger = None
+success = True
+
+
+def check_size(root):
+    width = float(root.attrib['width'])
+    height = float(root.attrib['height'])
+    if width < 400 or width > 500:
```
Hjdskes: Just FYI: Python supports `400 < width < 500` as well. Just a matter of preference really, but this can alternatively be `if not 400 < width < 500`

data/gnome/check-svg.py
```diff
@@ -0,0 +1,99 @@
...
+
+    logger.info("Found {} {}s".format(highest + 1, prefix))
+
+
+def check_leds(root):
+    check_elements(root, "led")
+
+
+def check_buttons(root):
+    check_elements(root, "button", 3)
```
Hjdskes: I might be missing the obvious but where does the 3 come from? Do we require at least 3 buttons, and if there are more that are missing we'll find out from bug reports but Piper won't crash because of it?

Hjdskes:
Also, how do you decide whether to log a warning or an error with `success = False`?

Hjdskes:
You can find commits implementing bullets 1 and 2 in my branch: https://github.com/Hjdskes/libratbag/commits/whot-wip/svg-parsing-tool. I've also fixed the G502 with a style property, and made a missing leader an error instead.

That leaves bullet 3, which admittedly isn't that important because Piper will still function without it. I also noticed that the script outputs errors for the roccat-kone-xtd's wheel tilt buttons, because they're missing their leaders.

data/gnome/check-svg.py
whot: magic assumption. all configurable mice should have at least 3 buttons, right? so those are the ones we can test for properly, anything else is more hardware-specific

data/gnome/check-svg.py
whot: ooh, nice. fixed locally, will be in the force-push

whot:
> , how do you decide whether to log a warning or an error

something that looks wrong but where we can't be sure is a warning. something that's definitely wrong is an error.

> A missing leader should be an error because without it

how are we doing this for the wheel tilt buttons?

whot:
@Hjdskes I merged your patches and addressed the comments I had regarding the 502. This won't pass the check suite, we need to figure out what we do for the leader lines on wheel tilt buttons

Hjdskes:
> how are we doing this for the wheel tilt buttons?

Wheel tilt buttons require these elements just as any other button does. I've force pushed to my [branch](https://github.com/Hjdskes/libratbag/commits/whot-wip/svg-parsing-tool) (78f72c4a5298ccb9686bda36fe70f540b639f745), making sure not to include detritus this time (though there are some thing I'm not sure are required, feel free to remove those if they aren't) and putting it before the commit that introduces the script.

whot:
@Hjdskes  thanks, cherry-picked from your branch, reshuffled a few things. ended up squashing all the commits regarding the `check-svg.py` into one - no need to have more than one here.

Issue #282, opened 2017-08-15
[hidpp10] Issue when reading the current profile index.


cvuchener:
Testing the G500 and G500s with multiple profiles, I have several issues regarding CMD_PROFILE and the default profile. Since I am not sure this is due to a misunderstanding of the protocol or the G500(s) not supporting multiple profiles properly. I'd like to have feedback from owner of devices that should support it: G9, G9x, G700, G700s.

When turning the mouse on, the G500 loads the profile from page 2, the G500s loads the profile from page 5, **disregarding whatever is in the profile directory on page 1**. The CMD_PROFILE register reads 0 on the G500, 3 on the G500s (so profiles are counted from page 2 on both). Buttons bound to profile switching actions don't work.

If the CMD_PROFILE register is written at least once, the mouse reads profiles according to the profile directory and profile switching actions work.

I don't know any way to distinguish between these two modes.

cvuchener:
Testing with a G500s with this profile directory:

    00000000: 05 00 01 06 00 01 ff ff ff ff ff ff ff ff ff ff  ................

Just after plugging the mouse, it uses the first profile (on page 5). The buttons bound to profile switching do not work.

Reading the current profile:

    [debug:register] Getting short register 0x0f
    [debug:report] Send HID report: 10 00 81 0f 00 00 00
    [debug:report] Recv HID report: 10 00 81 0f 00 03 00
    [debug:register] Results: 00 03 00

After loading a profile index 0 from the directory:

    [debug:register] Setting short register 0x0f
    [debug:register] Parameters: 00 00 00
    [debug:report] Send HID report: 10 00 80 0f 00 00 00
    [debug:report] Recv HID report: 10 00 80 0f 00 00 00

Reading the current profile changed accordingly:

    [debug:register] Getting short register 0x0f
    [debug:report] Send HID report: 10 00 81 0f 00 00 00
    [debug:report] Recv HID report: 10 00 81 0f 00 00 00
    [debug:register] Results: 00 00 00

Pressing the button now works and after switching the profile, the correct index is returned:

    [debug:register] Getting short register 0x0f
    [debug:report] Send HID report: 10 00 81 0f 00 00 00
    [debug:report] Recv HID report: 10 00 81 0f 00 01 00
    [debug:register] Results: 00 01 00

Switching to a profile from its address (is that the second behavior?), do work:

    [debug:register] Setting short register 0x0f
    [debug:register] Parameters: 01 05 00
[...]

Issue #261, opened 2017-07-29
G500s no profile set as active profile


adrien-martin:
Hello,

I'm facing a strange problem, ratbag-command seems to detect my mouse (a G500s) but can not print information or configure anything.

```
# ratbag-command --verbose=debug list
ratbag debug: Logitech G500s Laser Gaming Mouse is device '/dev/hidraw0'.
ratbag debug: Logitech G500s Laser Gaming Mouse is device '/dev/hidraw5'.
ratbag debug: report ID 01
ratbag debug: report ID 03
ratbag debug: report ID 10
ratbag debug: report ID 11
ratbag error: unknown tag: 0x12
ratbag error: unknown tag: 0x08
ratbag error: unknown tag: 0x12
ratbag error: unknown tag: 0x08
ratbag error: libratbag bug: Logitech G500s Laser Gaming Mouse: no profile set as active profile
No supported devices found
```

```
# udevadm info /sys/class/input/event0 | grep RATBAG_DRIVER
E: RATBAG_DRIVER=hidpp10
```

The libratbag version is 0.9.0 provided by debian (sid), but the only patch I found was for the doc files.

According to the error message, setting a profile could fix the problem but I don't know how to change it without ratbag.

cvuchener:
I can reproduce the bug with my current installed version of ratbag-command, but it is not a debug build. I made a debug build from the latest revision, and it looks like it has already been fixed. Could you try with the current development version?

------------

When looking at this, I also discovered that there may be some problem with the way we get the current active profile for HID++1.0 devices.

Using a G500s with three profiles, I always get "3" as the current profile after plugging in the mouse and changing the profile using the buttons. If I change the profile by sending a command to the device, the current profile is then read properly, even if I later change the profile using the mouse buttons.

libratbag test if the value is out of bound and replace it with 0, so it should work but it may be lying about the current profile (if there are more than one profile). I don't know any way to improve the current behavior.

adrien-martin:
I built from master and it works too, sorry I should have tried before.

For the profiles I don't know, mine seems to have only one profile (says "Profiles supported: 1", and verbose=raw don't find more).

cvuchener:
Officially, the G500 and G500s don't support multiple profiles. It is actually possible to have more than one profile but it does not work perfectly. It is best to use those mice with only one profile, anyway. It is just that I thought it was the cause of the bug at some point.

adrien-martin:
Ah interesting, could be useful some day.

[...]

PR #310, opened 2017-08-30
improve error for missing device files

phomes:
Before 340e5b7d86fc9cc04dc563dc755d1efffe9a97c5 I got:
ratbag error: libratbag bug: Logitech G9 Laser Mouse: no profile set as active profile
No supported devices found

After 340e5b7d86fc9cc04dc563dc755d1efffe9a97c5 I get:
ratbag error: Unable to locate device files: No such file or directory
ratbag error: Unable to locate device files: No such file or directory
ratbag error: Unable to locate device files: No such file or directory
No supported devices found

With this PR we give a hint to the user about where libratbag is looking for the files.

### DIFF
--- a/src/libratbag-data.c
+++ b/src/libratbag-data.c
@@ -326,8 +326,8 @@ ratbag_device_data_new_for_id(struct ratbag *ratbag, const struct input_id *id)
 
 	n = scandir(LIBRATBAG_DATA_DIR, &files, filter_device_files, alphasort);
 	if (n <= 0) {
-		log_error(ratbag, "Unable to locate device files: %s\n",
-			  n == 0 ? "No files found" : strerror(errno));
+		log_error(ratbag, "Unable to locate device files in %s: %s\n",
+			  LIBRATBAG_DATA_DIR, n == 0 ? "No files found" : strerror(errno));
 		return NULL;
 	}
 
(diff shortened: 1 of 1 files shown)


phomes:
The output is from "./ratbag-command list"

Using "./ratbagctl.devel list" gives:
ratbag error: Unable to locate device files in /usr/local/share/libratbag: No such file or directory
ratbag error: Unable to locate device files in /usr/local/share/libratbag: No such file or directory
ratbag error: Unable to locate device files in /usr/local/share/libratbag: No such file or directory
ratbag error: Unable to locate device files in /usr/local/share/libratbag: No such file or directory
test_device: Test device

phomes:
btw I am looking into the problem with "no profile set as active profile". It regressed since v0.9

cvuchener:
> btw I am looking into the problem with "no profile set as active profile". It regressed since v0.9

Are you talking about #261? By the way, if you are looking into the G9 profiles, could you have a look at #282?

phomes:
I bisected it to 2c799eca30a39459e7cd5fd10ee81995d0a8bbcc and then noticed that I was not using an updated hwdb. My mouse was missing the value for RATBAG_HIDPP10_PROFILE_COUNT.

I will take a look at #282. I just need to get things working again with the symlinks etc.

PR #320, opened 2017-09-01
ratbagd: split the test device out into a seprate file

whot:
It always bothered me that it was in the middle of the normal ratbagd, which
made fiding it more effort than neccessary. Split it out into a separate file
so it's easily identifiable.


ratbagd/ratbagd.c
```diff
@@ -39,6 +39,8 @@
 #include "shared-macro.h"
 #include "libratbag-util.h"
 
+#include "ratbagd-test.h"
```
Hjdskes: Why not do the `ifdef` here?

whot: we still need the `ratbagd_init_test_device` unconditionally (though that can probably be worked around as well. Plus, I'm not a big fan of ifdef'd includes

PR #321, opened 2017-09-01
set LANG to UTF-8 in circleci

whot:


bentiss:
Thanks!
I tried yesterday to use the web interface to globally set en_US.UTF-8 as environment variable, but it looks like you found the expected encoding from Ubuntu.

PR #322, opened 2017-09-01
Drop key modifiers from action type Key

whot:
Originally I thought we could transparently wrap this in macros inside libratbag, but that's a bit more work. So meanwhile, let's just drop the modifiers so we're at least a step closer to sanity.

I'm not 100% sure anymore whether we should really map this to macros, it seems that the keys have quite a fair bit of limitation (i.e. what key can be mapped) which macros don't have, so maybe keeping it separate is better here.

### DIFF
--- a/DBusInterface.md
+++ b/DBusInterface.md
@@ -238,10 +238,9 @@ uint of the current button mapping (if mapping to button)
 
 Enum describing the current special mapping (if mapped to special)
 #### `KeyMapping`
-- type: `au`, read-write, mutable
+- type: `u`, read-write, mutable
 
-Array of uints, first entry is the keycode, other entries, if any, are
-modifiers (if mapped to key)
+uint of the keycode in linux/input.h form (if mapping to key)
 #### `Macro`
 -type: `a(uu)`, read-write, mutable
 
--- a/circle.yml
+++ b/circle.yml
@@ -1,6 +1,8 @@
 libratbag_references:
   default_settings: &default_settings
     working_directory: ~/libratbag
+    environment:
+      LANG: C.UTF-8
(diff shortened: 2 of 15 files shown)


Hjdskes:
If I understand this correctly, this now means that a physical button can be mapped to a single key press (and release)? I'm not sure how useful that is, and this use-case certainly can be captured my a simple macro as well. Perhaps, in that case, we should remove it altogether and **only** offer macros; are there any devices that do not support macros but do support key presses?

In any case, you should drop 40c0d2f5d53ecbe85c5eb2cda8e7396319b7d23f and update `tools/ratbagd.py` as well. The rest of the changes look good to me.

(Also, how are we doing on dropping `ratbag-command`? Just wondering since you're now basically required to update a deprecated utility.)

bentiss:
I am not a big fan of this PR either. I think the plan was to remove it all together and let the _drivers_ handle the transformation between macro and key event.

> are there any devices that do not support macros but do support key presses?

There are. Any Logitech non gaming device that support _redirecting_ can do that (theoretically). Some, however require the redirection to be handled in software, so we can put whatever we want (and emulate macros).

> (Also, how are we doing on dropping ratbag-command? Just wondering since you're now basically required to update a deprecated utility.)

For users, `ratbag-command` is deprecated, yes. For developers, it's still have interesting features that `rabtagctl.devel` doesn't. For instance, looking at the raw events from/to the device is much easier with `ratbag-command`. Also, using gdb with `ratbag-command` is easier. We talked about it at GUADEC with @whot, but I'd rather be certain that `ratbag-command` is not useful to anybody before ditching it.

whot:
Closing, see #329

Issue #280, opened 2017-08-14
Logitech G502 Proteus Core ERR_INVALID_ADDRESS


PlasmaPower:
Whenever I try to "apply" in piper, it says "Failed to commit changes". I also get this error from ratbagd: `ratbag error:     HID++ error from the device (255): ERR_INVALID_ADDRESS (02)`

May be related to #24, not sure.

PlasmaPower:
Not sure if this is related, but it seems the LED capabilities are also incorrectly detected.

cvuchener:
Which version are using (v0.9 or latest git)?

Can you reproduce the issue with ratbag-command? If you do, post the output using the `--verbose=raw` option.

> Not sure if this is related, but it seems the LED capabilities are also incorrectly detected.

How exactly are they incorrectly detected? libratbag's logitech driver only support RGB LED effects. It is expected if no LED capabilities are reported.

PlasmaPower:
@cvuchener I'm using latest git. I get the same error with ratbag-command. Here's the output of `sudo ./ratbag-command --verbose=raw button 5 action set button 6 /dev/input/event4`: https://gist.github.com/PlasmaPower/4beaea7976caf169ffd92dabbf5cd2c9

Two LEDs are detected, but I don't believe either are configurable, and I certainly don't think either are RGB. I think that's only in the spectrum version of the mouse. I only have the core version.

cvuchener:
I don't know why there is a "invalid address" error. The writing looks good to me. But the profile data seems to be corrupted: all profiles are disabled and profiles data is offset by 16 bytes. Did libratbag do that? Did you modify the profiles in another way before?

PlasmaPower:
@cvuchener I've never modified this mouse before now. The profiles are all stock (and they work fine). Libratbag only showed the currently enabled profile, and I didn't try to add a new profile or anything likke that. Maybe I should use the official Windows tool first, to setup the profiles?

amfern:
i tested it on same mouse type g502 without RGB feature too, and it worked.
`sudo ./ratbag-command --verbose=raw button 5 action set button 6 /dev/input/event2:`
https://gist.github.com/amfern/aac977f1da47916b60e742f15e96952e

PlasmaPower:
Interesting. @amfern Have you setup your profiles with the Windows tool?

amfern:
yes the profiles were changed from windows with logitech drivers before i tried with libratbag

PlasmaPower:
[...]

Issue #122, opened 2017-01-31
[hidpp20] Some devices have different sector size.


cvuchener:
The G402 has bigger sectors/pages with 1024 bytes. But libratbag uses hard-coded `HIDPP20_PAGE_SIZE` of 256 bytes instead of the size from `hidpp20_onboard_profiles_info`. Thus writing the CRC at an invalid location.

Parameters for `CMD_ONBOARD_PROFILES_MEMORY_READ` and `CMD_ONBOARD_PROFILES_MEMORY_ADDR_WRITE` need fixing for supporting bigger sector. `section` is actually 16 bits and in big endian in bytes 2 and 3. And bytes 4 and 5 of `CMD_ONBOARD_PROFILES_MEMORY_ADDR_WRITE` is the data length as 16 bits big endian (currently hard-coded to 0x0100 = 256 bytes).

I cannot test the G402 myself, it is a supposed bug. I found this from dumps from @dslul. https://github.com/GSeriesDev/gseries-tools/issues/3#issuecomment-276332598

PR #331, opened 2017-09-07
hidpp20 memory rework

bentiss:
Fixes both #280 and #122 (hopefully).

This is the first step required before we are finally able to write macros on the device.
The "new" memory API only allows full sectors to be read/written. This ensures we always write the crc and check for it.

Also, now we better detect if the device is using factory profiles or not (see #280).

I still haven't yet implemented caching for the memory. This will be required for macros and to speed up a little bit the parsing of the device.

Please test!

Note: I also quickly wrote a tool to reset the memory to 0x00. This is hopefully the factory default (can't remember if this is true or not).

### DIFF
--- a/src/hidpp20.c
+++ b/src/hidpp20.c
@@ -1473,7 +1473,7 @@ hidpp20_onboard_profiles_allocate(struct hidpp20_device *device,
 
 	profiles->num_profiles = info.profile_count;
 	profiles->num_rom_profiles = info.profile_count_oob;
-	profiles->num_buttons = info.button_count <= 16 ? info.button_count : 16;
+	profiles->num_buttons = min(info.button_count, 16);
 	profiles->num_modes = HIDPP20_DPI_COUNT;
 	profiles->num_leds = HIDPP20_LED_COUNT;
 	profiles->has_g_shift = (info.mechanical_layout & 0x03) == 2;
--- a/src/hidpp20.c
+++ b/src/hidpp20.c
@@ -339,7 +339,7 @@ hidpp20_feature_set_get(struct hidpp20_device *device)
 	if (!feature_count)
 		return -ENOTSUP;
 
-	flist = zalloc(feature_count * sizeof(struct hidpp20_feature));
+	flist = zalloc((feature_count + 1) * sizeof(struct hidpp20_feature));
 
 	/* feature set count does not include the root feature as documented here:
 	 * https://6xq.net/git/lars/lshidpp.git/plain/doc/logitech_hidpp_2.0_specification_draft_2012-06-04.pdf
--- a/src/hidpp20.c
+++ b/src/hidpp20.c
@@ -1132,6 +1132,12 @@ hidpp20_onboard_profiles_read_sector(struct hidpp20_device *device,
 	count = sector_size;
 
 	for (offset = 0; offset < sector_size; offset += 16) {
+		/*
+		 * the firmware replies with an ERR_INVALID_ARGUMENT error
+		 * if we try to read past sector_size - 16, so when we are left with
+		 * less than 16 bytes to read we need to read from sector_size - 16
+		 */
+		offset = (sector_size - offset < 16) ? sector_size - 16 : offset;
 		hidpp_set_unaligned_be_u16(&msg.msg.parameters[2], offset);
 		buf = msg;
 		rc = hidpp20_request_command(device, &buf);
--- a/src/hidpp20.c
+++ b/src/hidpp20.c
@@ -1941,13 +1941,16 @@ hidpp20_onboard_profiles_read_led(struct hidpp20_led *led,
 	led->brightness = brightness;
 }
 
-int hidpp20_onboard_profiles_read(struct hidpp20_device *device,
-				  unsigned int index,
-				  struct hidpp20_profiles *profiles_list)
+static int
+hidpp20_onboard_profiles_parse_profile(struct hidpp20_device *device,
+				       struct hidpp20_profiles *profiles_list,
+				       unsigned index,
+				       bool check_crc)
 {
 	union hidpp20_internal_profile *pdata;
-	_cleanup_free_ uint8_t *data = NULL;
 	struct hidpp20_profile *profile = &profiles_list->profiles[index];
+	uint16_t sector = index + 1;
+	_cleanup_free_ uint8_t *data = NULL;
 	unsigned i;
 	int rc;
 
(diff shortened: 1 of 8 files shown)


SonicFrog:
I've done some quick testing. 
This new memory model works fine with the G900 as far as I can tell.
There are still a few problems though:

- It doesn't handle sector sizes that are not multiple of 16 (as the G603 has)
- It fails to commit changes for the G502 as mentionned in #280 

I have a fix in the works for the first of these, and I can see what's the problem exactly with the G502 next week.

bentiss:
> It fails to commit changes for the G502 as mentionned in #280

If you happen to have access to a G502 in the stock factory settings, could you run `./build/hidpp20-dump-page /dev/hidrawX` with `hidrawX` being the hidraw node given from the dmesg?

This way, I can reflash the device into the factory default and test this.

Locally, if I reset the memory to 0x00 with the new nuclear tool, it doesn't export the error on my G502.

SonicFrog:
> could you run ./build/hidpp20-dump-page /dev/hidrawX
Looks like the flash is already all zeroes. I attached the ROM dump, should be the factory defaults.
[g502romdump.txt](https://github.com/libratbag/libratbag/files/1284993/g502romdump.txt)

bentiss:
> Looks like the flash is already all zeroes.

Then in that case I do not understand what is going on. It works for me if I reset the flash with 0.

The bug in #280 is that the firmware sees that the dictionary is invalid (error 02), which is true because we used to write all profiles as disabled. Now, at least locally, the dictionary should be a rewrite of the ROM one, in which all 3 profiles are enabled.

SonicFrog:
Well after writing the ROM content on the flash with a correct dictionary it seems to work fine... Dunno what happened there. I still get some ERR_INVALID_ARGUMENT from hidpp20-dump-page though.

src/hidpp20.c
```diff
@@ -2073,49 +2064,54 @@ int hidpp20_onboard_profiles_write(struct hidpp20_device *device,
...
-		return rc;
+		goto out;
 
-	if (rc != sizeof(pdata.data))
-		return -EIO;
-
-	return 0;
+ out:
+	free(data);
+	return rc < 0 ? rc : 0;
```
whot: these code paths seems like they could be improved using the `_cleanup_free_` helpers, but this isn't something I'll insist on

tools/ratbag-command.c
```diff
@@ -441,8 +441,10 @@ ratbag_cmd_info(const struct ratbag_cmd *cmd,
 		printf("  Profile %d (%s)%s\n", i,
 		       ratbag_profile_is_enabled(profile) ? "enabled" : "disabled",
 		       ratbag_profile_is_active(profile) ? " (active)" : "");
-		if (!ratbag_profile_is_enabled(profile))
+		if (!ratbag_profile_is_enabled(profile)) {
+			ratbag_profile_unref(profile);
 			continue;
```
whot: yes, but moving the profile to inside the for loop and making it a `_cleanup_` seems like it's the better way to fix this

src/hidpp20.c
```diff
@@ -2020,6 +1987,39 @@ int hidpp20_onboard_profiles_read(struct hidpp20_device *device,
...
+		uint8_t *d = data + 4 * i;
+
+		if (d[0] == 0xFF && d[1] == 0xFF)
+			break;
+
+		profiles->profiles[i].index = d[1];
+		profiles->profiles[i].enabled = !!d[2];
+	}
+
+	free(data);
```
whot: follow-up: use `_cleanup_free_` here

src/hidpp20.c
```diff
@@ -2179,6 +2128,29 @@ int hidpp20_onboard_profiles_write(struct hidpp20_device *device,
...
+		if (profile->enabled) {
+			rc = hidpp20_onboard_profiles_write_profile(device,
+								    profiles_list,
+								    i);
+			if (rc < 0)
+				return rc;
+		}
+	}
+
+	return hidpp20_onboard_profiles_write_dict(device, profiles_list);
```
whot: hmm, I wonder: what happens on partial failures here? either when two profiles write correctly and one doesn't, or when the profiles write but the dict fails?

tools/hidpp20-reset.c
```diff
@@ -0,0 +1,112 @@
...
+static void
+usage(void)
+{
+	printf("Usage: %s [sector] /dev/hidraw0\n", program_invocation_short_name);
+}
+
+int
+main(int argc, char **argv)
+{
+	_cleanup_close_ int fd = 0;
```
whot: should be -1, otherwise you're working with stdin here

src/hidpp20.c
```diff
@@ -1129,6 +1129,11 @@ hidpp20_onboard_profiles_read_sector(struct hidpp20_device *device,
 	count = sector_size;
 
 	for (offset = 0; offset < sector_size; offset += 16) {
+		/*
+		 * the firmware replies with an ERR_INVALID_ARGUMENT error
+		 * if we try to read past sector_size - 16, so when we are left with
+		 * less than 16 bytes to read we need to read from sector_size - 16
+		 */
```
whot: for the commit message: please use present tense, because you're describing what the patch does when applied. and a hidpp20 prefix please, so "hidpp20: add comment for clarification"

Also, can we just squash this one into the previous commit please?

src/hidpp20.c
bentiss: good question. Depending on where the error happens and what error is triggered, we might just have either part of the profiles working or the device in the factory mode (if the crcs are wrong).

I guess we should resync the state of the mouse at that point. Probably in a later commit.

tools/hidpp20-reset.c
bentiss: good catch

tools/ratbag-command.c
bentiss: I somehow missed that comment. Added an extra commit for that on top of this PR

src/hidpp20.c
```diff
@@ -2022,6 +1993,37 @@ int hidpp20_onboard_profiles_read(struct hidpp20_device *device,
...
+	_cleanup_free_ uint8_t *data = NULL;
+
+	assert(profiles);
+
+	data = hidpp20_onboard_profiles_allocate_sector(profiles);
+
+	rc = hidpp20_onboard_profiles_read_sector(device, 0, profiles->sector_size, data);
+	if (rc < 0) {
+		free(data);
+		return rc;
```
whot: `free(data)` needs to be removed here

tools/hidpp20-reset.c
```diff
@@ -0,0 +1,112 @@
...
+reset_sector(struct hidpp20_device *dev, uint16_t sector_size,  uint16_t sector)
+{
+	int rc = 0;
+	uint8_t *data = zalloc(sector_size);
+
+	/* this returns error 4: hardware error and is expected */
+	rc = hidpp20_onboard_profiles_write_sector(dev, sector, sector_size, data, false);
+	if (rc == 4) {
+		rc = 0;
+		goto out;
```
whot: that goto seems a bit unnecessary, and this is also a candidate for `_cleanup_` (but you can do that later)

SonicFrog:
Looks good to me! I've done some testing with a few more models and all seem to work fine!

bentiss:
Thanks both of you. I am fixing the two nitpicks and will merge in a few

Issue #316, opened 2017-08-31
Fix return values in ratbagd dbus methods


Hjdskes:
See https://github.com/libratbag/libratbag/pull/275#discussion_r136026896.

Hjdskes:
I've started working on this one, and replacing the `return r;` at the end of setter methods (i.e., upon success) to `return 0;` is a matter of a search and replace.

However, what values do we return in case of errors? I.e. in the following case, what do we return?

```
r = <some libratbag function>
if (r < 0)
    return r;

<continue in case of success and return 0>
```

Likewise, what do we do with instances of `return sd_bus_message_append` and the like?

whot:
What errors are we expecting here? afaict the only real error we can get are dbus errors - where we're screwed anyway - or programming errors when we try to append the wrong format or something. Those ones we can assert on.

bentiss:
Given that we already need to have a signal from ratbagd telling "you screwed, I can not commit". How about the dbus methods always return the dbus status (as @whot said, either "you screwed in the arguments" or "something went wrong at the dbus level"), and if the libratbag return code is not null, then we emit the "BOO, pointing finger at you, please resync" signal while returning success?

bentiss:
I guess we can close this one now?

Hjdskes:
Well, there's still what I mentioned in https://github.com/libratbag/libratbag/pull/336#issuecomment-328616146:

> Also, this PR currently only changes those functions where we'd return r upon error. It does not, for example, check for errors in ratbagd-led where we didn't do that before either. I can add this if required.

PR #336, opened 2017-09-08
Fix ratbagd return values in dbus methods

Hjdskes:
As discussed in #316. For now I went with @whot's suggestion of assertions, but we can easily change this to @bentiss' suggestion to instead emit a signal over DBus indicating an error *before* trying to commit and *then* being notified of failures.

There's also a bunch of FIXMEs where I think we could clamp values instead of returning; this might not be necessary for @bentiss' suggestion but I think it's good to do if we stick with the current approach.


ratbagd/ratbagd-profile.c
```diff
@@ -291,8 +291,7 @@ static int ratbagd_profile_set_active(sd_bus_message *m,
 		return r;
 
 	r = ratbag_profile_set_active(profile->lib_profile);
-	if (r < 0)
-		return r;
+	assert(r == 0);
```
bentiss: I am not sure I read @whot's comment like that in #316:
To me, the assert would be on the errors after the call of ratbag_*, not the error on the call of ratbag_*. Because here you assume the device will always play nice, and a disconnection for instance while you are calling the dbus method would make the whole ratbagd abort.

I think we should not assert on libratbag errors, but only for dbus like errors, which we already don't check as a matter of fact.

whot: yep, agree. anything that's a dbus error is a bug in ratbagd or a system failure we don't recover from. A ratbag error means something else is wrong and should trigger a signal for the client to re-sync.

Hjdskes:
Ah, got both your comments in #316 now. I think the new approach is what you're after.

> I think we should not assert on libratbag errors, but only for dbus like errors, which we already don't check as a matter of fact.

We do check for dbus errors, and return those error values over the bus. Internal (ratbag) errors are now not returned; we always return success and emit the `Resync` signal on the device in case of errors.

ratbagd/ratbagd-resolution.c
```diff
@@ -304,6 +269,41 @@ ratbagd_resolution_get_rate(sd_bus *bus,
...
+	struct ratbagd_resolution *resolution = userdata;
+	unsigned int rate;
+	int r;
+
+	r = sd_bus_message_read(m, "u", &rate);
+	if (r < 0)
+		return r;
+
+	/* basic sanity check */
+	if (rate > 5000 || rate % 100)
```
bentiss: Shouldn't this return a -EINVAL?

bentiss: Never mind, this is fixed in a following commit

ratbagd/ratbagd-resolution.c
```diff
@@ -102,14 +103,20 @@ static int ratbagd_resolution_set_active(sd_bus_message *m,
...
-	if (r < 0)
-		return r;
+	if (r < 0) {
+		sd_bus *bus = sd_bus_message_get_bus(m);
+		sd_bus_emit_signal(bus,
+				   ratbagd_device_get_path(resolution->device),
+				   RATBAGD_NAME_ROOT ".Device",
+				   "Resync",
+				   NULL);
+	}
```
whot: I'm gonna be a pain here but:
1. please squash the "fix return values" patches into one patch. They're the same thing, there's no point separating them by file barrier
2. add a `ratbagd_device_resync(device, m)` helper function and call that so you don't need the same boilerplate everywhere. This make it easier to do extra resync work in that function in the future

ratbagd/ratbagd-resolution.c
```diff
@@ -287,13 +252,13 @@ ratbagd_resolution_set_resolution(sd_bus *bus,
...
-			    sd_bus_message *reply,
-			    void *userdata,
-			    sd_bus_error *error)
+ratbagd_resolution_get_report_rate(sd_bus *bus,
+				   const char *path,
+				   const char *interface,
+				   const char *property,
+				   sd_bus_message *reply,
+				   void *userdata,
+				   sd_bus_error *error)
```
whot: This looks like it introduces some whitespace issues? Or is that just the github UI showing it in a weird way?

Hjdskes: Just GitHub showing it in a weird way; I had to align the parameters again due to adding `_report`.

ratbagd/ratbagd-resolution.c
Hjdskes: > I'm gonna be a pain here

That's what code reviews are for :) Take another look, it should be good now!

PR #340, opened 2017-09-12
SVG for G300 in GNOME theme

phomes:
A SVG for the G300 for GNOME theme. It works fine in the buttons view but only one of the leaders is shown on the resolutions view. Not sure why. check-svg.py does not complain.

I did not include the LED yet as I have not yet added it in the driver.


whot:
In the resolutions view, you should only see leaders to the buttons configured to change dpi, all others are ignored. Is that what you're seeing?

From a quick view of the svg in github: I think you're missing the 20px margin around the device itself. The leaders should go to the edge, the device should have 20px at the top and bottom.

phomes:
Ah yes, the one leader shown is for changing resolution. Thanks for clearing that up.

I will fix up the 20px margin.

phomes:
I pushed a commit for the margin

PR #341, opened 2017-09-13
hidpp20: store correct size for feature list

SonicFrog:
The feature discovery was corrected in a previous commit. 

However the code still stores the wrong size.


SonicFrog:
Should be better now!

PR #344, opened 2017-09-14
make sure we check for ratbagd_process returncode

bentiss:
It was accidentally put as a dead code in 6cfca9e34b5.

This can't go pass the last return:
```
try:
  foo
except:
  return None
else:
  return foo
```
It is also verified by the fact that there is no return at the end of the function, while we were expecting an object.

It's 2 lines of code, so we can just have them duplicated.


bentiss:
Actually, no, please disregard, I'll send a better solution (moving down the last return would be better).

PR #346, opened 2017-09-14
hidpp20: fix resolution count detection

### LINKED ISSUE #343: hidpp20: reporting only 1 supported DPI resolution for G603
I'm not sure about this but after some investigation it seems that libratbag incorrectly sets the number of configured resolution in a profile to the number of sensors in the mouse. 

https://github.com/libratbag/libratbag/blob/548a40dba677d1f5c45e86611b4dab3a070e7964/src/driver-hidpp20.c#L498

Removal of this line does not affect mice that were working with it, and fixes the one mouse I've encountered that is affected by this "bug"

SonicFrog:
This should fix #343 without impacting the mice without onboard profiles.

### DIFF
--- a/src/driver-hidpp20.c
+++ b/src/driver-hidpp20.c
@@ -495,7 +495,12 @@ hidpp20drv_read_resolution_dpi_2201(struct ratbag_device *device)
 		  drv_data->sensors[0].dpi_max);
 
 	drv_data->num_sensors = rc;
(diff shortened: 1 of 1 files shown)


SonicFrog:
Done!

bentiss:
Nearly! :)

The "fix" should be on its own line in the commit message, not in the title (github might not bother, but this is not how we did for others)

bentiss:
Looks good now :)

PR #353, opened 2017-09-19
Openrazer driver backend

whot:
This PR adds a driver backend to connect to the openrazer DBus daemon to configure devices.

Still missing:
- [ ] openrazer needs to switch to the system bus, see https://github.com/openrazer/openrazer/pull/380
- [ ] long-term openrazer needs some dbus changes so the code is more generic, see https://github.com/openrazer/openrazer/issues/381
- [ ] we cannot read LED colours. This isn't yet exposed by openrazer's dbus daemon
- [ ] device SVG, I'll do that when we can actually connect to openrazer's dbus daemon
- [ ] better error handling. Right now, it's mostly `assert`

So all in all a **WIP, do not merge** and I'll figure out the details when openrazer is in shape for us to use it.

### DIFF
--- a/data/devices/razer-deathadder-chroma.device
+++ b/data/devices/razer-deathadder-chroma.device
@@ -1,0 +1,5 @@
+[Device]
+Name=Razer Razer DeathAdder Chroma
+DeviceMatch=usb:1532:0043
+Svg=fallback.svg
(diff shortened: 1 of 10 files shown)


CoelacanthusHex:
Is there any progress?

PR #355, opened 2017-09-19
G300: fix writing multiple profiles

phomes:
Writing multiple changed profiles was broken as we would exit the
loop after the first profile write. We were returning the number
of bytes written and this was interpreted as an error.

Fix it my returning 0 when we have written the correct number of
bytes to the mouse.

### DIFF
--- a/src/driver-logitech-g300.c
+++ b/src/driver-logitech-g300.c
@@ -480,7 +480,13 @@ logitech_g300_write_profile(struct ratbag_profile *profile)
 				       HID_FEATURE_REPORT,
 				       HID_REQ_SET_REPORT);
 
-	return rc;
+	if (rc < (signed)sizeof(*report)) {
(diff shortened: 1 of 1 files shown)


src/driver-logitech-g300.c
```diff
@@ -480,7 +480,13 @@ logitech_g300_write_profile(struct ratbag_profile *profile)
 				       HID_FEATURE_REPORT,
 				       HID_REQ_SET_REPORT);
 
-	return rc;
+	if (rc < (signed)sizeof(*report)) {
+		log_error(device->ratbag,
```
whot: can we cast this to the type of rc please? which I guess is `int` or something

phomes:
I was using signed in another place in the driver so I pushed a separate cleanup commit to change both to int.

PR #357, opened 2017-09-21
Non-RGB LEDs support for Logitech mice

### LINKED ISSUE #325: Not working with the Logitech G302
First I manually changed the USB ID for the G303 to match the G302 and after that it started working.

But only the DPI change. Anything else and ratbagd errors with HID++ error from the device (255): ERR_INVALID_ARGUMENT (02)

SonicFrog:
This adds support for the non-RGB LEDs feature in HID++2.0 mice. 
Should fix #325 
Feedback and advice on general structure welcome!


bentiss:
Thanks, this would be indeed a great addition.

However, when I try to run your branch locally with the G900 or the G502, I have a segfault at https://github.com/SonicFrog/libratbag/blob/wip/g302-fix/src/driver-hidpp20.c#L317
It looks like `led_info` is NULL, so I *think* there might be an issue in how the various devices are enumerating the features.

SonicFrog:
Still working on that one. The weird thing is that the G502 I have only has non-RGB LEDs and crashes inside the read_lead_1300 function.

SonicFrog:
The SEGFAULT should be fixed now, though there may some less obvious bugs still left.

bentiss:
Few comments on the PR:
- there is no good way to test this currently as ratbag-command and ratbagd/ratbagctl are not updated
- which means I have some unsupported -1 on the G900 that makes the `info` command of ratbagctl crash on the G900
- the G502 always seem to say that the modes are off even though the led is light up
- trying to set up something on the g502 returns an error saying that setting the LED is not supported here (did you add a new capability?)
- when starting ratbagd, the g502 lights off and on again... while for the G900 we just read the state in the profile, so I wonder why you are forced to use software controlled LED...

Regarding the code itself, there are a few formatting issues here and there. We can pinpoint those shortly before merging once it is fully functional.
If the assumption that you exported new capabilities is correct, then this is not something you should do (I think). The LED interface should stay the same whether or not you are using RGB LED. We should however add a new capability/function to query the numbers of bits we can set for the color.

And regarding the split of the patches, `add non-RGB led reading ` and `add non-RGB led updating` should come together in the same patch, and I think the last one should also be squashed with the previous too.
Basically, when you split code to have a smaller commit, you always need to ask 2 questions:
- is my code not actually adding dead code?
- can my code be bisected?

For the first one, you should make sure that any function is actually called and be reachable (from  a rough look, it seem the `led reading` and `led updating` commits are not raising a warning by gcc but nothing sets the capability, so basically there is no way to test this code until the last one)

For the second advice, you should always make sure the code is bisectable. Here, a git bisect will say that the faulty commit will be the last one when this one doesn't do much.

So it is fine to have big commits if there is no other way around it.

whot:
https://github.com/libratbag/libratbag/pull/360 has the start of LED color depth support

SonicFrog:
I guess there are a few points that were not totally clear, when I say non-RGB LEDs this actually means monochrome LEDs. There is no way at all to control the LED color through feature 0x1300. The LED just gives you a bitfield which indicates what effects it supports (breathing, blinking, etc...). 

Also LEDs advertised through 0x1300 do not always support the full range of effects available, currently this is handled by returning a -ENOTSUP when trying to apply an effect to a LED that does not support it. 

I've not added a new ratbag capability, I've only added one to the hidpp20_driver_data structure in order to be able to select the correct method to update the led, either 0x1300 or 0x8070. Though your problem might related to the above.

As for the on and off problem at startup, this is because the code sets the software control property at startup causing the LED control to go from FW to SW and a little flicker. I'll fix this by only turning on software control when writing a LED.

And finally, thanks for the git advices ;)

whot:
> this is handled by returning a -ENOTSUP when trying to apply an effect to a LED that does not support it.

That's stuff that needs to go into the `.device` file then so we don't rely on guesswork by the user.

cvuchener:
The data I get from my G502 spectrum or a [G402](https://github.com/cvuchener/hidpp/issues/1#issuecomment-276006867) does not seem to fit the hidpp20_led_sw_ctrl_led_info structure.

 - G502 spectrum
```
00 02 03 00 03 07 00 00 00 00 00 00 00 00 00 00
01 03 03 00 03 00 00 00 00 00 00 00 00 00 00 00
```
 - G402
```
00 02 03 00 03 07 00 00 00 00 00 00 00 00 00 00
01 04 01 00 80 00 00 00 00 00 00 00 00 00 00 00
```

With this structure:
```
struct hidpp20_led_sw_ctrl_led_info {
	uint8_t index;
	uint16_t type;
	uint8_t physical_count;
	uint16_t caps;
	uint8_t nvconfig_caps;
} __attribute__((packed));
```

That gives types 0x0203, 0x0303 or 0x0401. And zero nvconfig caps, although, on the G502 spectrum, functions 6 and 7 can control if the DPI LEDs stay on (LED 0 only). It would fit better with a 8 bit type.

SonicFrog:
Good catch! I'll fix this right away.

src/hidpp20.h
```diff
@@ -127,6 +127,126 @@ int hidpp20_batterylevel_get_battery_level(struct hidpp20_device *device,
...
+			uint16_t unused;
+			uint16_t delay;
+		} traveling;
+		struct {
+			uint16_t index;
+			uint16_t on_time;
+			uint16_t off_time;
+		} blink;
+		struct {
+			// Which physical LED to turn on
```
cvuchener: The G502 spectrum DPI LED has 3 physical LEDs, but the on.index accepted values are in the 1–5 range (for the 5 DPI index). The profile LED also has 3 physical LEDs (actually the same 3, they are shared), and the accepted value range is 1–3 (for 3 profiles). I don't think this index controls the physical LEDs directly.

SonicFrog: It may not be direct but setting this determines which physical LED(s) is/are lit up. Though I agree the formulation is not the clearest ;)

SonicFrog:
Finally managed to fix the issue with the G900. It appears it has both RGB and non-RGB LEDs. For now, I've made the two features mutually exclusive with a priority on RGB LEDs. Though, it seems to me like this would be worth a more in-depth discussion about how to handle the mice (mouse?) that have both.

cvuchener:
I did not say it explicitly but the G502 spectrum is a RGB mouse too (that's the difference with the G502 core). I actually expect most mouse with 0x8070 to have 0x1300 too. But it is not used for LED effects, so you can ignore it. The useful setting is the "DPI LEDs stay on" one, but I don't know how that would fit in libratbag.

whot:
>  The useful setting is the "DPI LEDs stay on" one, but I don't know how that would fit in libratbag.

we ignore it. libratbag is a generic interface and every device has some wonky settings here and there that may be exposed by the hardware (but no other mouse). it's usually not worth the effort of making this accessible so we just ignore it. I think that's the best case here too.

SonicFrog:
After some extensive tests, it works fine with:

- G300s
- G302
- G502
- G603
- G703
- G900
- G903

If anyone has one I haven't tested that would be great!

phomes:
> After some extensive tests, it works fine with:
> G300s

Does the G300s use the hid++2.0 driver? I thought that it used the same as the G300?

bentiss:
On Sep 26, 2017 15:11, "Ogier Bouvier" <notifications@github.com> wrote:

After some extensive tests, it works fine with:

   - G300s
   - G302
   - G502
   - G603
   - G703
   - G900
   - G903

If anyone has one I haven't tested that would be great!


Nice. Sorry if I do not me this right now. I am at a conference until the
rest of the week and can't really work on libratbag. If anyone else wants
to do the review, feel free to pick it up.

SonicFrog:
> Does the G300s use the hid++2.0 driver? I thought that it used the same as the G300?

It doesn't, but I checked every mice I had on hand.

src/hidpp20.h
```diff
@@ -269,7 +390,7 @@ hidpp20_color_led_effects_get_zone_info(struct hidpp20_device *device,
 
 int
 hidpp20_color_led_effects_get_zone_infos(struct hidpp20_device *device,
-					 struct hidpp20_color_led_zone_info **infos_list);
+					struct hidpp20_color_led_zone_info **infos_list);
```
whot: might be github but the alignment seems off here

src/hidpp20.h
```diff
@@ -127,6 +127,128 @@ int hidpp20_batterylevel_get_battery_level(struct hidpp20_device *device,
...
+			uint16_t index;
+		} on;
+	};
+} __attribute__((packed));
+_Static_assert(sizeof(struct hidpp20_led_sw_ctrl_led_state) == 9, "Invalid size");
+
+int hidpp20_led_sw_control_read_leds(struct hidpp20_device* device,
+		struct hidpp20_led_sw_ctrl_led_info** info_list);
+
+_Static_assert(sizeof(struct hidpp20_led_sw_ctrl_led_state) == 9, "Invalid size");
```
whot: duplicate, please remove

src/hidpp20.h
```diff
@@ -127,6 +127,128 @@ int hidpp20_batterylevel_get_battery_level(struct hidpp20_device *device,
...
+
+/**
+ * Retrieves the number of non-RGB logical leds from the mouse
+ * @return the led count or a negative errno
+ */
+int hidpp20_led_sw_control_get_led_count(struct hidpp20_device* device);
+
+/**
+ * Retrieves the info for a given logical led
+ * return 0 or a negative errno in case of error
```
whot: `@return`

src/hidpp20.h
```diff
@@ -127,6 +127,128 @@ int hidpp20_batterylevel_get_battery_level(struct hidpp20_device *device,
...
+ */
+int hidpp20_led_sw_control_get_led_info(struct hidpp20_device* device,
+		uint8_t led_idx,
+		struct hidpp20_led_sw_ctrl_led_info* info);
+
+int hidpp20_led_sw_control_read_leds(struct hidpp20_device* device,
+	 struct hidpp20_led_sw_ctrl_led_info** info_list);
```
whot: doc missing

src/hidpp20.h
```diff
@@ -127,6 +127,128 @@ int hidpp20_batterylevel_get_battery_level(struct hidpp20_device *device,
...
+
+/**
+ * Get the current sw control value
+ * @return 0 if the leds are sofware-controlled 1 if they are firmware controlled
```
whot: if it's 0/1 only, can we make it a bool please?

src/hidpp20.h
```diff
@@ -127,6 +127,128 @@ int hidpp20_batterylevel_get_battery_level(struct hidpp20_device *device,
...
+ */
+int hidpp20_led_sw_control_get_sw_ctrl(struct hidpp20_device* device);
+
+/**
+ * Sets who controls the LEDs
+ * ctrl = 0, the software controls the LEDs
+ * ctrl = 1, the firmware controls the LEDs
+ * @return 0, or a negative errno
+ */
+int hidpp20_led_sw_control_set_sw_ctrl(struct hidpp20_device* device, uint8_t ctrl);
```
whot: same here, bool please

src/hidpp20.h
```diff
@@ -127,6 +127,128 @@ int hidpp20_batterylevel_get_battery_level(struct hidpp20_device *device,
...
+
+/**
+ * Gets the state of a LED
+ * @return 0 or a negative errno
```
whot: "0 on success or a negative errno"

same for the next function

src/driver-hidpp20.c
```diff
@@ -247,17 +247,57 @@ hidpp20drv_read_button(struct ratbag_button *button)
...
 
+	led->colordepth = RATBAG_LED_COLORDEPTH_MONOCHROME;
+
+	switch (info->type)
+	{
+	case HIDPP20_LED_TYPE_LOGO:
+		led->type = RATBAG_LED_TYPE_LOGO;
+		break;
+	case HIDPP20_LED_TYPE_BATTERY:
+	case HIDPP20_LED_TYPE_DPI:
```
whot: both of those should have their own `RATBAG_LED_TYPE` enum value

src/driver-hidpp20.c
```diff
@@ -247,17 +247,57 @@ hidpp20drv_read_button(struct ratbag_button *button)
...
+		break;
+	case HIDPP20_LED_MODE_ON:
+		led->mode = RATBAG_LED_ON;
+		break;
+	case HIDPP20_LED_MODE_BREATHING:
+		led->mode = RATBAG_LED_BREATHING;
+		led->hz = state.breathing.period;
+		led->brightness = state.breathing.max_brightness;
+		break;
+	}
```
whot: what happens if any other mode is set? we should default this to 'ON' imo, even when it's obviously incorrect.

src/driver-hidpp20.c
```diff
@@ -285,6 +325,21 @@ hidpp20drv_read_led(struct ratbag_led *led)
...
+	struct hidpp20drv_data *drv_data = ratbag_get_drv_data(device);
+
+	if (drv_data->capabilities & HIDPP_CAP_COLOR_LED_EFFECTS_8070) {
+		hidpp20drv_read_led_8070(led, drv_data);
+		return;
+	}
+
+	if (drv_data->capabilities & HIDPP_CAP_LED_SW_CONTROL_1300)
+		hidpp20drv_read_led_1300(led, drv_data);
+}
```
whot: can we make this an `if/else if`? It feels a big asymmetrical

src/driver-hidpp20.c
```diff
@@ -389,16 +444,60 @@ hidpp20drv_update_button(struct ratbag_button *button,
...
+
+	if (!(h_led.mode & led_caps)) {
+		hidpp_log_error(&data->dev->base, "LED %d does not support effect %s(%04x), supports %04x\n",
+						led->index, hidpp20_sw_led_control_get_mode_string(h_led.mode),
+						h_led.mode, led_caps);
+		return -ENOTSUP;
+	}
+
+	rc = hidpp20_led_sw_control_set_sw_ctrl(data->dev, true);
+
```
whot: skip this empty line here, same below

src/driver-hidpp20.c
```diff
@@ -692,6 +807,28 @@ hidpp20drv_read_color_leds(struct ratbag_device *device)
...
+	int rc;
+
+	if (!(drv_data->capabilities & HIDPP_CAP_LED_SW_CONTROL_1300))
+		return 0;
+
+	free(drv_data->led_infos.leds);
+	drv_data->led_infos.leds = NULL;
+
+	rc = hidpp20_led_sw_control_read_leds(drv_data->dev, &drv_data->led_infos.leds);
+
```
whot: no empty line

SonicFrog:
Should be good now!

PR #359, opened 2017-09-22
Add a ratbag_dispatch() method to allow hooking into a mainloop

whot:
This is something we'll eventually need when we start monitoring events. We don't have a user for this just yet, so we shouldn't merge it.

I'm sending this as PR now because the code is written and I think that's the approach we want. So once we need it, we can rebase this and push it out.

### DIFF
--- a/ratbagd/ratbagd.c
+++ b/ratbagd/ratbagd.c
@@ -489,9 +489,20 @@ static int sighandler(sd_event_source *source,
 	return 0;
 }
 
+static int ratbagd_dispatch(sd_event_source *s, int fd,
+			    uint32_t revents, void *data)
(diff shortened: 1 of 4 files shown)


Hjdskes:
Probably missing something obvious here but what would such a user be in this case? We currently have the DBus API that also allows to monitor for changes; and since we don't export libratbag as a library anymore I fail to see how this is relevant.

whot:
any event processing inside libratbag will need it. This is internal API, not external so it's ratbagd that keeps calling into libratbag to process stuff - invisible to the client. Potential users are any driver API that uses e.g. dbus to communicate with some other API (like openrazer)

PR #360, opened 2017-09-22
led: add support for color depth

whot:
The initial value of 400 in the enum is chosen so it looks weird when someone
tries to interpret the value as integer. Values of 1, 2, 4, etc. could be
confused with something realistic.


bentiss:
Actually, testing this on the G502, with monochrome, I think we should default in libratbag to RGB when we initialize a LED. This is the default we currently assume, and once all the drivers will all overwrite the value with the correct depth, we can remove this fallback.

bentiss:
Forget about it, the code is doing the right thing, I was running ratbagctl without the .devel.... Should probably completely wake up before commenting :)

whot:
Isn't that what the  first `src/libratbag.c` hunk does? I have a feeling there's a 'not' missing somewhere in your comment :)

PR #366, opened 2017-09-28
G9 fix commit

phomes:
This fixes writing profiles on the G9. The problem was when setting the current resolution, which is a short report on the G9 but a long on other models. While looking into it I noticed that we are getting/setting the current resolution unnecessarily many times, and that we set it before the profile write which causes the device to jump to its default resolution.

### DIFF
--- a/src/hidpp10.c
+++ b/src/hidpp10.c
@@ -2109,12 +2109,24 @@ hidpp10_set_current_resolution(struct hidpp10_device *dev,
 			       uint16_t yres)
 {
 	unsigned idx = dev->index;
-	union hidpp10_message resolution = CMD_CURRENT_RESOLUTION(REPORT_ID_LONG, idx, SET_LONG_REGISTER_REQ);
+	union hidpp10_message resolution = CMD_CURRENT_RESOLUTION(REPORT_ID_SHORT, idx, SET_REGISTER_REQ);
+	union hidpp10_message resolution_long = CMD_CURRENT_RESOLUTION(REPORT_ID_LONG, idx, SET_LONG_REGISTER_REQ);
+	int res;
 
-	hidpp_set_unaligned_le_u16(&resolution.data[4], hidpp10_get_dpi_mapping(dev, xres));
-	hidpp_set_unaligned_le_u16(&resolution.data[6], hidpp10_get_dpi_mapping(dev, yres));
+	switch (dev->profile_type) {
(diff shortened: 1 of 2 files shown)


src/hidpp10.c
```diff
@@ -2109,12 +2109,23 @@ hidpp10_set_current_resolution(struct hidpp10_device *dev,
...
 {
 	unsigned idx = dev->index;
-	union hidpp10_message resolution = CMD_CURRENT_RESOLUTION(REPORT_ID_LONG, idx, SET_LONG_REGISTER_REQ);
+	union hidpp10_message resolution = CMD_CURRENT_RESOLUTION(REPORT_ID_SHORT, idx, SET_REGISTER_REQ);
+	union hidpp10_message resolution_long = CMD_CURRENT_RESOLUTION(REPORT_ID_LONG, idx, SET_LONG_REGISTER_REQ);
+	int res;
+
+	switch (dev->profile_type) {
+	case HIDPP10_PROFILE_G9:
+		hidpp_set_unaligned_le_u16(&resolution.data[4], hidpp10_get_dpi_mapping(dev, xres));
```
cvuchener: Shouldn't it be a 8 bit value?

phomes: Good catch. It should. hidpp10_get_dpi_mapping returns a uint8_t so we can just write that directly. The result is the same but it is more correct. I'll force push an updated patch

src/hidpp10.c
```diff
@@ -2109,12 +2109,23 @@ hidpp10_set_current_resolution(struct hidpp10_device *dev,
...
+		res = hidpp10_request_command(dev, &resolution);
+		break;
+        default:
+		hidpp_set_unaligned_le_u16(&resolution_long.data[4], hidpp10_get_dpi_mapping(dev, xres));
+		hidpp_set_unaligned_le_u16(&resolution_long.data[6], hidpp10_get_dpi_mapping(dev, yres));
 
-	hidpp_set_unaligned_le_u16(&resolution.data[4], hidpp10_get_dpi_mapping(dev, xres));
-	hidpp_set_unaligned_le_u16(&resolution.data[6], hidpp10_get_dpi_mapping(dev, yres));
+		res = hidpp10_request_command(dev, &resolution_long);
+        }
```
whot: please always add a break, even when it's the default case

PR #368, opened 2017-09-29
Remove the 'unknown' LED type

whot:
We have the `.device` files now, so any information the driver can't get from the device we can pull from the device files if we need to. So instead of the 'unknown' LED type assert in libratbag to make sure any new device support is eager to fix this immediately.

Note: currently based on #367, only the last 4 commits count here.
CC @lrenaud


Hjdskes:
Not sure why you want my review on this; most of this seems to be related to libratbag internals where I can't say if this is a good solution. The ratbagd changes look good to me, and can be used as-is with Piper if I am not mistaken.

whot:
rebased on master to fix the conflicts

whot:
@Hjdskes don't undersell yourself, you've worked enough with libratbag to have a say in its design :)

@bentiss - can I get a review for this please?

bentiss:
FYI, I pushed 670720b40fa4 before and c3be9e1af13b after this series.

bentiss:
And also 5c71f95 to make flake8 happy

PR #374, opened 2017-09-30
Hidpp10 leds from device

phomes:
The G9 at least does not seem to provide information about the LED count. This adds a new entry 'LEDs' to the [Driver/hidpp10] section in the .device file. If a value is set there it will override anything that may already be set by the driver.


src/libratbag-data.c
```diff
@@ -104,6 +106,13 @@ init_data_hidpp10(struct ratbag *ratbag,
 	if (error)
 		g_error_free(error);
 
+	error = NULL;
+	num = g_key_file_get_integer(keyfile, group, "LEDs", &error);
```
whot: Change it to `Leds` please, we're using `Dpi` instead of `DPI` in those files too. And add the entry to the `device.example` file so it's documented.

phomes:
updated and pushed. Thanks

PR #375, opened 2017-10-04
test: set RATBAG_TEST in code, not in meson

whot:
It's a fixed value so it's much easier to set this in the test binaries rather
than requiring the user to set this in the env before running a test manually.


whot:
I don't think so, it's now always set in the source of the program we want to run. and using the env is the only way to pass this down to libratbag (though admittedly now that it's not a library anymore we could just add an API for that).

PR #376, opened 2017-10-04
Logitech G703 wired support

SonicFrog:
Adds a device file for the Logitech G703 connected through wired USB, as well as an SVG for the gnome theme.


data/gnome/logitech-g703.svg
```diff
@@ -0,0 +1 @@
+logitech-g403.svg
```
bentiss: Looks like this is not a SVG :)

Hjdskes: This is a symlink to logitech-g403.svg.

bentiss: Right, I have been mistaken by github's web ui.

In that case, wouldn't it be better to just have `logitech-g403.svg` in the `.device` file and add a comment saying both are the same. Having a separate file is just misleading IMO.

Hjdskes: There is at least one other symlink that follows the current approach, as far as I know.

Issue #380, opened 2017-10-06
Logitech wireless mice are not identified properly


SonicFrog:
All wireless Logitech gaming mice use the same receiver (046d:c539). This receiver is a variant of the unifying receiver except it supports pairing with only one device. 

Currently libratbag names all mice using this receiver as Logitech USB Receiver and uses the G900 svg. In short this means that the G403, G603, G703, G903 use the wrong SVG and are misnamed

![2017-10-06-124057_664x244_scrot](https://user-images.githubusercontent.com/3202531/31274535-9e051bec-aa93-11e7-8619-9d676b0f58fe.png)
 
This is fixable by using the receiver during device discovery to ask the paired device to give its name and product id.

qidydl:
Plugging in the mouse directly in order to use Piper is a fairly easy temporary workaround, maybe that could be noted in documentation or a quick message could be displayed in the UI?

PR #381, opened 2017-10-06
Fix Logitech wireless identification

SonicFrog:
This is a working fix for #380, though I wanted some feedback on design first. 

- First, I've had to merge liblur into libhidpp in order to talk to the USB receiver during the hidpp20 driver probe phase
- Since the SVG is not known until speaking with the receiver, the ratbag_device_data structure needs to be updated during the probe phase or we need to change ratbag's device identification procedure. 
- This fix avoids making big changes to the general device discovery since this is Logitech-specific at the cost of making the ratbag_device_data structure mutable from outside `libratbag-data.c`. What do you think is the best way?
- Small change to liblur to consider the Logitech gaming mouse USB receiver (046d:c539) as a unifying receiver

### DIFF
--- a/data/devices/data-parse-test.py
+++ b/data/devices/data-parse-test.py
@@ -164,7 +164,7 @@ def check_section_hidpp10(section):
 
 
 def check_section_hidpp20(section):
-    permitted = ['DeviceIndex']
+    permitted = ['DeviceIndex', 'IsWireless']
     for key in section.keys():
         assert(key in permitted)
 
--- a/data/devices/logitech-wireless-receiver.device
+++ b/data/devices/logitech-wireless-receiver.device
@@ -1,0 +1,11 @@
+# Logitech G403, G603, G703, G900 and G903 over wireless USB
+[Device]
+Name=Logitech G403, G603, G703, G900 and G903
+DeviceMatch=usb:046d:c539
+Driver=hidpp20
+# This is not used since the real svg name is discovered later
+Svg=logitech-g900.svg 
+
+[Driver/hidpp20]
+DeviceIndex=1
+IsWireless=true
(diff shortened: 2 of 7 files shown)


bentiss:
TBH, this is something I wanted to do in the kernel, not in libratbag. And this was the series I mentioned during our call early September (that I have locally in my tree since March).
I'd rather have this in the kernel because it'll make other tools behave properly with proper logs. I'll come back later with a build on Fedora if you are using Fedora, or I can point you to a branch and you can recompile the kernel.

SonicFrog:
Oh, guess I forgot we talked about it... I'll just close this, and yes I'm using Fedora though I can just do the build myself if you give me the link :)

whot:
@bentiss it's a year later and afaict this hasn't been pushed into the kernel anywhere? Did this get lost?

bentiss:
> @bentiss it's a year later and afaict this hasn't been pushed into the kernel anywhere? Did this get lost?

Not really. Each of my attempts to push it and ask for feedback resulted in a /dev/null answer. So I still need to push it.

But I agree, in the meantime, it would be a nice feature to have.

jonathanplatzer:
Is there any foreseeable progress regarding the merge into the kernel?
Or would I be better of building libratbag on my own with this PR included?

bentiss:
This should be naturally handled in kernel v5.2+. At least the `0xc539` usb receiver will.

FFY00:
Then we should make a new release.

bentiss:
> Then we should make a new release.

well, there is still a long road to go before this gets in a common distribution. I have scheduled the series for 5.2, but we are at 5.1-rc6 now. So there is at least 2+2+8=12 weeks before 5.2 is out, and then I guess distributions will take at least a couple of weeks before updating.

Also, I'd like to get the powerplay mouse mat handled by the kernel first. Just saying in case somebody has one around :)

PR #382, opened 2017-10-06
hidpp10: remove empty callback

phomes:
write_profile is deprecated and none of the other drivers have an empty
implementation like this.


whot:
the ubuntu CI failure is the startup race condition again, can be ignored here

PR #387, opened 2017-10-09
Switch to a flat probe() function instead of the callbacks

whot:
Note: this builds on #368


whot:
Rebased and force-pushed, thanks

phomes:
I have tested with the logitech G300. LGTM

phomes:
and tested with G9 (HID++1.0). It works fine there as well with a patch to fix up an unrelated issue.

bentiss:
OK, then I consider this to be enough. If G-Skill breaks, we'll hear it one way or one other.

PR #389, opened 2017-10-11
Add ratbagd_led_has_mode() and the required dbus additions

whot:
To be able to query which modes are available.

Exposed via ratbagctl as
./build/ratbagctl.devel led 0 capabilities event4

### DIFF
--- a/src/driver-hidpp20.c
+++ b/src/driver-hidpp20.c
@@ -295,6 +295,10 @@ hidpp20drv_read_led_1300(struct ratbag_led *led, struct hidpp20drv_data* data)
 	default:
 		led->mode = RATBAG_LED_ON;
 	}
+
+	ratbag_led_set_mode_capability(led, RATBAG_LED_ON);
+	ratbag_led_set_mode_capability(led, RATBAG_LED_BREATHING);
+	ratbag_led_set_mode_capability(led, RATBAG_LED_OFF);
 }
 
 static void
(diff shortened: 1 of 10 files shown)


src/driver-hidpp20.c
```diff
@@ -344,6 +344,14 @@ hidpp20drv_read_led(struct ratbag_led *led)
 		hidpp20drv_read_led_8070(led, drv_data);
 	else if (drv_data->capabilities & HIDPP_CAP_LED_SW_CONTROL_1300)
 		hidpp20drv_read_led_1300(led, drv_data);
+	else
+		return;
+
+	ratbag_led_set_mode_capability(led, RATBAG_LED_ON);
+	ratbag_led_set_mode_capability(led, RATBAG_LED_CYCLE);
```
SonicFrog: Monochrome LEDs don't support the cycle mode since there's no color to cycle between. Moreover not all LEDs even support the basic on mode (some can only be put in breathing mode). For the monochrome I would rather fetch it from the mouse in hidpp20drv_read_led_1300.

whot: I moved this to the respected read_lead_XXXX and dropped the cycle bit. For anything more specific it's prob. better to do a follow-up patch because it seems we need some .device file entries for this, don't we?

whot:
Updated, with a small test case

bentiss:
Sorry, but this one needs to be rebased after the sphinx/doc merge.

whot:
rebased, merging this one now to avoid another cycle

src/driver-hidpp20.c
SonicFrog: Actually no, at least not for the monochrome LEDs (and as far as I know all RGB LEDs support the 4 effects), since they advertise their supported effects through the 0x1300 feature. The caps attribute of struct hidpp20_led_sw_ctrl_led_info is a bitfield indicating which effects are supported.

PR #392, opened 2017-10-12
Use Sphinx to generate the DBus API documentation

whot:
This is just the basic conversion - the Read-the-docs theme allows for a lot more documentation to be added at some later point.


whot:
> That seems quite complex :/ I am confident there are no other ways, but still...

There's the breathe extension I just found: https://breathe.readthedocs.io/en/latest/
That may move the doxygen+xml bit to sphinx directly, but in the end we'll do the same thing anyway.

>  isn't doc/dbus.rst autogenerated? If so, is it relevant to add it to the commit?

No, it's not. That's the handwritten one. The big problem with the dbus API is that it doesn't exist anywhere but on dbus. It's not like a class or something, it really needs to be hand-written.

> I think I'd like to have the circle.yml updated to generate the documentation also in the CI, that would help knowing the dependencies and make sure the whole thing compiles

I can do that easily enough, but fwiw I think the documentation generation + upload will be a manual task, once the DBus API changes we'll likely won't have too many changes here. Anyway, let's see how this goes

whot:
> There's the breathe extension [...]

Ok, this one isn't useful for us. It doesn't let us link to enum values which is the main thing we need.  https://github.com/whot/libratbag/tree/wip/sphinx-breathe are my efforts, but I didn't get anywhere.

circleci integration is added now, and I'v pushed the output to `gh-pages`, so it's already showing up on http://libratbag.github.io/

PR #393, opened 2017-10-13
DBus API change to switch to discrete resolution lists

whot:
This builds on top of #392 

Changes from a dpi min/max to a discrete list of "here are the resolutions we support". This makes it easier for clients to do the right thing.


bentiss:
Not sure why, but you need to rebase this one too.

phomes:
This broke piper. Is an update for it being worked on?

Hjdskes:
I guess the breakage was.. intentional, considering @bentiss' comment :) I just opened https://github.com/libratbag/piper/pull/186.

PR #394, opened 2017-10-13
circle.yml: deploy docs when the master branch is updated

bentiss:
Note that the push requires a specific private key in the libratbag circle CI organization, meaning that only the master branch of libratbag will be allowed to push to the doc repo.

Note also that only the master branch will trigger the request:
- any PR will not trigger the request
- any push on the master branch of a personal repo will fail while attempting to write the docs, if the repo is setup to use circleCI too.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -125,6 +125,38 @@ doc_build: &doc_build
     - *export_logs
     - store_artifacts:
         path: ~/libratbag/build/doc/html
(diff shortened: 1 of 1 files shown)


bentiss:
Actually, I don't think anyone cares but me and maybe @whot. Let's just merge this and see how it behaves. Worse case, we revert this.

PR #395, opened 2017-10-13
circleCI: cache DNF data accross jobs

bentiss:
We don't really need to redownload everything for all the various fedora jobs.
Launch an extra Fedora job that retrieves and store the DNF cache, and then make use of it in the rest of the workflow.

We do not gain much right now in term of time spent. We should get a rough 1 min when deploying the docs though on the master branch.

This goes on top of PR #394.


bentiss:
And taking this one too, because I can

PR #397, opened 2017-10-13
Update ratbagctl man

bentiss:
Should be closer to the truth now.

### DIFF
--- a/tools/ratbagctl.1
+++ b/tools/ratbagctl.1
@@ -1,72 +1,201 @@
-.TH ratbagctl 1 "Apr 13, 2016" ratbagctl
+.TH RATBAGCTL "1" "@version@"
 .SH NAME
 ratbagctl \- inspect and modify configurable mice
 .SH SYNOPSIS
-.B ratbagctl
-.RB [ \-h ]
-.RB [ \-V ]
-.br
-.B ratbagctl list-devices
-.br
-.B ratbagctl show-device
-.RI < device >
-.br
-.B ratbagctl show-profile
-.RI < device "> <" profile >
-.br
-.B ratbagctl show-resolution
-.RI < device "> <" profile "> <" resolution >
-.br
-.B ratbagctl show-button
-.RI < device "> <" profile "> <" button >
+.B ratbagctl [OPTIONS] {COMMAND} ... device
 .SH DESCRIPTION
-.SS Common options
-.TP
-.BR \-h
-displays a short help message (it can be used with any form of the
+.PP
+The
+.I ratbagctl
+tool queries or changes a device's settings.
+.PP
+This tool usually needs ratbagd to be running.
+.SH OPTIONS
+.TP 8
+.B \-\-version, \-V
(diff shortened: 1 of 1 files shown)


tools/ratbagctl.1
```diff
@@ -1,72 +1,202 @@
...
+.SH OPTIONS
+.TP 8
+.B \-\-version, \-V
+Show program's version number and exit.
+.TP 8
+.B \-\-verbose, \-v
+Print debugging output.
+.TP 8
+.B \-vvv
+Print more debugging output (with protocol output).
```
whot: this should just say something like 'Multiple -v options increase the verbosity`, because -vv is valid too

tools/ratbagctl.1
```diff
@@ -1,72 +1,202 @@
...
+Set the name of the profile
+.TP 8
+.B enable
+Enable a profile
+.TP 8
+.B disable
+Disable a profile
+.TP 8
+.B [resolution|dpi|rate|button|led] ...
+Use profile N for the other commands.
```
whot: 'for the specified command'

tools/ratbagctl.1
```diff
@@ -1,72 +1,202 @@
...
+Use resolution N for COMMAND
+.PP
+.B Available COMMANDs:
+.RS
+.TP 8
+.B get
+Print selected resolution
+.TP 8
+.B [dpi|rate] ...
+Use resolution N for the other commands.
```
whot: same as above

tools/ratbagctl.1
```diff
@@ -1,72 +1,202 @@
...
+.B button N action set special S
+Set the button action to special action S
+.TP 8
+.B button N action set macro ...
+Set the button action to the given macro
+.PP
+.B Macro syntax:
+.HP 8
+A macro is a series of key events or waiting periods. Keys must be specified
+in linux/input.h key names.
```
whot: we should probably use input-event-codes.h now, it's been in place for a while

tools/ratbagctl.1
```diff
@@ -1,72 +1,202 @@
...
+Use led N for COMMAND
+.PP
+.B Available COMMANDs:
+.RS
+.TP 8
+.B mode [on|off|cycle|breathing]
+The mode to set as current
+.TP 8
+.B color RRGGBB
+The color to set as current
```
whot: ... something about hex

tools/ratbagctl.1
```diff
@@ -1,72 +1,202 @@
...
+ratbagctl profile 0 resolution active set 4 eventX
+.TP 8
+ratbagctl profile 0 resolution 1 dpi get eventX
+.TP 8
+ratbagctl resolution 4 rate get eventX
+.TP 8
+ratbagctl dpi set 800 eventX
+.SH NOTES
+.PP
+There is currently no guarantees that the output format of
```
whot: 'is no guarantee' or 'are no guarantees', one of the two :)

PR #398, opened 2017-10-15
Initial driver and svg for steelseries rival 310

phomes:
This is the initial step towards a driver for the steelseries mice. In the first version this is just for the "rival 310" but the driver should be usable for the other models with minor adjustments in the driver.

The driver does not yet support button mappings or any of the non-solid LED modes. Resolutions, report rate, and solid colors on two LEDs are working.

I open this PR now for two reasons. I would like to include it in the current form. I have found a couple of other developers I would like to attract to the project. And being able to set some but not all available settings is better than nothing I guess :) 

The other reason is to discuss some issues with the device. It does not support reading out the current onboard configuration. This is of course bad for the experience in piper. For now the driver just picks some default values during probe that I found sensible What do we do for such a device? One solution could be to store the configuration on disk and load that during probe. Not ideal, but better than a full reset every time. Or should we simply not support devices with such limitations?

One of the LEDs is placed within the scroll wheel. I think we need a new LED TYPE for that. I have seen some of the logitech mice do the same. Is there a reason such a type does not exist yet?


Hjdskes:
> One of the LEDs is placed within the scroll wheel. I think we need a new LED TYPE for that. I have seen some of the logitech mice do the same. Is there a reason such a type does not exist yet?

My G403 is one such Logitech device, just FYI.

As for the SVG; the colors seem to be a bit off when I compare them to e.g. the G403 SVG.

phomes:
Thanks. I am colorblind so any more concrete feedback on how to adjust the colors is appreciated :) I just copied the colors otherwise used in some of the existing SVGs. It may be a bit dark this time, but the mouse is black and only the part the right and a small chunk on the left are in a lighter color.

src/driver-steelseries.c
```diff
@@ -0,0 +1,271 @@
...
+#include "hidpp-generic.h"
+
+#define STEELSERIES_NUM_PROFILES	1
+#define STEELSERIES_NUM_BUTTONS		6
+#define STEELSERIES_NUM_DPI		2
+#define STEELSERIES_NUM_LED		2
+#define STEELSERIES_DPI_MIN		100
+#define STEELSERIES_DPI_MAX		12000
+
+// not sure these are used for
```
whot: pls use /* comments */, that goes for the rest of the driver too

whot:
@phones: the dark colour is the one we use for the LEDs on other mice. I think if you swap the body for the same colour as the buttons and change the LEDs to have the dark colour, that should be good enough.

> it does not support reading out the current onboard configuration

see the `RATBAG_DEVICE_CAP_QUERY_CONFIGURATION` capability. we're basically pushing this to piper.

> One of the LEDs is placed within the scroll wheel. I think we need a new LED TYPE for that. I have seen some of the logitech mice do the same. Is there a reason such a type does not exist yet?

No specific reason, please add it.

phomes:
Updated for comments. Also added a patch for scroll wheel LED and one to allow unsetting capabilites.

src/libratbag-enums.h
```diff
@@ -342,6 +342,7 @@ enum ratbag_led_type {
 	RATBAG_LED_TYPE_SIDE,
 	RATBAG_LED_TYPE_BATTERY,
 	RATBAG_LED_TYPE_DPI,
+	RATBAG_LED_TYPE_SCROLL_WHEEL,
```
whot: looking at this, I think we should just call it 'wheel'. The physical types are an aid but don't need to be 100% accurate (see also side vs left/right side) and 'wheel' is more generic than scroll wheel - and less to type. With that, ack by me for this one.


also, please make sure this commit is before the other one that uses the new type

meson.build
```diff
@@ -331,6 +333,7 @@ gnome_svg_files = files(
 	'data/gnome/logitech-g9.svg',
 	'data/gnome/logitech-g900.svg',
 	'data/gnome/roccat-kone-xtd.svg',
+	'data/gnome/steelseries-rival310.svg',
```
whot: the svg still looks a lot darker than all the other ones. not a killer but it makes the contrast to the LED symbol worse. did you use the same color that the other mice use for the main body? maybe it's github's SVG rendering that's different here.

src/libratbag-data.c
```diff
@@ -46,7 +46,8 @@ enum driver {
 	ROCCAT,
 	ETEKCITY,
 	GSKILL,
-	LOGITECH_G300
+	LOGITECH_G300,
+	STEELSERIES
```
whot: add a comma here, so we don't need to update both lines next time

phomes:
Updated for comments. The SVG is adjusted a bit more to use the same colors as the G403. There was also an issue with the opacity on some elements.

src/libratbag-data.c
```diff
@@ -254,6 +254,8 @@ parse_ledtypes(char **strv, enum ratbag_led_type *types, size_t ntypes)
 			types[i] = RATBAG_LED_TYPE_SIDE;
 		else if (streq(s, "battery"))
 			types[i] = RATBAG_LED_TYPE_SIDE;
+		else if (streq(s, "scrollwheel"))
+			types[i] = RATBAG_LED_TYPE_WHEEL;
```
whot: sorry, one nore nit: drop the 'scroll' please, 'wheel' is enough (everywhere)

phomes:
dropped "scroll" in "scroll wheel" even for human readable strings. That was the last one. Pushing.

PR #402, opened 2017-10-17
hidpp20 keybindings support (again)

bentiss:
Next step before supporting full macro write, allow to detect keybindings and translate them as macro.

Currently only onboard profiles in HID++2.0 support writing back them, so we might want to convert everyone before actually merging this (given that we will change the state of the profiles for the other drivers).

Anyway, reading works OK for HID++ 1.0 and 2.0, it would be cool if the other maintainers could give a try at their drivers so we merge it safely.

This goes on top of #401, so only the last 2 patches matter.

### DIFF
--- a/src/driver-etekcity.c
+++ b/src/driver-etekcity.c
@@ -366,10 +366,20 @@ etekcity_read_button(struct ratbag_button *button)
 	unsigned j;
 	int rc;
 
+	device = button->profile->device;
 	action = etekcity_button_to_action(button->profile, button->index);
 	if (action)
 		ratbag_button_set_action(button, action);
 	button->type = etekcity_raw_to_button_type(button->index);
+
+	if (action->type == RATBAG_BUTTON_ACTION_TYPE_KEY) {
+		if (ratbag_button_macro_new_from_key(button)) {
+			log_error(device->ratbag,
+				  "Error while reading button %d\n",
+				  button->index);
+			button->action.type = RATBAG_BUTTON_ACTION_TYPE_NONE;
+		}
+	}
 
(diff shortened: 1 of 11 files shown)


phomes:
I set two key combinations in LGS on my G300. The device only supports keys, not macros. One button I set to "A" and the other "ctrl+T". In ratbagctl they show as:
  Button: 5 type pinkie is mapped to macro '↕RESERVED'
  Button: 6 type pinkie2 is mapped to macro '↕RESERVED'

In Piper they show as ↕A and ↕T.

phomes:
On my G9 (HID++1.0) I set a macro to ABC.
Ratbagctl correctly sees it as:
  Button: 4 type thumb2 is mapped to macro '↓A 1ms ↑A 1ms ↓B 1ms ↑B 1ms ↓C 1ms ↑C'

In Piper I can see the start of the macro on the button. Clicking it to configure goes directly to the list of button options, so I cannot confirm the last part of the macro. But the start looked correct.

phomes:
I started to play with this for the steelseries driver today. It works well to convert to/from the macros. The thing that got me back to this PR was that I was looking into why I got "UNKNOWN" as the button action for some buttons. Turns out that ratbagd converts the action if it is of type KEY:
```
	if (type == RATBAG_BUTTON_ACTION_TYPE_KEY)
		type = RATBAG_BUTTON_ACTION_TYPE_UNKNOWN;
```
So _KEY is already non-functional and I wanted to start on getting the (single key press) macro working instead of fixing the old way. With this patch I can finally assign keys in Piper. (No modifiers yet, but could be added).

The issue I saw with "RESERVED" in ratbagctl seems to have gone away.

bentiss:
Obsoleted by #591

PR #404, opened 2017-10-18
hidpp20 led capabilities

whot:
Grab the capabilities from the device where possible instead of just assuming we have them.

I'm not 100% sure I read the monochrome-only effects bit right, @SonicFrog please confirm that bit, thanks.


whot:
It's not in the `info` flag but you can run `ratbactl led 1 capabilities`. Maybe not the best interface :)

PR #411, opened 2017-10-20
use SVGs from the source dir in .devel

phomes:
When using ratbagctl.devel we would use the .device files
from the source dir. This patch also makes it use the .svg
files from the source dir. This makes it easier to edit the
svg and less surprising now that it follows the .devices files.

We need to treat the .device and .svg file a bit different. The
reason is that in soure the device files are in a separate dir
'data/devices' but we do not use the 'devices' dir when installed.

### DIFF
--- a/src/libratbag-data.c
+++ b/src/libratbag-data.c
@@ -446,15 +446,18 @@ ratbag_device_data_new_for_id(struct ratbag *ratbag, const struct input_id *id)
 	struct dirent **files;
 	int n, nfiles;
 	const char *datadir;
+	char devicedir[PATH_MAX] = {0};
 
 	datadir = getenv("LIBRATBAG_DATA_DIR");
 	if (!datadir)
 		datadir = LIBRATBAG_DATA_DIR;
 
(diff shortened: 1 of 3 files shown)


phomes:
I pushed an updated patch with the change to install the .device files into a 'devices' subdirectory to match the source layout.

src/libratbag-data.c
```diff
@@ -446,15 +446,18 @@ ratbag_device_data_new_for_id(struct ratbag *ratbag, const struct input_id *id)
...
 	int n, nfiles;
 	const char *datadir;
+	char devicedir[PATH_MAX] = {0};
 
 	datadir = getenv("LIBRATBAG_DATA_DIR");
 	if (!datadir)
 		datadir = LIBRATBAG_DATA_DIR;
 
-	n = scandir(datadir, &files, filter_device_files, alphasort);
+	sprintf(devicedir, "%s/devices", datadir);
```
whot: always use snprintf please, especially for unverified data

PR #412, opened 2017-10-20
Flat resolution range

phomes:


phomes:
@whot makes sense. I have pushed a new patch

src/driver-logitech-g300.c
```diff
@@ -386,16 +386,17 @@ logitech_g300_read_profile(struct ratbag_profile *profile)
 	ratbag_profile_for_each_resolution(profile, resolution) {
 		struct logitech_g300_resolution *res =
 			&report->dpi_levels[resolution->index];
+		uint dpis[10] = { 250, 500, 750, 1000, 1250, 1500, 1750, 2000,
+				 2250, 2500 };
```
whot: never seen `uint` before :) it's not a standard type, please use `unsigned int` instead. Otherwise, feel free to push this one

phomes: woops. I was mixing up the code style with another project.

PR #415, opened 2017-10-23
HID++ 2.0: Support non gaming mice again

bentiss:
We used to support them, but they are kind of crashing. Kind of fixing them then.

### DIFF
--- a/src/driver-hidpp20.c
+++ b/src/driver-hidpp20.c
@@ -716,6 +716,9 @@ hidpp20drv_read_resolution_dpi(struct ratbag_profile *profile)
 	struct ratbag_device *device = profile->device;
 	struct ratbag *ratbag = device->ratbag;
 	struct hidpp20drv_data *drv_data = ratbag_get_drv_data(device);
+	struct ratbag_resolution *res;
+	unsigned int default_dpi = 1000;
+	unsigned int default_rate = 500;
 	int rc;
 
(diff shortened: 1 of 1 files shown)


whot:
the second patch looks ok, but I'm not sure we want the first one - why do we need to support the devices  if they're clearly not meeting the requirements we have?

bentiss:
> why do we need to support the devices if they're clearly not meeting the requirements we have?

The reason is that if you happen to have one of these mouse connected to your system, libratbag (or ratbagd) will fail to start given that the assert will be raised.
An other solution would be to blacklist or whitelist Logitech devices, but this is maybe not something we want. Actually, we could require feature 0x8100 to be present, but given the MX Master can have the DPI changed, I still think we should support it somehow.

whot:
uhm, what's the status of this? sorry , I lost track, but is this waiting on me to click the merge button?

bentiss:
> uhm, what's the status of this? sorry , I lost track,

It is still needed AFAICT. I can not reproduce the assert which kills ratbagds, but the MX master requires it if we want to configure it.

>  but is this waiting on me to click the merge button?

Pretty much, yes.

r00tbox:
I switched back to the master, and hope this gets merged into the master soon, so i can use my MX Master again.

whot:
Done, thanks for testing

PR #418, opened 2017-10-26
steelseries: add writing buttons

phomes:


whot:
both the fedora build failures were caused by an rpm failure, merging

PR #422, opened 2017-11-11
Convert LED effect rate in Hz to effect duration in ms

Hjdskes:
This is a start to convert an LED's effect rate in Hz to effect duration in ms, as discussed in https://github.com/libratbag/piper/issues/175. This commit isn't working because I do not trust myself to change the drivers. I did peek at them and only driver-hidpp20.c needs changing. Feel free to take this work and squash in the changes required to the driver.


Hjdskes:
Do I need to manually change tools/ratbagc.py as well?

whot:
Closing in favour of #423

PR #423, opened 2017-11-14
Convert LED effect rate in Hz to effect duration in ms

whot:
As discussed in https://github.com/libratbag/piper/issues/175

This is the finished one from Jente in #422


Hjdskes:
Isn't this also missing the driver bits?

bentiss:
> Isn't this also missing the driver bits?

For logitech devices, I know we messed up from the start and actually exported the time in ms instead of report rate.

For the others, well, we should figure it out with the devices themselves, but I believe they are also messed up in the same way Logitech was, given that the UI was the same :)

Hjdskes:
Ah, great!

whot:
I looked at all the driver files and most of them don't export an effect rate because they either have no LEDs or they don't support the breathing/cycling mode. So there wasn't anything to fix there

Issue #433, opened 2017-12-08
ratbagd stack corruption


Budovi:
Hi,

there is a stack corruption bug in the `ratbagd` preventing the change in the profile state.
```
$ ratbagctl profile 0 enable event16
Traceback (most recent call last):                                                                                                                                                                                                           
  File "/usr/lib/python3.6/site-packages/gi/_propertyhelper.py", line 426, in obj_set_property                                                                                                                                               
    prop.fset(self, value)                                                                                                                                                                                                                   
  File "/usr/bin/ratbagctl", line 491, in enabled                                                                                                                                                                                            
    self._set_dbus_property("Enabled", "b", enabled)                                                                                                                                                                                         
  File "/usr/bin/ratbagctl", line 189, in _set_dbus_property                                                                                                                                                                                 
    2000, None)                                                                                                                                                                                                                              
GLib.Error: g-dbus-error-quark: GDBus.Error:org.freedesktop.DBus.Error.NoReply: Message recipient disconnected from message bus without replying (4)
```
Running daemon:
```
$ sudo ./ratbagd
*** stack smashing detected ***: <unknown> terminated
Aborted
```
Backtrace:
```
Program received signal SIGABRT, Aborted.
#0  0x00007ffff6d328a0 in raise () from /usr/lib/libc.so.6
#1  0x00007ffff6d33f09 in abort () from /usr/lib/libc.so.6
#2  0x00007ffff6d75517 in __libc_message () from /usr/lib/libc.so.6
#3  0x00007ffff6e0597f in __fortify_fail_abort () from /usr/lib/libc.so.6
#4  0x00007ffff6e05942 in __stack_chk_fail_local () from /usr/lib/libc.so.6
#5  0x000055555555b872 in ratbagd_profile_set_enabled (bus=<optimized out>, path=<optimized out>, interface=<optimized out>, property=<optimized out>, m=0x5555557ba650, userdata=0x5555557a66d0, error=0x7fffffffe5e0)
    at ../ratbagd/ratbagd-profile.c:336
#6  0x00007ffff7949b8c in ?? () from /usr/lib/libsystemd.so.0
#7  0x00007ffff7990b34 in ?? () from /usr/lib/libsystemd.so.0
#8  0x00007ffff7990fa4 in ?? () from /usr/lib/libsystemd.so.0
#9  0x00007ffff7977840 in ?? () from /usr/lib/libsystemd.so.0
[...]

PR #438, opened 2017-12-18
ratbagd: ensure boolean passed to sd_* are 4 bytes wide

SonicFrog:
This makes sure that every boolean passed to sd_* functions is 4 bytes long as required by the dbus specification


whot:
Linking with #433

PR #443, opened 2018-01-30
SteelSeries sensei 310

phomes:
This adds support for the SteelSeries Sensei 310. The mouse is very close to the Rival 310, but has two additional buttons and different default button mappings.

### DIFF
--- a/src/driver-steelseries.c
+++ b/src/driver-steelseries.c
@@ -63,26 +63,6 @@
 #define STEELSERIES_BUTTON_KBD			0x51
 #define STEELSERIES_BUTTON_CONSUMER		0x61
 
-static const enum ratbag_button_type button_types[6] =
-{
-	RATBAG_BUTTON_TYPE_LEFT,
-	RATBAG_BUTTON_TYPE_RIGHT,
-	RATBAG_BUTTON_TYPE_MIDDLE,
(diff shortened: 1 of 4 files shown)


src/driver-steelseries.c
```diff
@@ -161,12 +185,28 @@ steelseries_probe(struct ratbag_device *device)
...
+				actions = button_actions_6;
+			} else if (button_count == 8) {
+				types = button_types_8;
+				actions = button_actions_8;
+			} else {
+				log_debug(device->ratbag,
+					  "Device with %d buttons not supported\n",
+					  button_count);
+				return -ENOTSUP;
+			}
```
whot: Does it give us a lot of benefit to have a `_6` and a `_8` array here? I think it might be better to just build the arrays before the `for_each_button` loop, in the form of:
```
enum ratbag_button_type buttons[8] = {UNKNOWN};
buttons[0] = ...
buttons[1] = ...
if (nbuttons == 6)
  buttons[5] = ...
else
  buttons[5] = ...
```

This is just a nitpick and personal preference, so I'm not going to block on it. Would be good to check out how the code looks like with an approach like that.

phomes: Thanks. The code is clearer that way. I separated the code to set the button defaults out to a function button_defaults_for_layout.

PR #446, opened 2018-02-07
Add logitech G600 svg

gandalf3:
I've now added Button and LED leaders to the SVG. It currently appears thusly:

logitech-g600

----

How should all the side buttons on the G600 be managed? They end up crammed in at an awkward angle from the top perspective. Would it be acceptable to include a side perspective for easier viewing?
Also, does anything special need to be done to the SVG to account for the shift button?

Thanks!


whot:
> Would it be acceptable to include a side perspective for easier viewing?

I think in this case a partial side-view would be useful, similar to the detail view on printed maps where you have an area circled/outlined and that outlined area in more detail in a corner. Does that explanation make sense?

> Also, does anything special need to be done to the SVG to account for the shift button?

I don't think we have real support for that yet. And I'm not sure how exactly to handle this - anyone got any ideas? @bentiss, @Hjdskes maybe?

bentiss:
> I don't think we have real support for that yet. And I'm not sure how exactly to handle this - anyone got any ideas? @bentiss, @Hjdskes maybe?

IIRC the Roccat mouse you have already has such button. The [original design](https://github.com/libratbag/libratbag/blob/master/data/default/roccat-kone-xtd.svg) simply had both modes attached to each leader. Then the gnome refresh of the designed appeared, and the functionality [disappeared](https://github.com/libratbag/libratbag/blob/master/data/gnome/roccat-kone-xtd.svg).

IMO it's more of a design issue (then piper issue) than a libratbag issue for the moment. For the roccat, IIRC, libratbag exports the number of buttons times the modes (2). So here, I am not so sure what to do with such mode.

@Hjdskes any ideas on how we can present this to userspace (as this might involve changes in libratbag after we make the decision)

Hjdskes:
I'm not sure that a side view is going to be particularly easier to read, and if I am not mistaken @jimmac advocated against having multiple angles. Keep in mind that the mouse buttons will be highlighted when hovered. I think for now you can maybe exaggerate a little bit by making the buttons slightly larger than they should be scale-wise, so they are even/a little easier to notice.

As for the shift mode -- I haven't seen that before. I'm not sure how I missed that and why no one mentioned it earlier! I'm thinking that Piper can read when the shift key is pressed and, when it knows that the current mouse supports this, update its UI to reflect the new state. So to configure e.g. <kbd>Shift + left mouse button</kbd>, you would press <kbd>Shift</kbd> in Piper and then select the button mapping to the left mouse button.

I think this is the most transparent, as there would be no need for another layer in the SVG or for multiple leader lines and double the amount of buttons in the UI. To make this feature discoverable, the SVG's could have a little arrow or something in some corner that is highlighted when <kbd>Shift</kbd> is pressed.

gandalf3:
Here's an implementation of @whot's partial side-view idea:

logitech-g600

I'm not sure if that's exactly what you had in mind, but I think it makes the buttons quite a bit more decipherable. For what it's worth, I think it also helps the aesthetic by allowing the main mouse drawing to keep a very orthographic projection (when looking at the mouse from straight overhead you only really see the sides of the top buttons; my first version mixed and matched perspectives a bit which looked odd to me after staring at it long enough).

Hjdskes:
That looks quite nice, but none of us are UX experts so we might not be the best people to decide. @whot do we want to consult with @jimmac?

gandalf3:
@Hjdskes Are you talking about <kbd>Shift</kbd> on the keyboard? Or the "G shift" button on the mouse?
If the latter, is it even possible to read that button's state? Iirc the logitech tool just has a checkbox/UI switch for switching between showing/setting the normal binds and the "G shifted" binds.

Hjdskes:
I was referring to the G shift button, sorry if the markup made it confusing. If libratbag can expose this button's state, what I wrote above should be possible.

whot:
@gandalf3 yes, that's pretty much what I had in mind. I would experiment with moving this around a bit, it may not need to be immediately next to the buttons. You could - like in paper maps - draw a shape (like a circle) around the buttons and have that surround the button configuration.

This one has the side-view but not the outline what the details represent, but it's close enough:
https://viennamap360.com/carte/image/de/stadtrundfahrt-wien.jpg

CC @jimmac, in case he ignored the previous comment ;)

Hjdskes:
It just occurred to me that with that placement, you're going to have trouble with the leader lines. You're also going to run into issues with all the buttons in the UI, I think: there'd simply be too many of them.

Faalagorn:
In Lioncast Windows configuration tool for their non-MMO LM20 mouse you have to press a button to switch between front and side buttons. I guess it's the same for their MMO equivalent Lioncast LM30 mouse. Not sure if it's a good practice to follow, but it is one way to deal with it.

You can look in Google for the images and videos with the software how it looks like. I can also provide some screenshots, but I'd have to boot into Windows for that.

gandalf3:
I've now added leaders to the SVG. In order to try and manage the button overload on the left side, I overlayed the leaders for each "column" of buttons on top of one another. My assumption is that when a button is to be highlighted, its group is brought to the top and the fill/stroke colors changed. If that is indeed what happens, then I think having them overlayed won't be too confusing. Just in case, I also labeled each leader at the button end. Each label is also part of its respective leader group for added highlighting.

The main downside I can see so far is that the overlayed leaders become "thicker" looking than the non overlayed ones, because the partially transparent anti-aliasing pixels at the edges of the leader line add up and become darker than they otherwise would be. I'm not sure how to fix that, but it seems pretty minor.

Hjdskes:
> My assumption is that when a button is to be highlighted, its group is brought to the top and the fill/stroke colors changed.

The fill/stroke colors _of the button_ are changed and brought to the top (IIRC), but the leader lines are not highlighted nor brought to the top.

> The main downside I can see so far is that the overlayed leaders become "thicker" looking than the non overlayed ones, because the partially transparent anti-aliasing pixels at the edges of the leader line add up and become darker than they otherwise would be. I'm not sure how to fix that, but it seems pretty minor.

This can also be considered an advantage because it indicates that there are several overlapped lines there. I don't think this has to be a problem.

Can you test this SVG in Piper @gandalf3?

gandalf3:
@Hjdskes I'm not sure how to convince piper to display it without having any compatible profile for my mouse. Am I missing something?

Rikitik:
Any progress? i'm super new to librattbag (installed today) i have no idea how to use it. Piper says no device found when i connect the Logitech G600 so not sure if this is normal? It would nice if i know how to use ratbagd without piper. to check if the device is even seen

FFY00:
@Rikitik just run `ratbagctl` to see if any device is detected.

You can run `ratbagctl -h` to see all available commands as well as some examples. If after this you still need help with `ratbagctl`, just let me know :smile:.

Rikitik:
@FFY00 thank you. unfortunately ratbagctl says "No devices available"however the mouse can be used and shows up in libusb. So i'm not sure what's up

Hjdskes:
This means there is currently no driver for the G600 yet, so there is nothing you can do until there is one.

phomes:
The leader lines do not show up in piper. When I check the .SVG with the "check-svg.py" script I get the following errors. Can you take a look?

ERROR:logitech-g600.svg:Missing button: button0
ERROR:logitech-g600.svg:Missing button: button1
ERROR:logitech-g600.svg:Missing button: button2
INFO:logitech-g600.svg:Found 0 buttons
INFO:logitech-g600.svg:Found 0 leds

gandalf3:
Ah yes, I haven't set the group ids yet; I wasn't sure how to number the buttons. I'll be happy to name them, but I'm not sure how to tell which buttons should have which id (assuming it matters?)

phomes:
0. left
1. right
2. wheel
3. wheel left
4. wheel right
5. the extra button on the right side
6. profile cycle (G8)
7. resolution cycle (G7)
8. G9
9. G10
...
19. G20

gandalf3:
@phomes Thanks, I've added some things, but it doesn't completely pass check-svg.py yet. I'll probably be able to get it passing sometime tomorrow. Also, I notice scroll up and down aren't included in your list. Just to be sure, is that correct?

Rikitik:
My guess is that wheel is one button and scroll up and down is the behavior of the button not an action on itself

gandalf3:
Now passes check-svg.py without errors

phomes:
Really nice. It is starting to look very good in piper too. We will need a bit more space between the leader lines though:

g600

gandalf3:
@phomes  Nice! It looks like it's going to be a tight squeeze on the left side, but I'll see what I can do. Is there a way I can preview it in piper?

gandalf3:
@phomes  Oh perfect, taking a look now :)

phomes:
I pushed my WIP driver in #496. Do you know how to compile and test libratbag and piper from source?

Hjdskes:
> We will need a bit more space between the leader lines though:

Perhaps check-svg.py should include a check for a minimum vertical spacing between leader lines.

gandalf3:
I've been looking at it a bit, and the GTK theme I'm using seems to have slimmer drop downs; for me they are only overlapping by about 1px. Should I just make sure it works with a particular theme? (if so, which one?) Or should piper handle this differently somehow?

Hjdskes:
Ensure that it works with Adwaita please, the default GTK theme.

phomes:
The PR to add the driver no longer adds references to the svg. Please add those to your PR. The references must be added for [meson.build](https://github.com/libratbag/libratbag/blob/26493aceaf094b508cf637f3f006c01f8338c4d0/meson.build#L303) and the [device file](https://github.com/libratbag/libratbag/blob/26493aceaf094b508cf637f3f006c01f8338c4d0/data/devices/logitech-g600.device#L5)

phomes:
Looks good to me now. Note that we do not get a notification when you push new commits so I only noticed your updates now.

Here is a preview with adwaita:
piper-g600

PR #452, opened 2018-02-14
steelseries: support the original Rival

phomes:
Initial support for the original Rival mouse.
Setting resolutions, report rate and LED color and
modes.

No key mappings yet, and breathing duration is fixed.


whot:
build error isn't our fault

Issue #271, opened 2017-08-02
Add macro capabilities


whot:
Add capabilities so a client can detect which macros can be configured on the device. This becomes important when we remove KeyMapping support to handle that transparently.

bentiss:
Well, just to be more clear, the 2 Logitech HID++ don't support macro, the roccat driver, etekcity and gskill do support them IIRC.

phomes:
For the steelseries mice I have the following restrictions on macros:
- Only 1 key press (no real macro support onboard)
- No timeouts can be used obviously
- Max 3 modifiers used in combination (limitation in protocol)
- Keys can be either keyboard or hid consumer

FFY00:
Should the key mapping be removed internally too, or just from the API?

Keeping it internally allows drivers to simplify some of the code. Drivers would choose if they make use of it or not.

Issue #243, opened 2017-07-24
Drop the plain key mapping


whot:
Let's drop the key mappings and expose only the macro mappings. For devices that cannot handle macros but can do the key mapping (???) we can transparently map the macros to the right function.

gbdlin:
Macro mapping and key mapping are totally different... You cannot really set for mouse button to hold a key as long as you're holding button with a macro.

PR #454, opened 2018-02-14
Add MacroLength to .device files

phomes:
I need to store in the .device file for the steelseries mice whether they support mapping buttons to keys or not. I cannot query it and it the feature does not follow any of the other settings, like DeviceVersion.

The steelseries mice do not support storing real macros on the device itself, only key+modifiers. This can be seen as a single-step macro. Since it seems we will need a global option for macro length anyway I thought that using a macro length of 0 could be used to indicate that no keys can be set.

I have only updated the .device files for the steelseries mice in this PR.

Is it a reasonable option? And is there a better name than MacroLength?

### DIFF
--- a/data/devices/steelseries-kinzu-v2.device
+++ b/data/devices/steelseries-kinzu-v2.device
@@ -9,3 +9,4 @@ DeviceVersion=1
 Buttons=3
 Leds=0
 DpiList=400;800;1600;3200
+MacroLength=0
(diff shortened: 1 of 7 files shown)


phomes:
Related to #271 and #243

whot:
the keys in the config file are private API and can change at any time (short of the possible annoyance of having to run sed). I would leave this as a private key for the steelseries driver for now, with the bonus of you can rename it to something more explicit.

phomes:
No problem. The patch is updated to move it to the steelseries section.

PR #455, opened 2018-02-16
dbus service: use the actual ratbagd service name and not the alias

yaneti:
The alias can be used only when the service is enabled

### DIFF
--- a/dbus/org.freedesktop.ratbag1.service.in
+++ b/dbus/org.freedesktop.ratbag1.service.in
@@ -2,4 +2,4 @@
 Name=org.freedesktop.ratbag1
 Exec=@bindir@/ratbagd
 User=root
-SystemdService=dbus-org.freedesktop.ratbag1.service
+SystemdService=ratbagd.service


bentiss:
I am not that familiar with dbus activation services. The point of having the full dbus name was to enable the dbus activation, so I am not so sure that your change will break it or not.

@whot, any ideas?

whot:
@yaneti please check commit b164fa622f951deebd6d85a8026cbccabed9a7ed. It's a while ago and I can't remember the details but presumably I did this for a reason :)

I think the effect of this was being able to have ratbagd.service as a separately-controlled entity from the dbus service (i.e. you could run ratbagd.service without dbus activation too).

yaneti:
@whot, So I don't know if older systemd / dbus worked in a different way but here with 
```
systemd-237-1.git78bd769.fc28.x86_64
dbus-1.12.0-1.fc28.x86_64
```
the change allows dbus activation to actually start the ratbagd.service when its not explicitly enabled. It works that way because unit aliases are only active once the service is enabled. example:
```
# systemctl status ratbagd.service
● ratbagd.service - Daemon to introspect and modify configurable mice
   Loaded: loaded (/usr/lib/systemd/system/ratbagd.service; disabled; vendor preset: disabled)
# ratbagctl 
Unable to connect to ratbagd over dbus

log:
Feb 21 01:41:36 d2 dbus-daemon[781]: [system] Activating via systemd: service name='org.freedesktop.ratbag1' unit='dbus-org.freedesktop.ratbag1.service' requested by ':1.779' (uid=0 pid=26954 comm="/usr/bin/python3 /usr/bin/ratbagctl " label="unconfined_u:unconfined_r:unconfined_t:s0-s0:c0.c1023")
Feb 21 01:41:36 d2 dbus-daemon[781]: [system] Activation via systemd failed for unit 'dbus-org.freedesktop.ratbag1.service': Unit dbus-org.freedesktop.ratbag1.service not found.
```

```
# systemctl enable ratbagd.service
Created symlink /etc/systemd/system/dbus-org.freedesktop.ratbag1.service → /usr/lib/systemd/system/ratbagd.service.
# systemctl status ratbagd.service 
● ratbagd.service - Daemon to introspect and modify configurable mice
   Loaded: loaded (/usr/lib/systemd/system/ratbagd.service; enabled; vendor preset: disabled)
   Active: inactive (dead)
# ratbagctl 
event5:    Logitech Gaming Mouse G300

log:
Feb 21 01:41:48 d2 dbus-daemon[781]: [system] Activating via systemd: service name='org.freedesktop.ratbag1' unit='dbus-org.freedesktop.ratbag1.service' requested by ':1.780' (uid=0 pid=26988 comm="/usr/bin/python3 /usr/bin/ratbagctl " label="unconfined_u:unconfined_r:unconfined_t:s0-s0:c0.c1023")
Feb 21 01:41:48 d2 systemd[1]: Starting Daemon to introspect and modify configurable mice...
Feb 21 01:41:48 d2 dbus-daemon[781]: [system] Successfully activated service 'org.freedesktop.ratbag1'
Feb 21 01:41:48 d2 systemd[1]: Started Daemon to introspect and modify configurable mice.
```
With the proposed change there is no need to explicitly enable the service
```
# systemctl status ratbagd.service 
● ratbagd.service - Daemon to introspect and modify configurable mice
   Loaded: loaded (/usr/lib/systemd/system/ratbagd.service; disabled; vendor preset: disabled)
   Active: inactive (dead)
# ratbagctl 
event5:    Logitech Gaming Mouse G300      
```

whot:
> the change allows dbus activation to actually start the ratbagd.service when its not explicitly enabled.

but that's the point of the alias, otherwise you cannot ever disable the service because dbus will always start it. With the alias, you have dbus activation but still leave the administrator free to disable the service.

whot:
@yaneti can you submit a PR for the deprecation warnings? should just be a simple search/replace

yaneti:
Well, I don't quite understand why ratbagd as a service should have to be explicitly enabled or disabled. Basically it should just work whenever one uses ratbagctl,  the desktop does something with the dbus interface  or some udev hook for specific hardware decides to start it.  And it should probably exit by itself if it doesn't find any support devices that it needs to manage

In a pinch disabling it should be done by masking.

Other examples:   org.freedesktop.Accounts , org.freedesktop.GeoClue2, org.freedesktop.fwupd, org.freedesktop.PackageKit..

yaneti:
PR #458  for the warnings

whot:
> Well, I don't quite understand why ratbagd as a service should have to be explicitly enabled or disabled. 

we're changing system-wide hardware state. Being able to disable it seemed like a good idea. But reading up on masking now - yeah, that seems like the correct solution over enable/disable.

PR #457, opened 2018-02-19
Helpers for conversion betwen macro and key+modifier binding

phomes:
This expands the approach in #402 to also cover modifier keys, and goes for a full conversion to macros instead of converting to/from the key binding action. The drivers can use the functions to convert between a key+modifier bit field and a single key press macro.

The second patch is a conversion of the logitech-g300 driver.

### DIFF
--- a/src/driver-logitech-g300.c
+++ b/src/driver-logitech-g300.c
@@ -151,6 +151,54 @@ logitech_g300_raw_to_button_action(uint8_t data)
 	return NULL;
 }
 
+static uint8_t logitech_g300_modifier_to_raw(int modifier_flags)
+{
+	uint8_t modifiers = 0x00;
+
+	if (modifier_flags & MODIFIER_LEFTCTRL)
+		modifiers |= 0x01;
+	if (modifier_flags & MODIFIER_LEFTSHIFT)
+		modifiers |= 0x02;
+	if (modifier_flags & MODIFIER_LEFTALT)
+		modifiers |= 0x04;
+	if (modifier_flags & MODIFIER_LEFTMETA)
+		modifiers |= 0x08;
+	if (modifier_flags & MODIFIER_RIGHTCTRL)
+		modifiers |= 0x10;
+	if (modifier_flags & MODIFIER_RIGHTSHIFT)
+		modifiers |= 0x20;
+	if (modifier_flags & MODIFIER_RIGHTALT)
+		modifiers |= 0x40;
+	if (modifier_flags & MODIFIER_RIGHTMETA)
+		modifiers |= 0x80;
+
+	return modifiers;
+}
+
+static int logitech_g300_raw_to_modifiers(uint8_t data)
+{
+	int modifiers = 0;
+
+	if (data & 0x01)
+		modifiers |= MODIFIER_LEFTCTRL;
+	if (data & 0x02)
+		modifiers |= MODIFIER_LEFTSHIFT;
+	if (data & 0x04)
+		modifiers |= MODIFIER_LEFTALT;
+	if (data & 0x08)
+		modifiers |= MODIFIER_LEFTMETA;
+	if (data & 0x10)
+		modifiers |= MODIFIER_RIGHTCTRL;
+	if (data & 0x20)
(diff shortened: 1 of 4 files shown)


src/libratbag.c
```diff
@@ -1909,3 +1909,132 @@ ratbag_button_macro_new(const char *name)
...
+			case KEY_RIGHTSHIFT: modifiers &= ~MODIFIER_RIGHTSHIFT; break;
+			case KEY_RIGHTALT: modifiers &= ~MODIFIER_RIGHTALT; break;
+			case KEY_RIGHTMETA: modifiers &= ~MODIFIER_RIGHTMETA; break;
+			default:
+				if (event.event.key != key)
+					return -EINVAL;
+
+				*key_out = key;
+				*modifiers_out = modifiers;
+				return 1;
```
bentiss: Shouldn't you check before returning if there are no other key pressed?
I can imagine a macro with `Ctrl-C Ctrl-V` that would be translated into `Ctrl-V` this way.

src/driver-logitech-g300.c
```diff
@@ -151,6 +151,54 @@ logitech_g300_raw_to_button_action(uint8_t data)
 	return NULL;
 }
 
+static uint8_t logitech_g300_modifier_to_raw(int modifier_flags)
+{
+	uint8_t modifiers = 0x00;
+
+	if (modifier_flags & MODIFIER_LEFTCTRL)
+		modifiers += 0x01;
```
bentiss: `|=` would be preferable if you are working on bit fields

Also do the same on the ones below

src/driver-logitech-g300.c
```diff
@@ -151,6 +151,54 @@ logitech_g300_raw_to_button_action(uint8_t data)
...
+
+	return modifiers;
+}
+
+static int logitech_g300_raw_to_modifiers(uint8_t data)
+{
+	int modifiers = 0;
+
+	if (data & 0x01)
+		modifiers += MODIFIER_LEFTCTRL;
```
bentiss: same here (and down below)

src/driver-logitech-g300.c
```diff
@@ -517,9 +576,22 @@ logitech_g300_write_profile(struct ratbag_profile *profile)
 		raw_button->code = logitech_g300_button_action_to_raw(action);
 		raw_button->modifier = 0x00;
 		raw_button->key = 0x00;
-		if (action->type == RATBAG_BUTTON_ACTION_TYPE_KEY) {
+
+		if (action->type == RATBAG_BUTTON_ACTION_TYPE_MACRO) {
```
bentiss: You should probably check here that the macro is compatible and is not a 'long' macro.

src/libratbag.c
phomes: The function is only intended to be used when we use a macro to do a simple keybinding. We should have caps (#271) that the client should respect so we do not try to set a `Ctrl-C Ctrl-V` on a device which can only do one key. I added a new guard to `ratbag_action_keycode_from_macro` which checks that the macro only contain one key.

src/driver-logitech-g300.c
phomes: ratbag_action_keycode_from_macro will now return an error for a long macro.

phomes:
I added a patch to port steelseries to use macros

src/driver-logitech-g300.c
```diff
@@ -151,6 +151,54 @@ logitech_g300_raw_to_button_action(uint8_t data)
...
+	if (modifier_flags & MODIFIER_LEFTMETA)
+		modifiers |= 0x08;
+	if (modifier_flags & MODIFIER_RIGHTCTRL)
+		modifiers |= 0x10;
+	if (modifier_flags & MODIFIER_RIGHTSHIFT)
+		modifiers |= 0x20;
+	if (modifier_flags & MODIFIER_RIGHTALT)
+		modifiers |= 0x40;
+	if (modifier_flags & MODIFIER_RIGHTMETA)
+		modifiers |= 0x80;
```
bentiss: We probably should have a BIT() macro equivalent to the kernel one :/

PR #461, opened 2018-03-06
It seems the driver is for Etekcity Scroll Alpha

Faalagorn:
Based on the [etekcity.svg](https://github.com/libratbag/libratbag/blob/master/data/default/etekcity.svg) and https://www.phoronix.com/scan.php?page=news_item&px=Libratbag-Announced
Need confirmation from @dvdhrm and @bentiss based on http://libratbag.github.io/device-notes/ to check IDs, unless it's for both both.

### DIFF
--- a/data/devices/etekcity-scroll-alpha.device
+++ b/data/devices/etekcity-scroll-alpha.device
@@ -1,6 +1,5 @@
-# FIXME: which device is this exactly?
 [Device]
-Name=Etekcity FIXME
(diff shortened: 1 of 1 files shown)


bentiss:
yes, the etekcity driver was for the Scroll Alpha. Unfortunately, we realized that Etekcity was just branding some mice from other vendors, and there is no common protocol between it mice. So this driver may very well be only for this particular mouse.

PR #466, opened 2018-03-24
steelseries: sleep between messages

phomes:
Adding a short sleep between calls makes the writes to the device more reliable.

I noticed that occasionally the mouse would only pick up some of the changed settings. The official configurator waits 10 ms between each message and 20 ms before the save. Adding these here seems to have made the writes more reliable.

### DIFF
--- a/src/driver-steelseries.c
+++ b/src/driver-steelseries.c
@@ -256,6 +256,7 @@ steelseries_write_dpi(struct ratbag_resolution *resolution)
 		buf[6] = 0x42; /* not sure if needed */
 	}
 
+	msleep(10);
 	ret = ratbag_hidraw_output_report(device, buf, buf_len);
 	if (ret < 0)
 		return ret;
@@ -282,6 +283,7 @@ steelseries_write_report_rate(struct ratbag_resolution *resolution)
 		buf[2] = 1000 / resolution->hz;
 	}
 
+	msleep(10);
 	ret = ratbag_hidraw_output_report(device, buf, buf_len);
 	if (ret < 0)
 		return ret;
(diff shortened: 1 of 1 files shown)


phomes:
@FFY00 have you noticed any problems?

phomes:
You did a very nice analysis of the protocol. Do you see any needed changes to the [implementation in libratbag](https://github.com/libratbag/libratbag/blob/master/src/driver-steelseries.c#L433)?

FFY00:
Yes, right now we can't get the value of the 2nd dpi. We should also implement the trigger option and allow the user to customize the breath/colorshift.

You should check https://github.com/flozz/rivalcfg/pull/43 again, I added a full specification of the protocol, including the trigger configuration.

phomes:
Oh, you expanded the analysis and documentation. Great work :+1: 

I am not sure I follow what you mean by not getting the 2nd DPI. We cannot get any values at all from the device yet. We do set both of the DPI, so I am not sure what you mean?

Libratbag is an abstraction across a large range of mice. We expose the common functionality. None of the other mice have a Trigger LED feature, and it is thus not configurable with the GUI in Piper. For that reason I simply did not implement it.

FFY00:
Yeah, sorry. I have the 1st dpi value set to 800, which is the value I got from the dpi get command so I thought it was working.

Do you have a slack/telegram/discord/irc channel for the project? I would be happy to help with the development.

whot:
> Do you have a slack/telegram/discord/irc channel for the project?

#libratbag on freenode, but expect it to be somewhere between empty and quiet most of the time

FFY00:
@whot yes, I am already there but thanks anyway. The only person I was able to talk to was bentiss and it still was a bit hard.

PR #467, opened 2018-03-26
Add support for trigger mode

FFY00:


phomes:
@whot @bentiss Do we want to support a Trigger mode for LEDs? The LED will light up for a set duration after clicking the configured button/key. I have not seen this feature on any device other than steelseries.

bentiss:
I think the Logitech mice have some advanced trigger mechanism for the macros too, but IIRC, it's only linked to the button that initiates the macro. And now that I think about it, I do not think the macro could control the LEDs.

So my initial thought would be that we should hold off until more devices are supporting this, because we don't know if the API is the correct one based on only one device. For the record, we already removed features in the past because they were only for one particular device/brand, and maintaining too many features is something we do not want to be confronted with.

FFY00:
@bentiss I totally understand. If it helps, I would be happy to maintain this feature.

FFY00:
I am going to close this at it seems the changes won't be merged anytime soon.

Issue #436, opened 2017-12-14
Sensitivity is unusably high and non-configurable on Rival 310


ahstro:
I just tried configuring my Rival 310 using libratbag (through Piper). All I changed was trying to set the scroll LED to solid orange and turning the SteelSeries logo off. The result I got was both LEDs turning off and the sensitivity going through the roof. Playing around with sensitivity or resolution in Piper does not change anything, nor does clicking the resolution button on the mouse itself.

phomes:
That is strange. I tested on mine again and I do not see any problems. Setting the LEDs work and it picks up the DPI set in Piper. Note that we cannot get the current resolution already on the mouse, so Piper will always start with some default settings. But that does not explain why the LEDs do not set correctly.

One theory could be different versions of firmware on our mice. I am working on a way to show the current version.

phomes:
I ran into a similar issue when working on support for the Sensei 310. I had to do a firmware update to get it working. An important thing to note is that it was also not working properly with steelseries own software. They pop up a message to install the "critial firmware update".

phomes:
When the mouse is detected and the driver is loaded we normally will query the mouse to get the current onboard settings. This seems to not be possible on steelseries. I have not found a way yet at least. So instead the idea is that the client (ratbagctl, piper, etc) will have the responsibility of storing the last configuration on disk and load that into the driver. This part is not yet implemented, so we always fall back to the default values that I arbitrarily chose. The two resolutions are set to 800 and 1600 DPI.

We also have dirty flags on parts of the ratbag profile. E.g. a LED or button will have a dirty flag to indicate whether it has been changed. We use these to skip setting parts that have not been changed. However, the resolutions do not have a dirty flag to set, so we always have to set them when we make any other change. That is why changing a LED will also happen to reset the resolution.

So we could solve this in three ways. Find a way to query the device. Implement the on disk storage of settings. Or by adding a dirty flag for the resolution so we can skip them.

It would be great if you want to tackle one of them. For the dirty flags you can see the sturcts here:
https://github.com/libratbag/libratbag/blob/master/src/libratbag-private.h#L251
ratbag_resolution does not have a dirty flag, while ratbag_led just below does.

FFY00:
I'll try to add the dirty flag but that's not the disk storage of settings should also be implemented.

phomes:
Yes, the dirty flag will only help for the very first write where we do not have any settings on disk to start from.

FFY00:
[...]

PR #468, opened 2018-03-26
Add dirty flag in resolution

FFY00:
Solves https://github.com/libratbag/libratbag/issues/436

I was only able to check in my Steelsereis Rival 310. Please test on other mice.

### DIFF
--- a/src/libratbag-private.h
+++ b/src/libratbag-private.h
@@ -267,6 +267,7 @@ struct ratbag_resolution {
 
 	bool is_active;
 	bool is_default;
+	bool dirty;
 	uint32_t capabilities;
 };
 
--- a/src/driver-steelseries.c
+++ b/src/driver-steelseries.c
@@ -580,6 +580,9 @@ steelseries_write_profile(struct ratbag_profile *profile)
 	int rc;
 
 	ratbag_profile_for_each_resolution(profile, resolution) {
+		if (!resolution->dirty)
+			continue;
+
 		rc = steelseries_write_dpi(resolution);
 		if (rc != 0)
 			return rc;
--- a/src/libratbag.c
+++ b/src/libratbag.c
@@ -977,6 +977,7 @@ ratbag_device_commit(struct ratbag_device *device)
 	struct ratbag_profile *profile;
 	struct ratbag_button *button;
 	struct ratbag_led *led;
+	struct ratbag_resolution *resolution;
 	int rc;
 
 	if (!device->driver->commit) {
@@ -997,6 +998,10 @@ ratbag_device_commit(struct ratbag_device *device)
 
 		list_for_each(led, &profile->leds, link)
 			led->dirty = false;
+
+		list_for_each(resolution, &profile->resolutions, link)
+			resolution->dirty = false;
+			
 	}
 
 	return RATBAG_SUCCESS;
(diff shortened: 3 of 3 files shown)


src/driver-hidpp20.c
```diff
@@ -1166,6 +1166,9 @@ hidpp20drv_commit(struct ratbag_device *device)
 		drv_data->profiles->profiles[profile->index].enabled = profile->is_enabled;
 
 		ratbag_profile_for_each_resolution(profile, resolution) {
+			if (!resolution->dirty)
+				continue;
+
```
bentiss: This test needs to be placed after `dpi_index = resolution->index;` or changing any other setting will switch the resolution to the first one (`dpi_index` will stay at `0`)

src/libratbag-private.h
```diff
@@ -267,6 +267,7 @@ struct ratbag_resolution {
 
 	bool is_active;
 	bool is_default;
+	bool dirty;			/**< profile changed since last commit */
```
bentiss: `s/profile/resolution/`

src/driver-etekcity.c
```diff
@@ -457,6 +457,9 @@ etekcity_read_profile(struct ratbag_profile *profile)
 	}
 
 	ratbag_profile_for_each_resolution(profile, resolution) {
+		if (!resolution->dirty)
+			continue;
+
```
bentiss: you are in `etekcity_read_profile()`, you probably shouldn't be checking for the dirty flag.

Also, etekcity and roccat are both using the old write profile stubs, and so they do not need to check for the dirty flag as you are already checking it in `libratbag.c`

src/driver-gskill.c
```diff
@@ -1113,6 +1113,9 @@ gskill_update_resolutions(struct ratbag_profile *profile)
 			&report->dpi_levels[report->dpi_num];
 
 		res = ratbag_profile_get_resolution(profile, i);
+		if (!res->dirty)
+			continue;
+
```
bentiss: gskill starts by clearing the memory allocated with the resolutions. I am sure you are basically resetting the DPI now. gskill shouldn't check for the dirty flag.

src/driver-hidpp10.c
```diff
@@ -574,6 +574,9 @@ hidpp10drv_commit(struct ratbag_device *device)
 
 		/* Update the current resolution in case it changed */
 		if (active_resolution) {
+			if (!resolution->dirty)
+				continue;
+
```
bentiss: This must be OK, not entirely sure however

src/driver-logitech-g300.c
```diff
@@ -549,6 +549,9 @@ logitech_g300_write_profile(struct ratbag_profile *profile)
 	report = &pdata->report;
 
 	ratbag_profile_for_each_resolution(profile, resolution) {
+		if (!resolution->dirty)
+			continue;
+
```
bentiss: Not entirely sure if this will break or not the current resolution...

src/driver-roccat.c
```diff
@@ -731,6 +731,9 @@ roccat_read_profile(struct ratbag_profile *profile)
 	}
 
 	ratbag_profile_for_each_resolution(profile, resolution) {
+		if (!resolution->dirty)
+			continue;
+
```
bentiss: Same than etekcity, you are in `read_profile()`

FFY00:
Thanks for the input. I guided myself mostly from the other dirty flag checks. I will make the necessary changes when I get home.

FFY00:
I only left the check on the steelseries driver.

phomes:
Can you squash the patches and force push a new patch?

The sum of the two patches create a white space change in gskill.
The change in ratbag_old_write_profile should also be reverted.

FFY00:
@phomes done.

FFY00:
Hum, weird. Can someone restart the CI test?

PR #472, opened 2018-04-03
Implement a structure for cycle

FFY00:
This should make the code easier to be maintained.

### DIFF
--- a/src/driver-steelseries.c
+++ b/src/driver-steelseries.c
@@ -76,6 +76,21 @@ struct steelseries_data {
 	int firmware_minor;
 };
 
+struct steelseries_point {
+	struct list link;
+
+	struct ratbag_color color;	/* point color */
+	uint8_t pos;			/* relative position in the cycle */
+};
+
+struct steelseries_led_cycle {
+	uint8_t led_id;			/* led id */
+	uint16_t duration;		/* cycle duration */
+	bool repeat;			/* if the cycle restarts automatically */
+	uint8_t trigger_buttons;	/* trigger button combination */
+	struct list points;		/* colors in the cycle */
+};
+
 static int
 steelseries_test_hidraw(struct ratbag_device *device)
 {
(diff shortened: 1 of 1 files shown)


FFY00:
Should I allocate the points on the stack? From the code I've seen, you seem to usually allocate this kind of things in the heap but I am not sure why.

phomes:
Thanks. The change to append to the end of lists instead of inserting at the start should probably be discussed in its own PR, as it affects all users of the current API.

Can you squash the remaining 3 patches into 1?

FFY00:
I changed the method to append to the end instead of the start because it was needed here https://github.com/FFY00/libratbag/blob/334251806ee9c82b872b5b70d7e0117e1a74601b/src/driver-steelseries.c#L109. But yes, I agree, it should be discussed in a separated PR. What do you mean by squashing the 3 patched into 1? You want me to make a single commit with all the changes?

phomes:
A single commit for the patches makes more sense as they make changes to code just introduced in the same PR.

FFY00:
Okay, I'll open other pull request for the `list_insert`. This PR stays on hold since it depends on that change.

FFY00:
Just bumping this up. If anyone has some time to review the changes it would be appreciated.

src/driver-steelseries.c
```diff
@@ -440,98 +508,84 @@ steelseries_write_led_v2(struct ratbag_led *led)
...
+
+		point2.color.red = 0x00;
+		point2.color.green = 0xFF;
+		point2.color.blue = 0x00;
+		point2.pos = 0x54;
+
+		point3.color.red = 0x00;
+		point3.color.green = 0x00;
+		point3.color.blue = 0xFF;
+		point3.pos = 0x54;
```
whot: is this a typo? should this be `0x55`? seems strange that we have `0x00 0x54 0x54 0x56`

[later]
Ok, this is a time-share. Why is it called pos? and if it's a normalized time-share why is it in hex? The three add up to 254, is this normalized to 255? If so, one is missing ;)

src/driver-steelseries.c
```diff
@@ -440,98 +508,84 @@ steelseries_write_led_v2(struct ratbag_led *led)
...
-	uint16_t duration, min_duration;
 	int ret;
 
-	buf[0] = STEELSERIES_ID_LED;
-	buf[2] = led->index;
+	struct steelseries_cycle cycle;
+	struct steelseries_point point1;
+	struct steelseries_point point2;
+	struct steelseries_point point3;
+	struct steelseries_point point4;
```
whot: an array would be better here tbh, makes things more obvious than separate variables.

src/driver-steelseries.c
```diff
@@ -65,6 +65,74 @@
...
+		buf[19] = 0x01;
+
+	buf[23] = cycle->trigger_buttons;
+
+	list_for_each(point, &cycle->points, link) {
+		if(point_count == 0){
+			buf[28] = point->color.red;
+			buf[29] = point->color.green;
+			buf[30] = point->color.blue;
+		}
```
whot: I'd appreciate a comment here to make clear that this isn't a missing a `continue;` statement here and that we really want to write the first point at two locations.

src/driver-steelseries.c
```diff
@@ -65,6 +65,74 @@
...
+
+		buf[31 + point_count*4] = point->color.red;
+		buf[32 + point_count*4] = point->color.green;
+		buf[33 + point_count*4] = point->color.blue;
+		buf[34 + point_count*4] = point->pos;
+
+		point_count++;
```
whot: please assert if `buf[...]` exceeds the buffer size

src/driver-steelseries.c
```diff
@@ -440,98 +508,84 @@ steelseries_write_led_v2(struct ratbag_led *led)
...
+
+		point3.color.red = 0x00;
+		point3.color.green = 0x00;
+		point3.color.blue = 0xFF;
+		point3.pos = 0x54;
+
+		point4.color.red = 0xFF;
+		point4.color.green = 0x00;
+		point4.color.blue = 0x00;
+		point4.pos = 0x56;
```
whot: you're writing red twice? once at 0x00 and once at 0x56. Is this correct?

src/driver-steelseries.c
```diff
@@ -440,98 +508,84 @@ steelseries_write_led_v2(struct ratbag_led *led)
...
 		return -EINVAL;
 	}
 
-	/* this seems to be the minimum allowed */
-	min_duration = buf[27] * 330;
-	if (duration < min_duration)
-		duration = min_duration;
-
-	hidpp_set_unaligned_le_u16(&buf[3], duration);
+	construct_cycle_buffer(&cycle, buf);
```
whot: for safety, always pass the buffer size with a buffer so you can easily assert in the function

src/driver-steelseries.c
```diff
@@ -65,6 +65,74 @@
...
+		point_count++;
+	}
+
+	buf[27] = point_count;
+
+	/* this seems to be the minimum allowed */
+	min_duration = buf[27] * 330;
+	duration = cycle->duration;
+	if (duration < min_duration)
+		duration = min_duration;
```
whot: we should have a `min` macro, so you can write
```
duration = max(min_duration, cycle->duration)
```

src/driver-steelseries.c
```diff
@@ -440,98 +508,84 @@ steelseries_write_led_v2(struct ratbag_led *led)
...
-		buf[27] = 0x01;
-		buf[28] = buf[31] = 0x00;
-		buf[29] = buf[32] = 0x00;
-		buf[30] = buf[33] = 0x00;
-		/* not sure why the duration is set for the steady color or why
-		   it is different for the two LEDs */
-		duration = led->index == 0 ? 10000 : 5000; /* 0x1027 or 0x8813 */
+		point1.color.red = 0x00;
+		point1.color.green = 0x00;
+		point1.color.blue = 0x00;
```
whot: A good and easy-to-read solution here would be:
````
const struct ratbag_color black  = { 0x00 };
point1.color = black;
````
and then repeat that for the other cases. Makes it immediately obvious what things are set to.

File: src/driver-steelseries.c [hunk 9]
```diff
@@ -65,6 +65,74 @@
...
+
+	list_init(&cycle->points);
+}
+
+static void
+construct_cycle_buffer(struct steelseries_cycle *cycle, uint8_t *buf)
+{
+	struct steelseries_point *point;
+	uint16_t duration, min_duration;
+	uint8_t point_count = 0;
```
whot: just a nitpick, I think `npoints` is more natural here but it really doesn't matter :)

src/driver-steelseries.c
FFY00: Yes. When the cycle resets, it goes directly from 255 to 0. If the color isn't the same, it will just change, without any fade.

src/driver-steelseries.c
FFY00: I don't really know, this is how the official driver does it I think. I didn't touch any of the values, I just changed the implementation to use `construct_cycle_buffer()`.

I called it pos as in the position in the cycle. I can change it if you want. It made more sense to me to think of it as the position offset rather than a delay, since the the actual duration of the cycle is variable.

src/driver-steelseries.c
FFY00: Will do.

src/driver-steelseries.c
FFY00: Yes, I should have done that, sorry.

src/driver-steelseries.c
FFY00: The buffer size is always the same. It's given by `STEELSERIES_REPORT_SIZE`. I can assert to that.

src/driver-steelseries.c
whot: calling it pos would be ok if there's a comment that explains what the position is that it refers to. This is hw-specific stuff and it's reverse-engineered, so always err on the side of too many comments that explain what you wanted to do because there's always the chance of misinterpretation. It's easy to fix something that's clearly wrong than something where you have to guess what the original intention was. Compare the fictitious example of:
```
buf[25] = 0xac; /* checksum for buf[0 - 10] */
````
with just an assignment of 
```
buf[25] = 0xac;
```
In the former you can immediately change it once you figure out that it's not the checksum (or from a different slice). The latter you're always left wondering *why* the original author set this buffer to 0xac and whether there was some intrinsic need for it.

File: src/driver-steelseries.c (hunk 6 above)
Comment (whot) [reply src/driver-steelseries.c]: pass the size in and your function becomes more flexible and less prone to erroneous copy/paste. Or even other changes when suddenly a profile buffer is different to a report buffer or whatever. The typing cost of passing the buffer size is close to zero, migth as well go for it :)

**Comment by FFY00** [comment]
Hey, sorry for bothering again. Are any more changes needed?

**Comment by phomes** [comment]
I thought that there was still an outstanding review wrt the 0x54 pos, but I can see that has changed. I will give the code a test run on my devices. Should be fine to go in for me after that test

File: src/driver-steelseries.c [hunk 10]
```diff
@@ -65,6 +65,78 @@
 #define STEELSERIES_BUTTON_KBD			0x51
 #define STEELSERIES_BUTTON_CONSUMER		0x61
 
+struct steelseries_point {
+	struct list link;
+
+	struct ratbag_color color;	/* point color */
```
phomes: Please align the comments so they fit for a tab size of 8 spaces

src/driver-steelseries.c
```diff
@@ -65,6 +65,78 @@
...
+	uint8_t pos;				/* relative position in the cycle */
+};
+
+struct steelseries_cycle {
+	uint8_t led_id; 			/* led id */
+	uint16_t duration;			/* cycle duration */
+	bool repeat;				/* if the cycle restarts automatically */
+	uint8_t trigger_buttons;	/* trigger button combination */
```
phomes: Please align the comments so they fit for a tab size of 8 spaces

src/driver-steelseries.c
```diff
@@ -65,6 +65,78 @@
...
 #define STEELSERIES_BUTTON_CONSUMER		0x61
 
+struct steelseries_point {
+	struct list link;
+
+	struct ratbag_color color;	/* point color */
+	uint8_t pos;				/* relative position in the cycle */
+};
+
+struct steelseries_cycle {
```
phomes: can we rename it to steelseries_led_cycle

src/driver-steelseries.c
```diff
@@ -65,6 +65,78 @@
...
+	uint8_t led_id; 			/* led id */
+	uint16_t duration;			/* cycle duration */
+	bool repeat;				/* if the cycle restarts automatically */
+	uint8_t trigger_buttons;	/* trigger button combination */
+	struct list points;			/* colors in the cycle */
+};
+
+static void
+create_cycle(struct steelseries_cycle *cycle)
```
phomes: rename to create_led_cycle. Also please move it to just above steelseries_write_led_v2

src/driver-steelseries.c
```diff
@@ -65,6 +65,78 @@
...
+	cycle->led_id = 0x00;
+	cycle->duration = 5000;
+	cycle->repeat = true;
+	cycle->trigger_buttons = 0x00;
+
+	list_init(&cycle->points);
+}
+
+static void
+construct_cycle_buffer(struct steelseries_cycle *cycle, uint8_t *buf, uint8_t buf_size)
```
phomes: rename to construct_led_cycle_report. Also please move it to just above steelseries_write_led_v2

src/driver-steelseries.c
```diff
@@ -65,6 +65,78 @@
...
+			/* this is not a mistake,
+				we need to write the first
+				point to 28,29,30 */
+			buf[28] = point->color.red;
+			buf[29] = point->color.green;
+			buf[30] = point->color.blue;
+		}
+
+		cycle_size += point->pos;
+		assert(cycle_size < 256);
```
phomes: also assert that buf[34 + npoints*4] is within buf_size

src/driver-steelseries.c
```diff
@@ -440,98 +512,72 @@ steelseries_write_led_v2(struct ratbag_led *led)
...
 		return -EINVAL;
 	}
 
-	/* this seems to be the minimum allowed */
-	min_duration = buf[27] * 330;
-	if (duration < min_duration)
-		duration = min_duration;
-
-	hidpp_set_unaligned_le_u16(&buf[3], duration);
+	construct_cycle_buffer(&cycle, buf, STEELSERIES_REPORT_SIZE);
```
phomes: sizeof(buf) would make this more flexible if we have to use a different report size in the future

src/driver-steelseries.c
```diff
@@ -65,6 +65,78 @@
...
+		buf[33 + npoints*4] = point->color.blue;
+		buf[34 + npoints*4] = point->pos;
+
+		npoints++;
+	}
+
+	buf[27] = npoints;
+
+	/* this seems to be the minimum allowed */
+	duration = min_threshold(cycle->duration, buf[27] * 330);
```
phomes: min_threshold is the same as a max, so why not use that?

FFY00:
Hey, I am not home this week and I dony have access to a computer. I'll fox this when I get back (should be 28/29/30 of this month).

FFY00:
I totally forgot about this. These last days have been crazy, have been rarely home. Right now I'm home but I'll need to leave tomorrow. I'm gonna start working on the patch, should be done today.

FFY00:
Sorry for bothering but can someone review the last changes? @phomes @whot

phomes:
Sorry about the late review. We are not notified when changes are pushed. You need to ping every time you have pushed a new version.

The updated code looks good to go in for me now. There is however a conflict with the latest changes to the driver that landed recently. Can you resolve those and push again (and ping :) )

FFY00:
Rebased against `origin/master`. The conflicts are now fixed. @phomes

FFY00:
@phomes can you give the final approval?

phomes:
Looks good. Thanks @FFY00

PR #474, opened 2018-04-05
libratbag: fix order when inserting new element in list

FFY00:
Required for https://github.com/libratbag/libratbag/pull/472


whot:
Thanks, but tbh, I'd rather have this as new call `list_append` to make it clear what the behaviour is. Code itself is fine, so fixing this is just a matter of moving things around.

FFY00:
Yeah, sure! I've made a new commit.

PR #482, opened 2018-05-22
remove clang analyzer false positive warnings

bentiss:
https://bugs.llvm.org/show_bug.cgi?id=3888

clang analyzer doesn't know how to handle ```__attribute__(cleanup())```.

There are only 2 warnings raised because the analyzer only look at the current compilation unit for checking the malloc call. So only those 2 are detected by clang and then marked as memory leak.

This also means our current continuous integration doesn't check for memory leaks.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -124,13 +124,13 @@ scan_build_run: &scan_build_run
     - *fedora_install
     - run:
         name: Install clang and find
-        command: dnf install -y clang-analyzer findutils
+        command: dnf install -y clang-analyzer findutils python3-beautifulsoup4
     - checkout
     - *ninja_scan_build
     - *export_logs
     - run:
         name: Check scan-build results
-        command: test ! -d ~/libratbag/build/meson-logs/scanbuild || test $(find ~/libratbag/build/meson-logs/scanbuild -maxdepth 0 ! -empty -exec echo "not empty" \; | wc -l) -eq 0 || (echo "Check scan-build results" && false)
+        command: python3 ~/libratbag/tools/check_scan_build.py ~/libratbag/build/meson-logs/scanbuild
 
 doc_build: &doc_build
   <<: *default_settings
--- a/src/libratbag-data.c
+++ b/src/libratbag-data.c
@@ -388,7 +388,7 @@ file_data_matches(struct ratbag *ratbag,
 	data->refcount = 1;
 	data->name = g_key_file_get_string(keyfile, GROUP_DEVICE, "Name", NULL);
 	if (!data->name) {
-		return false;
+		return false; // ignore_clang_sa_mem_leak
 	}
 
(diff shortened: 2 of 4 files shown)


src/libratbag-data.c
```diff
@@ -384,7 +384,13 @@ file_data_matches(struct ratbag *ratbag,
 	if (!match(id, match_strv))
 		return false;
 
+#ifndef __clang_analyzer__
 	data = zalloc(sizeof(*data));
+#else
+	/* clang static analyzer doesn't know how to handle _cleanup_
+	 * https://bugs.llvm.org/show_bug.cgi?id=3888 */
+	data = 1;
+#endif
```
whot: wtf? How does it not complain about this when we dereference it in the next line?

src/hidpp20.h
```diff
@@ -787,7 +787,13 @@ hidpp20_onboard_profiles_write_sector(struct hidpp20_device *device,
 static inline uint8_t *
 hidpp20_onboard_profiles_allocate_sector(struct hidpp20_profiles *profiles)
 {
+#ifndef __clang_analyzer__
 	return zalloc(profiles->sector_size);
+#else
+	/* clang static analyzer doesn't know how to handle _cleanup_
+	 * https://bugs.llvm.org/show_bug.cgi?id=3888 */
+	return 1;
+#endif
```
whot: I'd probably go with 
```
#ifdef __clang_analyzer__
return 1
#endif
```
instead to reduce the cognitive load on the normal branch. Even though it's just a single line here.

bentiss:
Second attempt, adding some comments in the code and make sure circle.yml doesn't complain about scanbuild results that are tagged as false positives.

This is cleaner. I hope I did not messed up the things in circle.yml, we'll see in the long run I guess

circle.yml
```diff
@@ -130,7 +130,12 @@ scan_build_run: &scan_build_run
     - *export_logs
     - run:
         name: Check scan-build results
-        command: test ! -d ~/libratbag/build/meson-logs/scanbuild || test $(find ~/libratbag/build/meson-logs/scanbuild -maxdepth 0 ! -empty -exec echo "not empty" \; | wc -l) -eq 0 || (echo "Check scan-build results" && false)
+        command: |
+          test ! -d ~/libratbag/build/meson-logs/scanbuild || (for r in ~/libratbag/build/meson-logs/scanbuild/*/report*.html
+          do
+            line=$(grep -Po 'a href="#EndPath">line \K(\d)+' $r)
+            grep -o "LN${line}\">.*ignore_clang_sa_" $r || exit 1
+          done)
```
whot: I think we're at the point where we should move this into a script as part of the git repo and just have circleCI supply the directory. Much easier to debug this way. Other than that, it looks good though.

bentiss:
I guess the rev-by still stands. Merging this so we have better CI for the future PR.

PR #485, opened 2018-06-17
Implement Steelseries Sensei Raw

elFarto:
This PR implements the Steelseries Sensei Raw. This device has a monochromatic LED which required a new command to be implemented (along with a device property, MonoLed to enable it). It also has a simpler key mapping command (basically no modifiers are allowed), which required tying a number of properties to a single device property (ShortButton).

I'm not sure I've implemented the features in the best way, let me know if there's a better way to do it.

I also had a go at the device image but I'm not sure it turned out great, I'm no artist.


elFarto:
Ok, macros are disabled, and breathing mode for the LEDs is supported and working. I had initially thought it could do macros, but it turned out it was just the Steelseries engine on windows doing it.

I had forgotten to add that ratbag doesn't quite support an adjustable brightness mono LED correctly (or an RGB LED with separate brightness like I have in another mouse). It all works, but from what I understand the brightness field is associated with the breathing mode, not the solid mode.

FFY00:
What do you mean? The brightness is controlled by the black channel in each color. Does this work differently on your mouse? That would be a bit weird.

elFarto:
For the Sensei Raw, there's nothing odd. Just a mono LED with 4 brightness settings. It could just be Piper that's put the brightness setting under the breathing tab.

The other mouse I'm working on (a Coolermaster one) has RGB settings (with 256 steps) and a separate brightness setting (with 64 steps).

phomes:
I think the .svg looks reasonable, but @Hjdskes usually is better at judging those than I am :) One thing to fix though is to remove the image layer.

Does everything work well from Piper with the device?

Hjdskes:
The colors in the SVG are off - take a look at sny of the others to see what I mean. I think it's best if all SVGs use the same palette. The semi transparent background you have now might also blend in too much, making it hard to actually see the image in Piper.

Also nitpick: the scroll wheel seems off-center. Other than that, nice work!

elFarto:
Found the issue with the SVG, I'd left the device layer half transparent while I was tracing around the photo. I've tweaked the colours and centred everything.
Everything works in Piper, with the oddity that the brightness for the LED is in the Breathing tab, not the Solid tab (although it works with both).

Hjdskes:
> Everything works in Piper, with the oddity that the brightness for the LED is in the Breathing tab, not the Solid tab (although it works with both).

This oddity is intentional, IIRC not all devices support setting a brightness for the solid mode. I'll open an issue for this in Piper so I will look into it when I can pick up development again.

Hjdskes:
As for the SVG, can you make the coloring consistent to e.g. https://github.com/libratbag/libratbag/blob/master/data/gnome/logitech-g303.svg? That is, using different shades of gray for different parts of the mouse. It's not critical, but I believe consistency is key when it comes to these images. It might also make it easier to distinguish different parts for the user.

Hjdskes:
The SVG looks good now, awesome!

PR #489, opened 2018-06-30
meson: generate pkgconfig file for libratbag

FFY00:
Generate a pkgconfig file so that libratbag can be easily used as a dependency in other projects.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -175,6 +175,14 @@ dep_libratbag = declare_dependency(
 	dependencies : deps_libratbag
 )
 
+pkgconfig.generate (
+        filebase: 'libratbag',
+        name: 'Libratbag',
+        description: 'A DBus daemon to configure gaming mice',
+        version: meson.project_version(),
+        libraries: lib_libratbag
+)
+
 #### libshared.a ####
 src_libshared = [
 	'tools/shared.c',


whot:
Do we have a specific use-case here? libratbag used to be a library (hence the name) with a separate dbus server (ratbagd). But eventually we found that that architecture doesn't make a lot of sense, you need root access and the only sensible way to abstract that away is a dbus server. ratbagd is generic enough that it can be used by any application, so I'm not sure exposing libratbag as a real library makes much sense.

FFY00:
I see this as the easiest way to ensure libratbag is installed. Right now, piper doesn't check at all if it is installed. Piper gets successfully installed even if libratbag is missing and it results in errors when trying to run. Maybe we don't need to expose the library but we should provide a way to make sure libratbag is installed. Providing a pkgconfig file, even without a library (I don't know if that's possible but I assume it is), allows build systems to know the path prefix of the libratbag instalation. This, for example, removes the need for hardcoded path lists in projects using libratbag's assets (svg files, etc.). I don't know if I made myself much clear, it's 4am and I'm on my phone :grin:.

whot:
hmm. pkg-config files are usually shipped by the development packages and are only applicable during build time and their primary reason is to sort out CFLAGS and the linker arguments. And we really have no build-time dependency on libratbag and we don't want to expose any of the internal libraries.

I think what you want to do (and a laudable thing it is :) would be better with a `find_program()` at build-time. If we check for `ratbagd` to exist, we can at least put a massive warning out.

FFY00:
I agree. pkg-config was a bad way to achieve this but the idea persists.

I am going to close this as there's to be implemented here. This needs to be fixed `piper`'s repository.

PR #490, opened 2018-07-02
Include Logitech G102/G203 device files

lrenaud:
Just a quick change to add support for the G102 and G203 mice.

These mice are functionally identical and share a device ID, they only differ in marketing materials and which countries sell each mouse.

### DIFF
--- a/data/devices/logitech-g102-g203.device
+++ b/data/devices/logitech-g102-g203.device
@@ -1,0 +1,7 @@
+# G102 and G203 (USB)
+[Device]
+Name=Logitech Gaming Mouse G102/G103
+DeviceMatch=usb:046d:c084
+Svg=logitech-g102-g203.svg
+Driver=hidpp20
+LedTypes=logo;side;
--- a/data/gnome/logitech-g102-g203.svg
+++ b/data/gnome/logitech-g102-g203.svg
@@ -1,0 +1,1 @@
+logitech-g303.svg
--- a/meson.build
+++ b/meson.build
@@ -244,6 +244,7 @@ man_lur_command = configure_file (
 data_files = files(
 	'data/devices/etekcity-scroll-alpha.device',
 	'data/devices/gskill-MX-780.device',
+	'data/devices/logitech-g102-g203.device',
 	'data/devices/logitech-g300.device',
 	'data/devices/logitech-g303.device',
 	'data/devices/logitech-g403.device',
@@ -290,6 +291,7 @@ test('data-parse-test', data_parse_test,
 
 gnome_svg_files = files(
 	'data/gnome/fallback.svg',
+	'data/gnome/logitech-g102-g203.svg',
 	'data/gnome/logitech-g300.svg',
 	'data/gnome/logitech-g303.svg',
 	'data/gnome/logitech-g403.svg',


FFY00:
I think th svg should be renamed to include both names.

lrenaud:
I didn't generate a new SVG file for the mouse. We have a few other mice where the SVG references some other mouse's SVG file, which is what I've done here.

Given that, is it still preferable to duplicate the SVG for now?

Hjdskes:
Symlinks are the preferred way to solve this.

FFY00:
Please correct me if I am wrong but we shouldn't hardcode the file name when using it in the code, right? It should be provided by the `.device` file. If this is correct, I don't see the point of creating a symlink since we can only have one `Svg` field in the `.device` file. And in that case, I think it would make more sense to have a file named `vendor-device1-device2.svg` instead of on named `vendor-device1.svg` providing the image for both `device1` and `device2`.

If there's no problem hardcoding the `.svg` file names, I think this design would make more sense.

`.../libratbag/devices/vendor-device1-device2.device`
```
[Device]
Name=Vendor Device1/Device2
Svg=vendor-device1-device2.svg
...
```

`.../libratbag/gnome/`
```
vendor-device1.svg -> vendor-device1-device2.svg
vendor-device2.svg -> vendor-device1-device2.svg
vendor-device1-device2.svg
```

Maybe I am making a big deal out of nothing but that seems to me like a bad way to handle things.

lrenaud:
There are two issues at play here. It is my understanding of @Hjdskes's comment that the symlink would be used in place of a proper G203/G102 SVG until one was made.

The second issue, and the one more relevant to your (@FFY00) question is naming of the device/svg file supporting the the G102 and G203 devices. The dual naming convention makes more sense to me as it also avoids potential confusion while we are simultaneously reusing the G303 SVG.

FFY00:
We both agree about the naming convention. What I don't understand is the need for symlinks since (I assume) we shouldn't hardcode the `.svg` names. The `.device` should be used to figure it them out instead.

@whot can you have a look at this when you have time?

lrenaud:
I think using the .device name to resolve the .svg files is a wise route in the long run, but wouldn't that make symlinks necessary regardless? When we add a new .device file we create the .svg file, and if an SVG doesn't exist yet then a symlink to a suitable replacement is put in place until one is made available. The alternate of course is to resolve to the fallback.svg file.

Either way, it sounds like rebuilding the system to resolve .device to .svg is beyond the scope of the current pull request. I've pushed the device file name changes to my branch.

FFY00:
Seems good to me! You should squash the commits into one before merging.

lrenaud:
Well, learning how to squash commits was a minor adventure.

Hopefully that means everything is in place.

FFY00:
Okay, I think it's ready for merging.

Hjdskes:
You should also include the svg symlink to meson.build.

lrenaud:
Sorry about that. Thank you for catching the bug. I've updated the meson.build and re-squashed the changes.

FFY00:
@lrenaud can you rebase instead of merging? I don't even know why is merge needed here.

FFY00:
No problem. If you need any help just let me know. You can find my email in my profile :smile:.

PR #496, opened 2018-07-15
G600

phomes:
This is a driver for the logitech G600. The logic is fairly close to that of the G300, but it was easier to start developing it as a separate driver. The result of trying to combine the two drivers was a large 
if/else code due to the differences in the report messages and a few differences in the features (LEDs mostly). I will be happy to merge it with the G300 if we do not want to keep them separate.

The driver still has a few bugs when used from Piper. The active resolution is not set when changing profile. And the duration for the LED effect does not have the correct max duration (16000 ms). There are probably more issues as the button setting has only been tested lightly. I am mostly posting it so @gandalf3 can start testing it and do the final adjustments of the SVG in #446 

This PR assumes that the .SVG is already present in the data/gnome dir


gandalf3:
Thanks! Looks good on first glance, LED and basic button binding and LED configuration seem to be working so far. Basic DPI configuration also seems to work.

* I see it's possible to bind a button to "battery check". Is this a leftover in the protocol from other mice, seeing as this is a wired mouse?
* Piper doesn't seem aware of which profile the mouse is in, is it supposed to? If I remember correctly, the official logitech client does synchronize the state.
* I can bind a button to the "G Shift" key (second mode), but I don't see a way to set up binds which only fire when second mode is active?

Rikitik:
Hows it looking guyz?

phomes:
There is a list of special mouse button functions which include "Battery level", "Ratchet Mode", "Profile Up", etc. Some of these may not make sense for certain devices. We should probably have a capability setting for those in some way.
Does piper not show the current active profile in the top left corner? It also shows the active profile in ratbagctl for me. 
"Second mode" feature is not support in Piper, so I have not tried to implement it in the driver yet.

gandalf3:
@phomes I see. Regarding the active profile, `ratbagctl profile active get` returns the profile selected in piper, but doesn't change when the profile is changed via buttons on the mouse. Also, selecting a profile in piper enables piper's apply button, but updates profile active on the mouse *before* the apply button is clicked (possibly a piper issue?).

bentiss:
Feel free to push the branch once most of the CI succeeds

Hjdskes:
Before we merge this, perhaps take a look at https://github.com/libratbag/piper/issues/220 which seems to be caused by this driver.

phomes:
I updated the driver with a correction for setting the active resolution on profile change in Piper. I also dropped the references to the .svg to pass CI. Those references should be added in #446

phomes:
#220 seems unrelated to the driver and CI passes so let's merge this

PR #503, opened 2018-07-30
update logitech G9 SVG

phomes:
When looking into why setting macros are not working on the G9 (HID++1-0) I noticed that the G9 svg was missing the wheel left/right buttons, and that thus some buttons were also numbered wrong.

Adds wheel left/right clicks and adjusts button numbering.

(Button5 seems to be configurable in the profile, but no such button
exists on the mouse)


phomes:
I added a patch to reduce it to a warning so CI can test that option

PR #504, opened 2018-08-07
ratbagd: send enums as signed int

### LINKED ISSUE #281: Libratbag or ratbagd should perform a range check on button_{g,s}et_special
For some unknown reason my device now has a special button mapping with value 4294967295; this number crashes Piper because it's not recognized in the bindings so there is 1) no RatbagdButton.ACTION_SPECIAL_FOO and 2) no entry in the map that translates these values to their human-readable strings with that value as key (ratbagctl is broken for the same reason). Granted, I could add a try/catch block around the map, but this might be better solved here where this value should be detected in either libratbag's or ratbagd's setter and getter methods to set/get ACTION_SPECIAL_UNKNOWN instead on any unknown value.

phomes:
Some of our enums use negative values for an invalid
setting. E.g. RATBAG_BUTTON_ACTION_SPECIAL_INVALID = -1

To allow using the same values in piper we should send the
enum values as signed ints.

Fixes #281

### DIFF
--- a/ratbagd/ratbagd-button.c
+++ b/ratbagd/ratbagd-button.c
@@ -129,7 +129,7 @@ static int ratbagd_button_get_special(sd_bus *bus,
 
 	special = ratbag_button_get_special(button->lib_button);
 
-	return sd_bus_message_append(reply, "u", special);
+	return sd_bus_message_append(reply, "i", special);
 }
 
 static int ratbagd_button_set_special(sd_bus *bus,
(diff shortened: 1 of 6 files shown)


Hjdskes:
Also, I'm not sure how the dbus api is documented but you may have to update this as well.

phomes:
Patch updated with changes for ratbagctl similar to those in piper. CI passes now

whot:
Hmm, I'm looking at this and now I'm wondering: why don't we have a layer in ratbagd that filters all these invalid values out? We *probably* need to add something like an Error event in the DBus API so that piper can display something useful to the user but this to me feels like we're keeping ratbagd simple and offloading all the error handling into piper. When really, we should be in control of the DBus API that we never send out-of-range vales.

That piper doesn't check the values and trusts them blindly is a different issue.

Hjdskes:
With https://github.com/libratbag/libratbag/pull/516 merged, I'll close this. Please reopen if needed.

PR #506, opened 2018-08-16
tools: make ratbagd a bit more robust regarding None object path lists

whot:
Not sure why this is happening exactly, but it's better to handle this case
without an immdiate exception.

Fixes https://github.com/libratbag/piper/issues/228


whot:
Done, piper commit 3e66aeb9e13c9ac0ccbdb424dcb4f72449bd7bc7

PR #509, opened 2018-08-16
meson: conditionalize systemd unit file installation

Johnnynator:
libratbag works fine also on systems that use elogind instead of full blown systemd.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -50,7 +50,7 @@ pkgconfig = import('pkgconfig')
 dep_udev = dependency('libudev')
 dep_libevdev = dependency('libevdev')
 dep_libsystemd = dependency('libsystemd', version : '>=227')
-dep_systemd = dependency('systemd')
+dep_systemd = dependency('systemd', required: false)
 dep_glib = dependency('glib-2.0')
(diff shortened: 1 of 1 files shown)


meson.build
```diff
@@ -437,35 +437,39 @@ configure_file(input : 'dbus/org.freedesktop.ratbag_devel1.conf.in',
...
+			'''.format(default_unitdir))
+			unitdir = default_unitdir
+		else
+			unitdir = intended_unitdir
+		endif
 	endif
 endif
 
 config_bindir = configuration_data()
 config_bindir.set('bindir', join_paths(get_option('prefix'), get_option('bindir')))
```
whot: are we using config_bindir anywhere else? If not, we can move it into the if condition, then you don't need to check for `dep_systemd.found()` twice.

Johnnynator: True. Has to be fixed in master then.

Johnnynator:
Closed since #510 got merges

meson.build
whot: nah, doesn't need to be fixed, just checked and it's used for the dbus service file too

PR #515, opened 2018-08-21
ci: add arch

FFY00:


FFY00:
I don't understand why ratbagctl-test isn't being skipped. It expects a file that is only present after instalation.

@whot, any ideas?

whot:
looks good, but I wonder why the CI failed. Can you double-check that please?

FFY00:
Acording to meson's documentation the executable should exit with code `77` so it gets marked as skipped. They dond't mention any alternnative on why a test would be marked as skipped.

http://mesonbuild.com/Unit-tests.html#skipped-tests

FFY00:
Can someone look into this? Whe building libratbag on a clean chroot, the test is skipped. Don't know why the same behavior isn't present here.

FFY00:
Shouldn't that be handled by the test?

FFY00:
Well that had something to do with it. Now the test is timming out.

whot:
maybe the dbus service doesn't start? or is otherwise restricted? Not sure but you should be able to reproduce this locally with a docker image, just run the same commands and see if it works out.

FFY00:
The dbus service should already be running.

https://www.archlinux.org/news/d-bus-now-launches-user-buses/

FFY00:
I'll deal with this tomorrow.

FFY00:
Here's whats going wrong:
```
ratbagd.devel error: Failed to start ratbagd: Permission denied
```

FFY00:
@whot can you give me access to circleci? Specifically SSH permissions so that I can debug this issue?

whot:
I honestly can't figure out how to do that. You don't show up anywhere in CircleCI (others like phomes do when I click on the invite link). Can you see libratbag in https://circleci.com/add-projects? Maybe you need to create a CircleCI account (or log in with github) once to be known to the system?

Have you tried the local docker approach? that may also be able to reproduce the issue

FFY00:
Probably because I am not in the organization? I tried running the circleci cli tool but that doesn't seem to allow us to interact with the build.

whot:
yep, that was it. you were in libratbag but not in the org.

FFY00:
So, it looks like https://github.com/libratbag/libratbag/blob/23ff4e753d68267096ea66e729337bb3653fc650/ratbagd/ratbagd.c#L553 is being triggered. No clue why. Need to do some debugging.

PR #516, opened 2018-08-22
Don't send -1 down the wire from ratbagd

whot:
Supersedes #504 

Check each unsigned int we're sending down the dbus API and verify it's not -1. If it is, print an error to the log so we can debug this quicker. This results in the proper DBus error codes so piper can handle the various issues correctly. Right now, it just crashes when it gets a -1 where it expects an unsigned int.

### DIFF
--- a/ratbagd/ratbagd-button.c
+++ b/ratbagd/ratbagd-button.c
@@ -58,6 +58,8 @@ static int ratbagd_button_get_type(sd_bus *bus,
 
 	t = ratbag_button_get_type(button->lib_button);
 
+	verify_unsigned_int(t);
+
 	return sd_bus_message_append(reply, "u", t);
 }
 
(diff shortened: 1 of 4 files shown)


ratbagd/ratbagd-button.c
```diff
@@ -128,6 +128,8 @@ static int ratbagd_button_get_special(sd_bus *bus,
 	enum ratbag_button_action_special special;
 
 	special = ratbag_button_get_special(button->lib_button);
+	if (special == RATBAG_BUTTON_ACTION_SPECIAL_INVALID)
+		special = RATBAG_BUTTON_ACTION_SPECIAL_UNKNOWN;
```
phomes: We should log when we fix up RATBAG_BUTTON_ACTION_SPECIAL_INVALID. It is likely a driver issue that should be fixed

whot: done!

PR #518, opened 2018-08-22
ratbagd: do not warn when converting invalid special action

phomes:
In the review of #516 I requested error logging when we converted an invalid special action to an unknown. That turned out to be a mistake.  Clients e.g. ratbagctl ask for all properties, so this getter function gets called often even if the button is not a special action. I only thought that we would call the getter of the actual button type, and that each conversion from invalid to unknown would thus be due to an internal confusion of the type. I was wrong.

We do not want to send the value for an invalid special action as
that is negative 1. Converting to unknown is good enough here so
let's just remove the error logging and convert in silence.


whot:
could copy the ratelimit logging bits from libinput as well if we care sufficiently. But this works just fine, thanks

PR #519, opened 2018-08-24
Make ratbagd and ratbagd.py more robust for errors

whot:
First patch makes sure we don't crash out if we ever get None from an array property - which can happen when an error occurs.

The second patch wraps all (well, most) the `sd_bus` calls that are pure scaffolding into a macro that logs an error if they fail. This way we at least get notified when a basic "open message" or "append value" call fails, causing a property to end up being None.

### DIFF
--- a/ratbagd/ratbagd-button.c
+++ b/ratbagd/ratbagd-button.c
@@ -137,7 +137,9 @@ static int ratbagd_button_get_special(sd_bus *bus,
 
 	verify_unsigned_int(special);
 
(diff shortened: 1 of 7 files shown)


Hjdskes:
Can you sync these ratbagd.py changes to Piper, too?

whot:
On Fri, Aug 24, 2018 at 01:01:49AM -0700, Jente Hidskes wrote:
> Can you sync these ratbagd.py changes to Piper, too?

done, thanks.

PR #520, opened 2018-08-26
Read input reports from different endpoint

phomes:
This is a bit of work in progress, but is still usable as is. The motivation for pushing this now is the plans for changing from udev for flatpak support in ratbagd which would require changes in the same code.

This PR will add support in libratbag for those devices that receive reports on one endpoint but send the response on another. This is done by mice from both steelseries and cooler master. In this PR we added the functionality needed in libratbag itself and starts to make use of it from the steelseries driver.

More work is needed to read a full profile from the steelseries mice, but @FFY00 has already understood a good part of the protocol needed. I would love to see followup patches with additional data read, and more testing of device/firmware differences.

### DIFF
--- a/src/libratbag-hidraw.c
+++ b/src/libratbag-hidraw.c
@@ -1223,7 +1223,7 @@ static int
 ratbag_hidraw_parse_report_descriptor(struct ratbag_device *device)
 {
 	int rc, desc_size = 0;
-	struct ratbag_hidraw *hidraw = &device->hidraw;
+	struct ratbag_hidraw *hidraw = &device->hidraw[0];
 	struct hidraw_report_descriptor report_desc = {0};
 	unsigned int i, j;
 	unsigned int usage_page, usage;
@@ -1291,7 +1291,7 @@ ratbag_hidraw_parse_report_descriptor(struct ratbag_device *device)
 }
 
 static int
-ratbag_open_hidraw_node(struct ratbag_device *device, struct udev_device *hidraw_udev)
+ratbag_open_hidraw_node(struct ratbag_device *device, struct udev_device *hidraw_udev, int idx)
 {
 	struct hidraw_devinfo info;
 	struct ratbag_device *tmp_device;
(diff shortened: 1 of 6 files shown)


src/libratbag-hidraw.c
```diff
@@ -1300,15 +1300,15 @@ ratbag_open_hidraw_node(struct ratbag_device *device, struct udev_device *hidraw
 	const char *sysname;
 	size_t reports_size;
 
-	device->hidraw.fd = -1;
+	device->hidraw[idx].fd = -1;
```
whot: this looks like it needs some OOB check or an assert here

src/libratbag-hidraw.c
```diff
@@ -1584,14 +1601,20 @@ ratbag_hidraw_output_report(struct ratbag_device *device, uint8_t *buf, size_t l
...
+}
+
+int
+ratbag_hidraw_read_input_report_index(struct ratbag_device *device, uint8_t *buf, size_t len, int hidrawno)
 {
 	int rc;
 	struct pollfd fds;
 
-	if (len < 1 || !buf || device->hidraw.fd < 0)
+	if (len < 1 || !buf || device->hidraw[hidrawno].fd < 0)
```
whot: same here, needs boundaries checks

src/libratbag-hidraw.c
```diff
@@ -1504,20 +1515,26 @@ ratbag_hidraw_get_usage(struct ratbag_device *device, unsigned int report_id)
...
 ratbag_close_hidraw(struct ratbag_device *device)
 {
-	if (device->hidraw.fd < 0)
+	ratbag_close_hidraw_index(device, 0);
+}
+
+void
+ratbag_close_hidraw_index(struct ratbag_device *device, int idx)
+{
+	if (device->hidraw[idx].fd < 0)
```
whot: same here, needs boundaries checks

phomes:
updated with asserts

PR #524, opened 2018-08-30
build: support building with elogind

Cogitri:
#510 adds support for building without systemd itself, but still probes for libsystemd. However, elogind by default doesn't install libsystemd.pc but only libelogind.pc.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -50,9 +50,14 @@ config_h.set_quoted('LIBRATBAG_DATA_DIR', libratbag_data_dir)
 pkgconfig = import('pkgconfig')
 dep_udev = dependency('libudev')
 dep_libevdev = dependency('libevdev')
-dep_libsystemd = dependency('libsystemd', version : '>=227')
 dep_glib = dependency('glib-2.0')
 dep_lm = cc.find_library('m')
+
+if get_option('logind-provider') == 'elogind'
+	logind_dep = dependency('libelogind', version : '>=227')
+else
+	logind_dep = dependency('libsystemd', version : '>=227')
+endif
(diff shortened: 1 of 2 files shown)


Cogitri:
The elogind option could also be integrated into the systemd option.

meson_options.txt
```diff
@@ -33,3 +33,9 @@ option('systemd',
   value : true,
   description : 'Build systemd unit files'
 )
+
+option('elogind',
```
whot: I think a better option here would be something like `logind-provider=systemd|elogind`. More expressive, especially since this is an either/or, so it should be a meson `combo` option.

meson.build
```diff
@@ -50,10 +50,15 @@ config_h.set_quoted('LIBRATBAG_DATA_DIR', libratbag_data_dir)
...
 dep_udev = dependency('libudev')
 dep_libevdev = dependency('libevdev')
-dep_libsystemd = dependency('libsystemd', version : '>=227')
 dep_glib = dependency('glib-2.0')
 dep_lm = cc.find_library('m')
 
+if get_option('elogind')
+	logind_provider_dep = dependency('libelogind', version : '>=227')
+else
+	logind_provider_dep = dependency('libsystemd', version : '>=227')
```
whot: oh, and this can just be `logind_dep` imo, that's clear enough too. Not that it matters that much.

whot:
sorry about the delay, github doesn't send email when you update a branch. merged now, thanks!

Cogitri:
No worries, thanks for your suggestions :)

PR #541, opened 2018-09-13
Fix coverity

whot:
no, not coverity fixes, but really "fix coverity". It doesn't like gcc7 and spews things like undefined type _Float128 by the millions (you can never have enough of those errors, I guess).

So let's give libratbag a special "be nice to coverity" option and let meson define all the types coverity is really unhappy about.


bentiss:
@bentiss says: yep

PR #544, opened 2018-09-24
Add Logitech G PowerPlay wireless configuration

jit26:
Tested with a Logitech G903, should also work with a Logitech G703.
May also work with other Logitech mice that work with a G900 receiver.


FFY00:
You mean Logitech's Lightspeed interface right?

PR #546, opened 2018-09-26
tools: remove the unnecessary prefix from the error code enum

whot:
It's already encoded as part of the enum, no need to have it again for the
value


whot:
done, see piper commit 81b6be06ac1fd998d531

Issue #526, opened 2018-09-02
No active resolution for M705


davidovitch:
I have a Logitech M705 mouse (sharing the same unified receiver with a Logitech MK700 keyboard), but the M705 does not have an `active resolution`.

This is what happens:
```
$ ratbagctl list
event4:    Logitech M705

$ ratbagctl profile active get event4
0

$ ratbagctl resolution active get event4
No active resolution. Please report this bug to the libratbag developers
0

$ ratbagctl dpi get event4
No active resolution. Please report this bug to the libratbag developers
0dpi

$ ratbagctl profile 0 get event4
Profile 0: (active)
Name : 'Profile 1'
Resolutions:
  0: <disabled>
Button: 0 type unknown is mapped to none
Button: 1 type unknown is mapped to none
Button: 2 type unknown is mapped to none

$ ratbagctl resolution 0 get event4
0: 800dpi @ 0Hz
```

UPDATE: For every `no active resolution` return message from librattag I get an `ratbagd[15026]: ratbagd error: event4: failed to fetch SVG` reported in the journal logs (systemd).

I am new to libratbag but wouldn't mind helping out debugging this issue if someone can point me in the right direction.

Hjdskes:
See https://github.com/libratbag/piper/issues/239

Please try with the latest libratbag from git to see if that solves your problem.

davidovitch:
Same outcome with latest git master. Not sure if it is relevant, but I am on Arch Linux and the [`libratbag` ](https://www.archlinux.org/packages/community/x86_64/libratbag/) package comes along with `ratbagd`, on Debian those seem to be two packaged separately (https://tracker.debian.org/pkg/ratbagd https://tracker.debian.org/pkg/libratbag).

whot:
`ratbagd` used to be a separate package until we realised it's not ideal to do that and we merged it into libratbag itself. I expect debian will retire the separate package soon.

Attach the output of `sudo ./build/ratbag-command -vvv info /dev/input/event19` please, with the correct event node of course.

davidovitch:
```
$ sudo ratbagctl -vvv info event4
event4 - Logitech M705
               SVG: /usr/share/libratbag/gnome/fallback.svg
      Capabilities: query-configuration
 Number of Buttons: 3
    Number of Leds: 0
Number of Profiles: 1
Profile 0: (active)
  Name : 'Profile 1'
  Resolutions:
    0: <disabled>
  Button: 0 type unknown is mapped to none
[...]

PR #548, opened 2018-09-26
tools: handle NULL svgs in ratbag-command

whot:
Showed up in #526
https://github.com/libratbag/libratbag/issues/526#issuecomment-420016451


whot:
I completely forgot about the fallback... but either way, I don't think it's the right thing to use here anyway. This is the tool we use to directly access the device, it's not shipped, so returning an SVG that's not the valid SVG is counterproductive. Different for ratbagd, where returning something useful is still better than not returning anything at all.

thx for the review, will merge this.

PR #555, opened 2018-11-12
Logitech G Pro Support

mike391:
Add support for the Logitech G Pro.  I based this off of the Logitech G203 configuration as these are essentially the same devices Driver, LED and Form factor wise. However the Logitech G Pro supports higher DPI/ improved sensor (up to 12,000 DPI) which is reflected in the new config.

I've been using this as my daily driver for about a week on Ubuntu 18.04 and haven't noticed anything out of the ordinary.

### DIFF
--- a/data/default/logitech-g-pro.svg
+++ b/data/default/logitech-g-pro.svg
@@ -1,0 +1,410 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<!-- Created with Inkscape (http://www.inkscape.org/) -->
+
+<svg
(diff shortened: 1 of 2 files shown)


whot:
I had to revert 2371960. The bit to add it to the meson.build file was missing so the test never ran on it, otherwise it should've failed immediately. AFAICT the `.device` file shouldn't even be parsed correctly by libratbag, please have a look at 3ffdc8daa07455deb11b1f8d72cc005e410463a1 to see the issues I found (there may be others).

PR #560, opened 2018-11-26
circle CI: switch fedora latest to fedora 28

whot:
Some change in F29 breaks the CI runs, so lets switch back to what we had
before so we can merge patches.


whot:
used my superpowers to merge this, so we can have test runs succeed in the other PRs

PR #561, opened 2018-11-26
circle CI: switch to ubuntu 18.10

whot:


bentiss:
FWIW, I just reverted this as 18.10 doesn't seem happy with the Circle CI setup we have.

PR #563, opened 2018-11-26
ratbagd: make Commit() truly asynchronous

whot:
Schedule the actual operation as a callback to handle at the end of the event
loop and return success. The Resync signal will handle the rest.

### DIFF
--- a/ratbagd/ratbagd-device.c
+++ b/ratbagd/ratbagd-device.c
@@ -256,20 +256,25 @@ static int ratbagd_device_get_profiles(sd_bus *bus,
 	return 0;
 }
 
+static void ratbagd_device_commit_pending(void *data)
+{
+	struct ratbagd_device *device = data;
+	int r;
+
+	r = ratbag_device_commit(device->lib_device);
+	if (r < 0)
+		ratbagd_device_resync(device, device->ctx->bus);
+}
+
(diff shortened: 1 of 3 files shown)


whot:
This should fix https://github.com/libratbag/piper/issues/177, I think

Hjdskes:
This does not fix the crash described in https://github.com/libratbag/piper/issues/177. Simply increasing the timeout for the async call does, but I'm not sure if this is just treating the symptom or the cause.

How did you think this would fix the issue? Is the goal here to bypass the need to use DBus' async calls and instead, adding the asynchronisity (?) to ratbagd itself, thereby bypassing the timeout altogether?

Hjdskes:
Scratch that, it does fix the build. I forgot to `systemctl restart ratbagd` :sweat_smile:. However, my question from above still stands, do we need any async DBus stuff in the ratbagd bindings now still?

whot:
yes, it makes us completely independent of dbus timeouts by having the async bits implemented by two separate bits in ratbagd.

That was the original intent of the `Resync` signal - `commit` always succeeds except for real errors and the rest depends on the device. If it needs 2 min to sync, well, we're good. Playing with dbus timeouts is just a race to find which device takes the longest.

And because it's dbus-independent async (i.e. not the type you'd usually see in dbus bindings) nothing is needed in the ratbagd bindings.

PR #566, opened 2018-11-26
ratbagd: update all properties before sending out Resync

### LINKED ISSUE #528: ratbagd: update all properties before Resync
The `Resync` signal notifies the DBus client that `Commit` has failed and that it needs to re-sync the device state. Except that the only way it can do so is by re-reading the properties, so we need to make sure ratbagd itself does the sync, and updates all properties before the resync signal.

Ref https://github.com/libratbag/piper/issues/182

whot:
Fixes #528

I can see the signals float past on dbus-monitor, but piper doesn't seem to update itself correctly (yet). I think this is an issue with piper rather than ratbagd though.

### DIFF
--- a/ratbagd/ratbagd-button.c
+++ b/ratbagd/ratbagd-button.c
@@ -433,3 +433,15 @@ struct ratbagd_button *ratbagd_button_free(struct ratbagd_button *button)
 	return mfree(button);
 }
 
+int ratbagd_button_resync(sd_bus *bus,
+			      struct ratbagd_button *button)
+{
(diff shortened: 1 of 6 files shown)


whot:
The signal itself is just that, a signal. DBus doesn't know what it means, that's just ratbagd context. so the properties need to be updated independently (which is what this commit does). If you hook up dbus-monitor then you can see the updated properties float past.

Anyway, more to your question: add a `return RATBAG_ERROR_DEVICE` in `ratbag_device_commit` to make sure `commit` fails, this way you trigger the resync and the property updates, etc.

Issue #558, opened 2018-11-15
Request: MX Anywhere 2 support


hexchain:
I have tried to copy over the MX Master device file and changed device IDs:

```
[Device]
Name=Logitech MX Anywhere 2
DeviceMatch=bluetooth:046d:b013
Driver=hidpp20
Svg=logitech-mx_master.svg
```

But `ratbagd` keeps crashing after an error:
```
ratbag error:     HID++ error from the device (255): ERR_INVALID_ARGUMENT (02)
```

```
       Message: Process 27792 (ratbagd) of user 0 dumped core.
                
                Stack trace of thread 27792:
                #0  0x000055cd4d00fd75 hidpp20drv_read_button_1b04 (ratbagd)
                #1  0x000055cd4d00d574 ratbag_test_driver (ratbagd)
                #2  0x000055cd4cff64e8 ratbagd_run_enumerate (ratbagd)
                #3  0x00007f731ef8e223 __libc_start_main (libc.so.6)
                #4  0x000055cd4cff66de _start (ratbagd)
```
```
(gdb) bt
#0  hidpp20drv_read_button_1b04 (button=0x55cd4ea27150) at ../src/driver-hidpp20.c:92
#1  hidpp20drv_read_button (button=0x55cd4ea27150) at ../src/driver-hidpp20.c:249
#2  hidpp20drv_read_profile (profile=<optimized out>) at ../src/driver-hidpp20.c:1006
#3  hidpp20drv_init_device (drv_data=<optimized out>, device=0x55cd4ea20e20) at ../src/driver-hidpp20.c:1295
#4  hidpp20drv_probe (device=<optimized out>, device=<optimized out>) at ../src/driver-hidpp20.c:1348
#5  0x000055cd4d00d574 in ratbag_test_driver (test_device=0x0, dev_id=<optimized out>, driver_name=0x55cd4ea31e90 "hidpp20", device=<optimized out>) at ../src/libratbag.c:335
#6  ratbag_assign_driver (test_device=0x0, dev_id=<optimized out>, device=<optimized out>) at ../src/libratbag.c:412
#7  ratbag_device_new_from_udev_device (device_out=<synthetic pointer>, udev_device=<optimized out>, ratbag=<optimized out>) at ../src/libratbag.c:484
#8  ratbagd_process_device (ctx=0x55cd4ea0c0b0, udevice=<optimized out>) at ../ratbagd/ratbagd.c:248
[...]

PR #568, opened 2018-11-27
Support MX Anywhere 2

hexchain:
Tested: device info, get/set DPI, remap button

Possible problem: "report rate" always says 0Hz - maybe mouse does not support this feature?

Still needed: an SVG of the device

Pictures of the device (found at https://support.logitech.com/en_us/product/mx-anywhere2):

Pressing the wheel does not seem to generate any event, it just switches between free/click mode, probably purely mechanically.

The square button in the middle is a "Gesture Button" by definition, but it is actually middle button by default, so I just set the button type to "middle button". The problematic button (https://github.com/libratbag/libratbag/issues/558#issuecomment-439732655) does not seem to be present on the device, although it has the `task_id` of "Gesture Button".

Some information about the device is inferred from XML files of a LogiOptions windows app installation. They are at `<somewhere>/LogiOptions/Software/Current` (I don't quite remember the leading path, sorry).

Please tell me if anything is wrong.


hexchain:
Rebased to remove an unrelated commit.

hexchain:
I've tried my best to draw the picture of the mouse, however my artistic skill can only go so far. It would be better if someone more talented can take a look at the SVG and possibly apply some fix to it.

hexchain:
Changing report rate in Piper actually works, and can be read verified by `ratbagctl info eventXX`. I have no idea why it reported 0Hz at first, though...

Hjdskes:
> I've tried my best to draw the picture of the mouse, however my artistic skill can only go so far. It would be better if someone more talented can take a look at the SVG and possibly apply some fix to it.

I think it looks great!

hexchain:
If you look closely at the left part (near the extra buttons) you'll find something wrong, but I don't know how to fix that... :-)

Hjdskes:
Can you show this on a screenshot? I don't have this device so it's hard testing.

hexchain:
> If you look closely at the left part (near the extra buttons) you'll find something wrong, but I don't know how to fix that... :-)
I mean, that part on SVG does not look really similar to the photo.

Hjdskes:
Oh, I see what you mean now. I think this is a minor issue only, as the image still easily maps to the real device so it serves its need well. It shouldn't have to hold back the PR ;)

hexchain:
I did some more testing, mainly about the report rate.

Start `ratbagd` after connecting the device:
```
$ ratbagctl rate get event21
0
$ ratbagctl rate get-all event21
500
$ ratbagctl rate set 800 event21
$ ratbagctl rate get event21
800
$ ratbagctl rate get-all event21
500
```
It seems that device report rate is changed inside `ratbagd` even if the device does not support it (i.e. no `HIDPP_PAGE_ADJUSTABLE_REPORT_RATE`). Should it be better to also have a ratbag capability for this HID++ feature?

whot:
I fixed the conflicts in the web UI but I'd rather have you rebase the branch so we can merge it nicely. The only conflict is in meson.build and that's simple enough to fix (reshuffling of the file names). Ping me when you've rebased and I'll click the magic merge button (or someone else can if they beat me to it)

hexchain:
Thank you @whot! I've rebased against current master.

whot:
I've squashed the commits for the .device and .svg files together with the meson build fix and pushed it directly as the series leading up to abe99fec6adaa224ec8d14e21fc7e3bf6b4c1991. This way we don't have half-working commits in the history.

This was faster than requesting another rebase, etc. sorry.

Closing this commit as the content is merged now, if not the branch as-is. Thanks!

hexchain:
Thank you!

PR #576, opened 2018-12-12
meson: conditional check to avoid use of depricated features after Meson v0.48.0

akstrfn:
When building the package in Archlinux I got the following:
```
DEPRECATION: Project targetting '>= 0.40.0' but tried to use feature deprecated since '0.48.0': python3 module
<snip>
WARNING: Deprecated features used:
 * 0.48.0: {'python3 module'}
```

``import(python3)`` was [depreciated in 0.48](https://github.com/mesonbuild/meson/blob/master/docs/markdown/Release-notes-for-0.48.0.md#python3-module-is-deprecated).

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -533,8 +533,13 @@ configure_file(input : 'dbus/org.freedesktop.ratbag1.conf.in',
 
 #### tools ####
 
-python = import('python3')
-py3 = python.find_python()
+if meson.version().version_compare('<0.48.0')
+  python = import('python3')
+  py3 = python.find_python()
+else
+  pymod = import('python')
+  py3 = pymod.find_installation()
+endif
 
 # ratbagctl calls ratbagd over DBus
 custom_target('ratbagctl',


meson.build
```diff
@@ -2,7 +2,7 @@ project('libratbag', 'c', 'cpp',
 	version : '0.9.903',
 	license : 'MIT/Expat',
 	default_options : [ 'c_std=gnu99', 'warning_level=2' ],
-	meson_version : '>= 0.40.0')
+	meson_version : '>= 0.48.0')
```
whot: that's going to break a few things. meson 0.48 was only released 2 months ago and [isn't shipped in Ubuntu](https://packages.ubuntu.com/search?keywords=meson). given how many issues we already have with Ubuntu I'd rather not add to them...

meson.build
```diff
@@ -533,8 +533,8 @@ configure_file(input : 'dbus/org.freedesktop.ratbag1.conf.in',
 
 #### tools ####
 
-python = import('python3')
-py3 = python.find_python()
+pymod = import('python')
+py3 = pymod.find_installation()
```
whot: you could escape this with a version check of meson itself and then make it dependent on 0.48 to do the right thing

circle.yml
```diff
@@ -66,7 +66,9 @@ fedora_prep_cache: &fedora_prep_cache
   steps:
     - run:
         name: Initializing Fedora dnf cache
-        command: dnf install -y --downloadonly libsolv tree git gcc gcc-c++ meson check-devel libudev-devel libevdev-devel valgrind python3-gobject python3-evdev glib2-devel python3-lxml
+        command: |
+          dnf install -y --downloadonly libsolv tree git gcc gcc-c++ python3 check-devel libudev-devel libevdev-devel valgrind python3-gobject python3-evdev glib2-devel python3-lxml ninja-build
+          pip3 install meson
```
whot: nope, not a fan of this. I hate having to use pip on distributions, especially those reasonably up-to-date because it'll also force all our users to run pip to build libratbag. We have enough issues with the python code as it is, we don't need to make this harder.

This would also prevent distributions shipping libratbag (pip install is a no-go in fedora's build system) which again is something I don't like because it makes getting and installing libratbag harder.

meson.build
akstrfn: I haven't done this intentionally but since you prefer that solution I'll change the PR

whot:
@akstrfn can you update the commit message please? The code is fine now but the commit message doesn't describe what the patch does anymore.

oh, and it's "deprecated", not "depreciated" (i remember it took me about 5 years to figure that out...)

FFY00:
@whot as this is just one commit you can use the squash option when merging and provide your own message. If there were multiple commits thay would be squashed together but as this is just one commit that doesn't really happen.

akstrfn:
Great tips, thanks :)
I'll update the commit tomorrow morning if you don't squash it.

whot:
@FFY00 go for it, I'm knee-deep in debugging something else atm

PR #578, opened 2018-12-13
Add Logitech MX Master 2S support

### LINKED ISSUE #419: MX Master 2S support
What do you need to add support for Logitech MX Master 2S?
Currently ratbag reports some info but Piper doesn't recognize that mouse.
I would like to contribute time/testing/code(?) to make this happen, just point me to the right direction. :)
```
$ sudo ratbag-command info /dev/input/event17
Device 'Logitech MX Master 2S'
Capabilities: res btn-key led
Number of buttons: 8
Number of leds: 2
Profiles supported: 1
  Profile 0 (enabled) (active)
    Resolutions:
      0: 1000dpi @ 0Hz (active)
    Button: 0 type left is mapped to 'button 1'
    Button: 1 type right is mapped to 'button 2'
    Button: 2 type middle is mapped to 'button 3'
    Button: 3 type side (backward) is mapped to 'button 4'
    Button: 4 type extra (forward) is mapped to 'button 5'
    Button: 5 type unknown is mapped to 'none'
    Button: 6 type wheel ratchet mode switch is mapped to 'ratchet mode switch'
    Button: 7 type unknown is mapped to 'none'
    LED: 0 type: logo, mode: off
    LED: 1 type: logo, mode: off
```

lluchs:
Works both via bluetooth and with the included universal receiver. The device files for the MX Master had a second USB id, so not sure whether I've missed anything here.

I traced the SVG from Logitech's official top-down image. The side buttons are not visible from that view, so I added an approximation to the image.

The thumb button appears as "Unknown". This is related to #124 

Fixes #419


lluchs:
The SVG will also work for the Logitech MX Master which currently doesn't have a gnome-style image. You might want to symlink the file.

codedcosmos:
Thanks so much!

PR #580, opened 2018-12-18
tools: sync ratbagd.py from Piper - drop async commit

whot:
ratbagd now implements commit as normal always-successful call with a
potential Resync signal later if something failed on the device.

PR just so I can have an extra pair of eyes on the other changes here, thanks!

### DIFF
--- a/tools/ratbagc.py.in
+++ b/tools/ratbagc.py.in
@@ -310,23 +310,15 @@ class RatbagdDevice(metaclass=MetaRatbag):
         """
         return open(self.get_svg(theme))
 
-    def commit(self, callback=None):
+    def commit(self):
         """Commits all changes made to the device.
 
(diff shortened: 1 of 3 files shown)


Hjdskes:
As in Piper, this should probably also now listen to the `Resync` signal to notify the user of any failure. Do you want to add that to this PR right away?

whot:
> As in Piper, this should probably also now listen to the Resync signal to notify the user of any failure. Do you want to add that to this PR right away?

Nah, don't have the time for this atm. It's a bit tricky now anyway, the `Resync` signal works nice for piper but not so much for ratbagctl which is supposed to exit immediately. We'd have to put a forced wait in for a few seconds just in case it *may* fail but that makes ratbagctl less useful.

Issue #579, opened 2018-12-13
G900 used instead of G Pro Wireless due to same DeviceMatch id


akstrfn:
G900 and G Pro have the same DeviceMatch=usb:046d:c539 and even though the repo now has G Pro Wireless support, svg file of G900 is used instead of G Pro Wireless. When the mouse is connected with a wire, it is properly recognized due to different DeviceMatch.

Some details of the problem can be found in PR #571 but since that PR is closed I opened this issue to further track the problem.

whot:
Just found #380 which explains the issue and what is required to fix it.

akstrfn:
I guess the best is to wait for the kernel to implement it. You would not accept some workarounds like #381?

whot:
I'd prefer the energy to be spent on fixing it in the correct place :)

akstrfn:
Makes sense :smile:

davidovitch:
For the record, should be fixed in combination with Linux 5.2: a04f2d1c24ab390213c313ace0e274a3db9922ef. See also https://www.phoronix.com/scan.php?page=news_item&px=Better-Logitech-Linux-5.2

akstrfn:
In the next weeks when 5.2 lands in my distro I'll do the tests.

FFY00:
The wireless PID (WPID) still needs to be added to the device file.

aTosser:
The WPID for G Pro Wireless is 4079 I have already made this change on my fork of both piper and libratbag and will submit PR when linux 5.2 is released.

FFY00:
@aTosser feel free to submit right now if you want :grin:.

aTosser:
I would prefer to wait until the WPID of the G900 is known. The USB receiver still shows up on Kernel 5.2. After updating all of the files for the G Pro W, I was getting duplicate entries because of the receive PID being in the G900 files. I don't want to remove the USB receiver from the G900 files but I also do not want duplicate devices showing up for anyone using G Pro W. When someone can tell me the WPID fro G900 We can just hit it all in one swoop.

bentiss:
AFAICT: https://github.com/libratbag/libratbag/commit/a04f2d1c24ab390213c313ace0e274a3db9922ef#diff-7bac8358e9b0ad1f5a1fea627b59d0d4 already added the WPID of my G900 and G403 :)

aTosser:
[...]

PR #581, opened 2018-12-18
data: add a test for duplicate DeviceMatch lines

whot:
This MR will fail the tests because we currently have two duplicates:

- [ ] logitech-g700s-wireless.device and logitech-g700-wireless.device
- [ ] logitech-g900-wireless.device and logitech-g-pro-wireless.device (#579)


whot:
I really don't know why this shows "all checks have passed" when they failed in circle ci...

whot:
@FFY00 this check would be useful to have, we can drop the duplicates now, can't we? The g-pro is the lightspeed device now. The G700s seems like more of a worry though.

FFY00:
Could you rebase this to `origin/master`? Should be fixed by now.

whot:
The g700 and g700s still use the same match on the receiver, that needs to be fixed frist.

FFY00:
Oh, then I don't think we are in any luck.

https://hansdegoede.livejournal.com/21731.html

I have one receiver but no mouse to connect it to. Someone needs to buy one of the old wireless mice (G700, G700s, G600, ... possibly others but I am not sure).

whot:
well, our current situation is broken anyway because we'll always pick whichever comes first, so one of g700 or g700s already loses out anyway. The only sensible choice here is to pick the lowest common denominator (G700 iirc) and remove the other one - same effect but we're at least not pretending otherwise :)

And whoever has a G700s will just have to live with a semi-broken device and handle it with a local data file.

FFY00:
But we don't support local data files. They would have to edit the system one.

whot:
yeah, but... eh? it's not like anything really changes from the current situation, right now you have to delete the other file to make sure you get the right one.

FFY00:
Still a better user experience IMHO, but if you want to do that go ahead, it doesn't matter that much.

whot:
it's not documented anywhere and I only know about it because of this failing test here. So I'm not sure that classifies as user experience (well, not positive one anyway). And honestly, I don't even know which one of the two it picks because it depends on the order of data files loaded and whether we append or prepend those into a list (or hashtable?). So right now, it's effectively unpredictable anyway. Anyway, I also punted this to @bentiss yesterday because he has one of those devices. So what I'll do for this PR is drop one, merge the duplicate check, then @bentiss can add the other one again once the kernel support is there.

whot:
admin superpowers! merged.

Issue #587, opened 2018-12-19
Support MX Anywhere 2S


whot:
Tested on 1d6995be724609e387be4bae9fe14c9de9eddb2d
```
[Device]
Name=Logitech MX Anywhere 2S
DeviceMatch=usb:046d:406a
Driver=hidpp20
Svg=logitech-mx-anywhere2.svg
```
Note: re-using the SVG from #568 

```
sudo ./build/ratbagctl.devel info event22        
event22 - Logitech MX Anywhere 2S
               SVG: logitech-mx-master-2s.svg
      Capabilities: query-configuration, resolution, switchable-resolution, button, button-keys
 Number of Buttons: 8
    Number of Leds: 0
Number of Profiles: 1
Profile 0: (active)
  Name : 'Profile 1'
  Resolutions:
    0: 1000dpi @ 0Hz (active)
  Button: 0 type left is mapped to 'button 1'
  Button: 1 type right is mapped to 'button 2'
  Button: 2 type unknown is mapped to 'button 3'
  Button: 3 type side is mapped to 'button 4'
  Button: 4 type extra is mapped to 'button 5'
  Button: 5 type wheel-left is mapped to 'wheel-left'
  Button: 6 type wheel-right is mapped to 'wheel-right'
  Button: 7 type unknown is mapped to none
```

Tested:
- [x] `dpi set 800`, `dpi get` shows 800dpi
- [ ] `rate set 1000` succeeds, but `rate get` shows 0
- [x] `profile 0 name get` shows "Profile 1"
  - `profile 0 name set 'foo'` throws an `RatbagErrorCapability`, see #583
- [x]  `button 0 action set button 2` succeeds but `button 0 get` shows 'button 1' (button sends `BTN_RIGHT`), see #588 
- [x] macros are not supported by this device

evilphish:
Hi, I also own this mouse and would like to see support for it in order to set the dpi. So far I have only connected the mouse via the built-in bluetooth connection of my laptop and not using the supplied extra dongle. The dongle I actually need for a different laptop anyway which has no bluetooth.

The big question is, can libratbag actually work with bluetooth mice or is USB the only way to interface with the devices at the moment? Is there anything I can contribute or test in order to help support this mouse and/or bluetooth devices?

yochananmarqos:
Why was this closed?

PR #588, opened 2018-12-19
hidpp20: if the button is remapped, use that

whot:
A non-zero value for the remapped field means the button is mapped to that
button.

"diverted" means to decouple it from normal HID reports and report it via
HID++ only. In our case here, we "map" it, which is different.

Fixes `button set`/`button get` mismatch listed in #587

### DIFF
--- a/src/driver-hidpp20.c
+++ b/src/driver-hidpp20.c
@@ -90,7 +90,7 @@ hidpp20drv_read_button_1b04(struct ratbag_button *button)
 
 	control = &drv_data->controls[button->index];
 	mapping = control->control_id;
-	if (control->reporting.divert || control->reporting.persist)
+	if (control->reporting.remapped)
 		mapping = control->reporting.remapped;
 	log_raw(device->ratbag,
 		  " - button%d: %s (%02x) %s%s:%d\n",


src/driver-hidpp20.c
```diff
@@ -90,7 +90,7 @@ hidpp20drv_read_button_1b04(struct ratbag_button *button)
 
 	control = &drv_data->controls[button->index];
 	mapping = control->control_id;
-	if (control->reporting.divert || control->reporting.persist)
+	if (control->reporting.remapped)
```
bentiss: I have strictly no ideas if this is correct or not.
The original code looked for the HW state while the new seems to be looking at the internal libratbag state. So definitively, I can't provide a good reason to nack/ack this.

whot: I don't think that's correct. This field is  [filled in by the hardware](https://github.com/libratbag/libratbag/blob/master/src/hidpp20.c#L1032), I noticed it in the debug output.

bentiss: OK, rereading the doc now, and looks like you are correct. That is some complex remapping to map in ones head.

PR #591, opened 2018-12-21
hidpp20: use macros for key+modifiers

bentiss:
Since a0830a8aaaa5168f4a2bb1c0d7594de50fcb9fb0, we can re-add the simple keybindings capabilities. It's time the HID++ 2.0 devices get this.

This used to be working with the KEY API, but this has been dropped in favor of the macro one.

This effectively replaces #402


bentiss:
Sorry for taking so long to look into this. I had to deal with macros for properly fixing #380, and I realized the code was now very simple to add.

whot:
merci

PR #592, opened 2018-12-23
tools: canonicalize path first in udev_device_from_path

jgottula:
This fixes a problem in which running `ratbag-command` with the path to a device node symlink (e.g. `/dev/logi-g502 -> /dev/input/event4`), such as one created by a udev rule, would not work correctly, due to the code failing to resolve the symlink prior to checking for the presence of the `"/dev/input/event"` prefix.

This resulted in nonsensical error messages (shown below) and the curious inability to run `ratbag-command` successfully with anything but the actual raw path to the device node itself (e.g. `/dev/input/event4` would work fine, but `/dev/logi-g502` would not, despite both being valid ways of getting to the same device node).

    Error: Can't open '/dev/logi-g502': Success
    Error: Device '/dev/logi-g502' is not supported

Note that this same problem also prevented the usage of autogenerated symlinks by udev, i.e. `/dev/input/by-id/` and `/dev/input/by-path/`.

### DIFF
--- a/tools/shared.c
+++ b/tools/shared.c
@@ -28,7 +28,7 @@ udev_device_from_path(struct udev *udev, const char *path)
 {
 	struct udev_device *udev_device;
 	const char *event_node_prefix = "/dev/input/event";
-	_cleanup_(freep) const char *path_canonical = NULL;
+	_cleanup_(freep) char *path_canonical = NULL;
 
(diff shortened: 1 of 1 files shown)


jgottula:
Change looks fine to me.

Let me know if you want me to force-push that change into one "squashed" commit with the original PR commit or something like that.

Hjdskes:
Fantastic, thanks :)

PR #505, opened 2018-08-13
data: add device file and svg for Logitech G603

damex:
wireless receiver is the same as on g403 and g900.

bluetooth seem to be occasionally working with my dw1820a wifi+bluetooth(4.1) adapter and detected as g603.

tested on ubuntu 18.04.1.

does not work over bluetooth with my intel 9260 wifi+bluetooth(5.0) returning error:

`ratbag error: G603: error opening hidraw node (Connection timed out)`

works perfectly over usb wireless adapter.

just configuration files. should be safe to merge.

```
[G603]# info
Device F5:51:9C:7A:17:8E (random)
	Name: G603
	Alias: G603
	Appearance: 0x03c2
	Icon: input-mouse
	Paired: yes
	Trusted: yes
	Blocked: no
	Connected: yes
	LegacyPairing: no
	UUID: Generic Access Profile    (00001800-0000-1000-8000-00805f9b34fb)
	UUID: Generic Attribute Profile (00001801-0000-1000-8000-00805f9b34fb)
	UUID: Device Information        (0000180a-0000-1000-8000-00805f9b34fb)
	UUID: Battery Service           (0000180f-0000-1000-8000-00805f9b34fb)
	UUID: Human Interface Device    (00001812-0000-1000-8000-00805f9b34fb)
	UUID: Vendor specific           (00010000-0000-1000-8000-011f2000046d)
	Modalias: usb:v046DpB01Cd0028
```

```
event18:   G603
```
```
event18 - G603
               SVG: /usr/share/libratbag/gnome/logitech-g603.svg
      Capabilities: query-configuration, resolution, switchable-resolution, profile, disable-profile, button, button-keys, led
 Number of Buttons: 6
    Number of Leds: 2
Number of Profiles: 1
Profile 0: (active)
  Name : 'Profile 1'
  Resolutions:
    0: 400dpi @ 1000Hz
    1: 800dpi @ 1000Hz (active) (default)
    2: 1600dpi @ 1000Hz
    3: 3200dpi @ 1000Hz
    4: <disabled>
  Button: 0 type unknown is mapped to 'button 1'
  Button: 1 type unknown is mapped to 'button 2'
  Button: 2 type unknown is mapped to 'button 3'
  Button: 3 type unknown is mapped to 'button 4'
  Button: 4 type unknown is mapped to 'button 5'
  Button: 5 type unknown is mapped to 'resolution-cycle-up'
  LED: 0 type: logo, depth: rgb, mode: off
[...]

PR #600, opened 2019-01-08
data: added Logitech G603 svg

zacjor:
Added a proper SVG for the Logitech G603 as requested in https://github.com/libratbag/libratbag/pull/505

Sorry if this was submitted improperly, I'm still pretty new to using git.

### DIFF
--- a/data/gnome/logitech-g603.svg
+++ b/data/gnome/logitech-g603.svg
@@ -14,7 +14,7 @@
    viewBox="0 0 409.2084 411.27099"
    version="1.1"
    id="svg8"
-   inkscape:version="0.92.2 2405546, 2018-03-11"
+   inkscape:version="0.92.3 (2405546, 2018-03-11)"
    sodipodi:docname="logitech-g603.svg"
    inkscape:export-filename="/home/jimmac/logitech-g403.png"
    inkscape:export-xdpi="96"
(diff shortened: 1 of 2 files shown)


zacjor:
Looks like there are still a few changes I need to make to the SVG. For one thing, the LED under the middle button is not configurable so I'll add it to the device layer. Second, I messed up the left and right click and center button. I should have those fixed in a moment.

Hjdskes:
> Edit: Correction, it is the LED/Logo that I'm having issues with, not the LED under the middle button. On the G603 it's just a printed logo and not an LED. Should I just remove the logo from the SVG completely in this case?

When I look at other (Logitech) devices, I see that the [G300](https://github.com/libratbag/libratbag/blob/master/data/gnome/logitech-g300.svg) and [G9](https://github.com/libratbag/libratbag/blob/master/data/gnome/logitech-g9.svg) also show the "logo" while this isn't an LED.

On the other hand, the [G500](https://github.com/libratbag/libratbag/blob/master/data/gnome/logitech-g500.svg), [G600](https://github.com/libratbag/libratbag/blob/master/data/gnome/logitech-g600.svg) and [G700](https://github.com/libratbag/libratbag/blob/master/data/gnome/logitech-g700.svg) have the logo on the device but not in the SVG.

@whot We should probably add something to the guidelines for this and make the SVGs consistent. Do you have a preference for one over the other?

whot:
IMO it doesn't matter as long as we're reasonably consistent. The logo adds a bit of 'depth' to it, makes it look ever so slightly more real. but it's far from required.

And given that most users only have one mouse, it's fine if we're a tad inconsistent right now.

zacjor:
@Hjdskes @whot I went ahead and removed the logo for the time being as I was having some issues with Inkscape. Otherwise, the SVG should be good to go.

PR #603, opened 2019-01-09
data/gnome/logitech-g-pro-wireless: don't use Logitech LED logo

Hjdskes:
We don't want this for copyright reasons.


whot:
bedankt

PR #610, opened 2019-01-29
data: add device file for Logitech G603

FFY00:
Cherrypicked device file from #505 

Closes #505


Hjdskes:
Can you add it to the meson.build list as well?

FFY00:
Sorry about that, totally forgot.

PR #623, opened 2019-02-18
DBus: drop the Svg and GetSvg() APIs

whot:
This is prep work for libratbag 1.0.

These APIs are now legacy, we assume that we're running sandboxed because that's the future. So let's drop them before we commit to maintain these forever.


Hjdskes:
If you fix the conflict we can merge this.

whot:
> Will you sync the ratbagd change to Piper as well? :)

yep, will do

> If you fix the conflict we can merge this.

sir, yessir. tomorrow, sir.

PR #624, opened 2019-02-19
Coverity fixes

whot:


whot:
FTR coverity is happy with this now

PR #626, opened 2019-02-19
ratbagd: map the default theme to the gnome theme

whot:
We don't install the default SVGs anymore and all the effort for new mice goes
into the gnome theme.


whot:
IMO having that one string there doesn't really cost us anything, dropping it means we basically make it 'impossible' to add other sets of devices.

PR #630, opened 2019-02-20
gskill: don't break after finding the active index

whot:
We still need to read the data of all other profiles. This broke in f26de4bb,
the previous loop only did the is-active modification

cc @Lyude


whot:
Assuming this is correct, if not we'll revert it :)

PR #631, opened 2019-02-20
Drop device capabilities

whot:
FTR, first patch is from #630 

This patchset drops the device capabilities. They are mostly superfluous because we should simply assume that if we have multiple profiles, we can switch between them and write to them. Same for buttons, etc.

Some of the button caps are now handled by the action types list. And the profile disable/default caps are better handled in the profile (because that's where the set_foo() calls are).

This is of course an API break, the Capabilities property on the device disappears.

Another subtle change: if our name isn't writable, simply don't export it (DBus empty string).


Hjdskes:
Oke, that took less effort than expected. One change I had to make in the ratbagd bindings was the following, due to https://github.com/libratbag/libratbag/pull/631/commits/f285b12f212cd00cb461969696fe2c3bd166375f#diff-4edf8a4ccbbc4924299aa67ad39e8283R388:

```diff
@@ -485,7 +455,11 @@ class RatbagdProfile(_RatbagdDBus):
     @GObject.Property
     def name(self):
         """The name of the profile"""
-        return self._get_dbus_property("Name")
+        name = self._get_dbus_property("Name")
+        if not name:
+            name = "Profile {}".format(RatbagdProfile.counter)
+            RatbagdProfile.counter += 1
+        return name
```

Perhaps we can include this in the linked diff instead, because `ratbagd_profile` already has the `index` as a member of its struct.

PR #634, opened 2019-02-20
Switch to using hidraw as base event source

whot:


bentiss:
I am fine with the change. However, I think I'll need to test this further with the actual devices before acking and merging this.

I think we need to make sure:
- wired USB devices are still working (Logitech, and one other)
- wireless USB still working (through lightspeed, so G900)
- bluetooth device still working (MX Master)

Also, the commit description could be slightly enhanced: please say that this will break ratbagctl but people are not supposed to hardcode "eventX"in their configuration file...

whot:
> please say that this will break ratbagctl but people are not supposed to hardcode "eventX"in their configuration file...

fwiw, I'm quite tempted to scramble the ID here anyway and just use something that cannot be directly mapped. it's annoying because it hurts us, but reduces the potential of misuse.

whot:
Updated commit message force-pushed.

FTR, tested on a g303, mx anywhere 2s and g500s, 2 wired, one LUR.

PR #636, opened 2019-02-20
Add a logo!

whot:
Because how can we have survived for that long without one?! Oh, right, we're a system daemon. Anyway, who cares. Looks is all that matters these days.


bentiss:
OK, but I wonder if we can not use SVG effects to make it color-breath like most gaming mice should do?

Hjdskes:
Where'd we use the logo? In the README and the Wiki? I'm missing those bits from this PR. I do like the logo, don't get me wrong :)

> OK, but I wonder if we can not use SVG effects to make it color-breath like most gaming mice should do?

Or we can color breathe the rats in different colors for extra discobombulation (pun intended).

whot:
> Where'd we use the logo? In the README and the Wiki?

wait, we're not getting tshirts made?

bentiss:
While trying to add the color effects (because we can), I realized that the rats are not properly sent to their layer. Fix coming in.

Hjdskes:
> wait, we're not getting tshirts made?

Do maintainers get one for free? :D

bentiss:
OK, done with the animation: https://github.com/bentiss/libratbag/tree/wip/logo
and the result: https://github.com/bentiss/libratbag/blob/wip/logo/data/logo.svg

bentiss:
too bad, I can't make github markdown show the animations when embedding the logo on the README.

But it seems that we can do it when the image is on github.io. So I added a new commit that adds this logo on the dbus api doc, and then we can pick this one from there :)

I'll open a new PR, might be just easier.

PR #637, opened 2019-02-21
Scramble user visible ids for ratbagctl

whot:
Right now:

```
$ ratbagctl list
event3: Logitech MX Anywhere 2S         
event5:   Logitech Gaming Mouse G303      
event6:   Logitech G500s Laser Gaming Mouse
```
This is bad because it encourages users to hardcode the event node in their scripts.  And we don't actually care about the event node anyway. With the patch set from #634:

```
hidraw0: Logitech MX Anywhere 2S         
hidraw2:   Logitech Gaming Mouse G303      
hidraw4:   Logitech G500s Laser Gaming Mouse
````

Which is better but still encourages users to hardcode the event node in their scripts. With this glorious patchset:

```
chinchilla: Logitech MX Anywhere 2S         
hamster:   Logitech Gaming Mouse G303      
woodrat:   Logitech G500s Laser Gaming Mouse
```
Boom! No more hardcoding because it should be fairly obvious that none of those refer to a specfic syspath. Plus it may encourage people to learn a bit more about rodents, because hands up if you knew that a tuco-tuco existed.

No-one? 

See, we like to be educational here too.

FTR: the name assigned is actually static while the hidraw node doesn't change but we could shuffle the array of names.

`ratbag-command` still uses the device nodes, Piper doesn't display the ID anyway.

### DIFF
--- a/ratbagd/ratbagd-device.c
+++ b/ratbagd/ratbagd-device.c
@@ -44,6 +44,7 @@ struct ratbagd_device {
 	struct ratbagd *ctx;
 	RBNode node;
 	char *sysname;
+	char *id;
 	char *path;
 	struct ratbag_device *lib_device;
 
@@ -274,7 +275,7 @@ ratbagd_device_get_capabilities(sd_bus *bus,
 
 const sd_bus_vtable ratbagd_device_vtable[] = {
 	SD_BUS_VTABLE_START(0),
-	SD_BUS_PROPERTY("Id", "s", NULL, offsetof(struct ratbagd_device, sysname), SD_BUS_VTABLE_PROPERTY_CONST),
+	SD_BUS_PROPERTY("Id", "s", NULL, offsetof(struct ratbagd_device, id), SD_BUS_VTABLE_PROPERTY_CONST),
 	SD_BUS_PROPERTY("Capabilities", "au", ratbagd_device_get_capabilities, 0, SD_BUS_VTABLE_PROPERTY_CONST),
 	SD_BUS_PROPERTY("Name", "s", ratbagd_device_get_device_name, 0, SD_BUS_VTABLE_PROPERTY_CONST),
 	SD_BUS_PROPERTY("Profiles", "ao", ratbagd_device_get_profiles, 0, SD_BUS_VTABLE_PROPERTY_CONST),
@@ -287,6 +288,7 @@ const sd_bus_vtable ratbagd_device_vtable[] = {
 int ratbagd_device_new(struct ratbagd_device **out,
 		       struct ratbagd *ctx,
 		       const char *sysname,
+		       const char *id,
 		       struct ratbag_device *lib_device)
 {
 	_cleanup_(ratbagd_device_freep) struct ratbagd_device *device = NULL;
@@ -297,6 +299,7 @@ int ratbagd_device_new(struct ratbagd_device **out,
 	assert(out);
 	assert(ctx);
 	assert(sysname);
+	assert(id);
 
 	device = zalloc(sizeof(*device));
 	device->ctx = ctx;
@@ -304,6 +307,7 @@ int ratbagd_device_new(struct ratbagd_device **out,
 	device->lib_device = ratbag_device_ref(lib_device);
 
 	device->sysname = strdup_safe(sysname);
+	device->id = strdup_safe(id);
 
 	r = sd_bus_path_encode(RATBAGD_OBJ_ROOT "/device",
 			       device->sysname,
@@ -357,10 +361,17 @@ struct ratbagd_device *ratbagd_device_free(struct ratbagd_device *device)
 	device->lib_device = ratbag_device_unref(device->lib_device);
 	device->path = mfree(device->path);
 	device->sysname = mfree(device->sysname);
+	device->id = mfree(device->id);
 
 	assert(!device->lib_device); /* ratbag yields !NULL if still pinned */
 
 	return mfree(device);
+}
+
+const char *ratbagd_device_get_id(struct ratbagd_device *device)
+{
+	assert(device);
+	return device->sysname;
 }
 
 const char *ratbagd_device_get_sysname(struct ratbagd_device *device)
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: ratbagd/ratbagd.c [hunk 1]
```diff
@@ -204,6 +204,67 @@ static const sd_bus_vtable ratbagd_vtable[] = {
 	SD_BUS_VTABLE_END,
 };
 
+/* 32 pre-picked names. If you have more than 32 devices connected and need
+ * ratbag to control them all, congratulations, you are official a niche
+ * use-case.
+ */
```
bentiss: in that case, we should add a google translate API and convert the rodents names to Swedish or Icelandic :)

ratbagd/ratbagd.c
```diff
@@ -204,6 +204,67 @@ static const sd_bus_vtable ratbagd_vtable[] = {
...
+	 */
+	if (sscanf(sysname, "%*[^0-9]%d", &index) == 1 &&
+	    index >= 0 && (unsigned)index < sizeof(device_ids_free) * 8 &&
+	    !!(device_ids_free & (1 << index)))
+	    goto assign;
+
+	index = ffs(device_ids_free);
+
+	/* This can only happen if you have >32 devices, so let's not worry
+	 * about this too much */
```
bentiss: On the T450s:
```
/dev/input/event0:	Lid Switch
/dev/input/event1:	Sleep Button
/dev/input/event2:	Power Button
/dev/input/event3:	AT Translated Set 2 keyboard
/dev/input/event4:	Logitech M325
/dev/input/event5:	Integrated Camera: Integrated C
/dev/input/event6:	Video Bus
/dev/input/event7:	PC Speaker
/dev/input/event8:	ThinkPad Extra Buttons
/dev/input/event9:	HDA Intel HDMI HDMI/DP,pcm=3
/dev/input/event10:	HDA Intel HDMI HDMI/DP,pcm=7
/dev/input/event11:	HDA Intel HDMI HDMI/DP,pcm=8
/dev/input/event12:	HDA Intel HDMI HDMI/DP,pcm=9
/dev/input/event13:	HDA Intel HDMI HDMI/DP,pcm=10
/dev/input/event14:	HDA Intel PCH Dock Mic
/dev/input/event15:	HDA Intel PCH Mic
/dev/input/event16:	HDA Intel PCH Dock Headphone
/dev/input/event17:	HDA Intel PCH Headphone
/dev/input/event18:	Synaptics TM3053-004
/dev/input/event19:	TPPS/2 IBM TrackPoint
/dev/input/event20:	Logitech K750
```

Just one external keyboard+mice and I am already at 21 devices. This still gives me 10 devices left, but I wonder if we should not just cycle over the list instead of returning here.

bentiss: one solution could be what jitsi or Ubuntu is doing: prepend an adjective.

whot:
> Does this replace #634 ?

No, it doesn't. That PR will just have less user-visible effect but in the end we should still use it because to get 32 hidraw devices, you really need 32 devices (-ish) whereas `event32` is quite achievable as you noticed.

ratbagd/ratbagd.c
whot: urgh. UTF-8 is already a pain without å and the like :)

ratbagd/ratbagd-device.c
```diff
@@ -274,7 +275,7 @@ ratbagd_device_get_capabilities(sd_bus *bus,
 
 const sd_bus_vtable ratbagd_device_vtable[] = {
 	SD_BUS_VTABLE_START(0),
-	SD_BUS_PROPERTY("Id", "s", NULL, offsetof(struct ratbagd_device, sysname), SD_BUS_VTABLE_PROPERTY_CONST),
+	SD_BUS_PROPERTY("Id", "s", NULL, offsetof(struct ratbagd_device, id), SD_BUS_VTABLE_PROPERTY_CONST),
```
Hjdskes: We should document this in the DBus API.

ratbagd/ratbagd.c
```diff
@@ -204,35 +204,98 @@ static const sd_bus_vtable ratbagd_vtable[] = {
...
+	 * Totally not necessary but we might as well have some fun here.
+	 * Where this fails or picks an id already chosen, we fall back to
+	 * "choose the first one free".
+	 */
+	if (sscanf(sysname, "%*[^0-9]%d", &index) == 1 &&
+	    index >= 0 && (unsigned)index < sizeof(device_ids_free) * 8 &&
+	    !!(device_ids_free & (1 << index)))
+	    goto assign;
+
+	index = ffs(device_ids_free);
```
Hjdskes: Never knew there was a *for fuck's sake* function in the standard library, interesting!

bentiss:
```
$> man ffs

FFS(3)    Linux Programmer's Manual    FFS(3)

NAME
       ffs, ffsl, ffsll - find first bit set in a word

SYNOPSIS
       #include <strings.h>

       int ffs(int i);

       #include <string.h>

       int ffsl(long int i);

       int ffsll(long long int i);

...
```

standard C function, cherry picked by Python :)

whot: i can never unsee the real meaning of ffs either :)

ratbagd/ratbagd-device.c
whot: already there, see `doc/dbus.rst`

```
          A unique ID describing this device. This ID should not be used for presentation to
          the user. This ID may be recycled when the device is removed. The
          content of the ID is undefined, the client should treat it as an
          opaque string.
```

ratbagd/ratbagd.c
whot: resolving based on #634 which reduces the number of nodes we have

ratbagd/ratbagd-device.c
Hjdskes: Oops, missed that. Sorry.

whot:
> I wonder if it is really necessary?

Yes. Google for the glibc `memmove` debacle a decade or so ago - and memmove had the warning in for decades before that. I'd rather not have ppl shout at me in 2 years time because they hardcoded the event node and their magic script is now in the ubuntu forums forever.

ratbagd/ratbagd.c
Hjdskes: *Those darn chinchillas!*

Hjdskes:
@whot Fair enough, consider this an ACK from me.

PR #638, opened 2019-02-21
Remove the SVG handling from libratbag

### LINKED ISSUE #628: Move the SVGs from libratbag to piper
Right now, ratbagd provides the SVGs to client. The SVGs are themed, with the theory that different clients can load differently-looking SVGs. Realistically we only have one theme: "gnome", used by piper. No-one has shown interest in other themes, including the original few SVGs we had before piper.

I'm proposing removing the themes from libratbag and pushing them into piper.
- there is nothing about the API that specifies what the SVG does, beyond "it's an SVG". This is problematic, because:
- the theme is a "piper" theme more so than a GNOME theme. piper has requirements on how the theme must behave which only exist because of piper's implementation. Button label names, leader lines, etc.
- if we are to make this a stable API, this behaviour of the SVGs becomes ABI too - we could never change the theme without potentially breaking other users
  - hypothetical issue: if piper's svg library changes/goes away and the new one renders slightly differently, we cannot update the svgs because they are part of the ABI. So we're doubly screwed.
- the only reason we have the fd passing is because we expect piper to run in a container and not have access to the host. Shipping it in piper makes that problem go away.

There are some notable downsides:
- adding a new device requires the user updates libratbag and piper
- other tools that rely on these SVGs (if they exist? ratbagctl doesn't count) will stop working. But see my ABI comment above.
- it is harder to enforce SVGs being added for new device if they're across two different repositories. That's a social problem, nothing technical stop us from saying "no .device file in libratbag until an SVG is in piper".

I think the benefit of not having to care about SVGs and magic ABI issues is greater than those drawbacks, so let's drop the SVGs and live merrily ever after

cc @bentiss, @Hjdskes, @FFY00, @phomes, @cvuchener

whot:
This removes the DBus SVG api but adds a new `Device.UsbId` property that makes it easier for clients to match devices to the SVG files.

Fixes #628 

Matching piper PR: https://github.com/libratbag/piper/pull/305


bentiss:
> This removes the DBus SVG api but adds a new `Device.UsbId` property

How about we switch back to the first SVG API: libratbag export the file name, and then piper just fetches the file in its database?
I am afraid adding the `Device.UsbId` will complexify the thins for no reasons. Plus it will be future proof if we have devices not handled by a VID/PID but a dbus interface or something else.

Hjdskes:
I would also like to explore @bentiss' suggestion.

whot:
uhm, not sure that's a good API. ratbagd effectively saying "i don't know about SVGs but here's a string you can use to find one if you have one". Which means we have the `.device` files in libratbag require an entry for each device just for someone else to maybe use that entry. That's a tad weird and prone to go wrong.

> Plus it will be future proof if we have devices not handled by a VID/PID but a dbus interface or

maybe so, but let's be honest, virtually every device out there has a vid/pid. and if we use dbus as backend, the actual *device* is still usb or bluetooth and we'll likely have access to the vid/pid and can forward it. worst case we can introduce a "dbus" bus string.

bentiss:
Maybe we should not call it `SVG` in libratbag, but having a unique representation string in the `.device` file that is forwarded as is on the wire might be easier for piper on the other side.
And we can say: this repr string could be `logitech-g900` or `roccat-kone`.

bentiss:
> maybe so, but let's be honest, virtually every device out there has a vid/pid. and if we use dbus as backend, the actual _device_ is still usb or bluetooth and we'll likely have access to the vid/pid and can forward it. worst case we can introduce a "dbus" bus string.

Problem is, piper doesn't know about that. Why would it need it? It talks to `chinchilla` over dbus, not `USB VID:PID`.

whot:
Long summary from a meeting @bentiss and I had this morning:

The base requirement is that we have some model-specific identifier that allows us to load the right SVG. *How* we do this is largely up to us. But one thing really matters: we need to separate what libratbag exports from how piper uses it.

One option is to have an opaque string like ``chinchilla`` that stays the same for our models. This provides absolutely no information to the client and forces them to have a database to map those strings into device models. libratbag would have to maintain that database too. That string becomes part of the ABI, forever. We need to make sure that the device remains correctly mapped to that string for the lifetime of libratbag.

The second option is to have a semi-opaque string like ``logitech-g5`` or the advertised model name ``Logitech G5``. This provides better information in that it is at least human-identifiable and makes things like SVG file names easy. However, similar restrictions apply, libratbag needs to maintain a database. If a new model gets released with the same name, we now have to revert to hacks like ``logitech-g5-2019``. We already have devices like the G9x and the G502 that need multiple entries, it's unclear what identifier to use. Because it still has to be treated as opaque string in the client, it carries no more information than ``chinchilla``.

The third option is to have a model identifier in the form of ``bustype:vid:pid:version``. This removes the need for a database on libratbag because we just forward the information from the device. It provides extra information (want to handle BT devices only? skip anything not starting in ``bluetooth``). Likewise if you only want to handle devices from one specific vendor. Plus, if this fails we can blame the vendor instead of us :)

Biggest drawback here is that it's not as human-friendly on the other side. So it'll still require some database to map ``usb...`` to ``logitech-foo.svg`` but given the drawbacks of the other two approaches, this is just something we have to live with.

Notable changes over the patchset originally submitted:
@bentiss and I agreed that instead of the patchset's `(bus, vid, pid)` tuple making this a defined string is more flexible. So we define ``usb:...`` and ``bluetooth:...``, in the future we may have a ``ble:...`` format that is completely different (no VID/PID?)

The ``version`` field (was not in the first patchset) means we have an escape where devices re-use VID/PID, we can just bump the version for those.

bentiss:
> The second option is to have a semi-opaque string like `logitech-g5` or the advertised model name `Logitech G5`. This provides better information in that it is at least human-identifiable and makes things like SVG file names easy. However, similar restrictions apply, libratbag needs to maintain a database. If a new model gets released with the same name, we now have to revert to hacks like `logitech-g5-2019`. We already have devices like the G9x and the G502 that need multiple entries, it's unclear what identifier to use. Because it still has to be treated as opaque string in the client, it carries no more information than `chinchilla`.

A few more arguments against this:
- this string would be written by a human, which means it will be error prone. And if we start shipping a device with a bad copy/paste from one file to one other, 2 different mice will have the same product string and we will be screwed. We could partially solve that by adding a check in meson, but that would be extra work
- the second extra issue is that it would force libratbag to take the decision on splitting the device based on an assumption on how it will be used. For example, the MX Master can be connected over LUR or bluetooth (and maybe directly from the USB cable). Either libratbag splits this device in 3 which would duplicate the files for no reasons (and will be error prone), either we keep only one device file, but then piper can't take the decision of showing a different SVG if the mouse is wired or not.

Which is why, for all of these reasons, @whot convinced me that we better have a uniquely generated ID, that ratbagd can guarantee its uniqueness and if it's not unique, then we can blame the hardware maker (or the driver).

whot:
done!

bentiss:
thanks

Issue #635, opened 2019-02-20
ratbagctl: use the device node as first argument


whot:
Having the device node last makes it easy for us to test the same line on various devices. That is not the most common use-case - normally we'd expect our users to have one mouse and run several commands on it (that's the case during debugging as well). So let's change the device node to be the first argument, it makes for easier typing/copy-paste, etc.

Plus it makes it a lot more obvious which one the full command is (`dpi set 800` in this example)
```
$ ratbagctl hidraw4 dpi set 800
```

PR #644, opened 2019-02-22
ratbagctl: put the device node at the beginning of the command

bentiss:
See #635 for a full description, or the commit message

### DIFF
--- a/tools/ratbagctl.in
+++ b/tools/ratbagctl.in
@@ -603,15 +603,8 @@ def dpi(string):
         return (int_x, int_y)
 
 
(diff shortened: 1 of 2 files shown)


bentiss:
Grmbl, and of course, we have tests to test the command line arguments, so they all fail.

bentiss:
forced push with the tests fixed

Hjdskes:
We also need to update documentation, such as the Wiki.

bentiss:
> We also need to update documentation, such as the Wiki.

Aye!

whot:
that's done now too btw

PR #651, opened 2019-02-26
LED color depth documentation

whot:
We now ignore monochrome LEDs because we probably can't do much with them anyway, so meh. We can revisit that decision if we need to.

### DIFF
--- a/doc/dbus.rst
+++ b/doc/dbus.rst
@@ -432,30 +432,62 @@ org.freedesktop.ratbag1.Led
         :type: u
         :flags: read-write, mutable
 
-        Enum describing the current mode, see
-        :cpp:enum:`ratbag_led_mode`.
+        Enum describing the current mode, see :attr:`Modes`.
 
 .. attribute:: Modes
 
         :type: au
         :flags: read-only, constant
 
-        A list of modes supported by this LED, see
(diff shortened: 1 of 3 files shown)


ratbagd/ratbagd-profile.c
```diff
@@ -528,6 +528,11 @@ int ratbagd_profile_new(struct ratbagd_profile **out,
 		if (!led)
 			continue;
 
+		/* We don't care about LEDs that we can't change the color
+		 * for */
+		if (ratbag_led_get_colordepth(led) == RATBAG_LED_COLORDEPTH_MONOCHROME)
+			continue;
```
bentiss: Nack

On the G502 non rgb, a quite popular mouse, setting the LED is important. Because the LED is blue and I remember not being able to sleep in the same room than this mouse because of the default breathing effect. I ended up unplugging the mouse IIRC, but this setting should still be presented to the user.

IMO, the Color dbus parameter should not be exposed, but the mode and brightness should

whot: removed from the patchset

whot:
Merging, because this is basically just a documentation change now anyway - and an enum value change.

PR #653, opened 2019-02-26
ratbagd: change the Resolution property to a variant

whot:
Instead of a fixed (uu) with an extra capability flag, let's just make this a
variant defined to be a u or a (uu), depending on whether we have a single or
a separate resolution.

In the ratbagd.py bindings we return a tuple for both, but a simple len()
check will tell us which one it is.

### DIFF
--- a/tools/ratbagd.py
+++ b/tools/ratbagd.py
@@ -534,16 +534,28 @@ class RatbagdResolution(_RatbagdDBus):
 
     @GObject.Property
     def resolution(self):
-        """The tuple (xres, yres) with each resolution in DPI."""
-        return self._get_dbus_property("Resolution")
+        """The resolution in DPI, either as single value tuple ``(res, )``
+        or as tuple ``(xres, yres)``.
+        """
+        res = self._get_dbus_property("Resolution")
+        if isinstance(res, int):
+            res = tuple([res])
+        return res
 
     @resolution.setter
-    def resolution(self, res):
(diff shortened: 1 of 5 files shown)


tools/ratbagd.py
```diff
@@ -534,16 +534,28 @@ def capabilities(self):
...
-    def resolution(self, res):
+    def resolution(self, resolution):
         """Set the x- and y-resolution using the given (xres, yres) tuple.
 
         @param res The new resolution, as (int, int)
         """
-        self._set_dbus_property("Resolution", "(uu)", res)
+        res = self.resolution
+        if len(res) != len(resolution) or len(res) > 2:
+            raise ValueError('invalid resolution precision')
```
Hjdskes: Wasn't there a problem with raising exceptions in GObject getters and setters? Or was it only the getters? I don't recall.

whot: yes, there is. I had a bug yesterday that I only found after changing this to a normal `@property`. Not sure how we can work around this though...

Hjdskes: We can ignore these invalid values with or without printing a warning and document these requirements as part of the API/documentation.

whot: yep, can you file a separate issue for this please so we don't lose track of it?

Hjdskes: Done: #662.

PR #661, opened 2019-02-28
updated logo look and feel

andyfitz:
Rough update but headed in the right direction

* Simplified silhouette of bag
* Reduced number of rats inside the Rat Bag to two 
* Eyes are still small and potentially still not printable at small scale
* Introduced rough stylized logotype based on Google's Sedgwick Ave Display font
* Reduced total file size by a heap


whot:
Thanks, I like it! just to verify: the license would be MIT by default because that's what libratbag is shipped under. Happy with that?

andyfitz:
Ack,  I'm happy with that.
License is MIT same as libratbag, you as the rights owner. 
Consider that burden my lame gift to you 😝

PR #663, opened 2019-02-28
meson: require >= 0.46.0

Hjdskes:
When building, I got

```
 * 0.46.0: {'Python Module'}
```

So let's bump the required version for meson.


Hjdskes:
Hmm... Ubuntu 17.10 doesn't have a new enough version for this.

whot:
this is a bug in meson, we have the offending call escaped behind meson version checks but it still picks it up with a complaint.

Hjdskes:
Good to know, thanks

PR #669, opened 2019-03-05
Install the device files in the top-level datadir

whot:
Now that the SVGs are gone, we don't need the subfolder anymore


whot:
well, the `.device` files are technically implementation defined anyway, so if we need to, we'll move them. But right now, I can't think of anything else we'd add there.

PR #671, opened 2019-03-05
gitignore: add build directories

FFY00:
Just a really minor enhancement.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,6 @@
 *.rej
 *.trs
 tags
+build/
+builddir/
+test-build/


Hjdskes:
I think stuff like this should go into your personal, global .gitignore.

FFY00:
I partly agree but I don't see any reason why we can't have them here too. Why do we even have a `.gitignore`? None of the entries are actually project specific, this is just to make it easier for the people who want to submit some patches and are not that familiar with git.

bentiss:
> Why do we even have a `.gitignore`?

IIRC one compelling argument for meson and else was to *not* ship a `.gitignore` at all. This let the users to just have their own...

whot:
indeed, we could remove the gitignore and not suffer any side-effects. This is a leftover from autotools which sprinkles stuff all over the source directory in the default invocation.

As a general rule, you shouldn't add directories to gitignore because it really blanks out that whole directory and you don't usually notice until you've pushed a commit that breaks on some other users' computer :)

The build directory is user-specific btw, it's an argument passed in by the user at meason time. Most use `build` but the docs refer to `builddir`, I use a few more for different build combinations. Adding user-specific stuff to gitignore is generally not a good idea either.

PR #674, opened 2019-03-07
Map disabling a button to the action type NONE

whot:
We already have `ACTION_TYPE_NONE` which is specified as disabling a button, there's no need for a separate DBus interface here. 

WIP, simply because I haven't tested it yet and there may be a corner case missing.


FFY00:
Closing in favor of #953

Issue #421, opened 2017-11-09
Adding device Roccat Kone XTD Optical / Roccat Kone EMP


SebastianPfliegel:
Hi there,

I quickly became a fan of the Roccat hardware as it was fully supported on Linux (libgaminggear). Unfortunately the developer has quit adding new hardware and therefore my new Roccat Kone EMP is unsupported. I'd love to support old and upcoming Roccat hardware on Linux. That's why I'm here!

I also call the Roccat Kone XTD Optical my own. I tried my luck to add support for this device (as it's quite similar to the Roccat Kone XTD) and failed.

This is my device file:
```
[Device]
Name=Roccat Kone XTD Optical
DeviceMatch=usb:1e7d:2e23
Driver=roccat
Svg=roccat-kone-xtd.svg
```

libratbagd show this error:
> ratbag error: libratbag bug: ROCCAT ROCCAT Kone XTD Optical: invalid dpi list

Where do you think I should take a look first?

whot:
libgaminggear is GPL which is why is one reason we haven't connected to it directly. and it's API is not the most straightforward to use for us.

We already have the roccat driver though so I guess trying to use that as a starting point is the best approach.  that error shouldn't happen because you don't have a DPI list. looks like something isn't handling an empty value correctly - this bit should be easy to debug though.

SebastianPfliegel:
Thanks for the quick reply! Yeah, I've seen issue #12.

I'll try to debug it on my own. Will let you know if I need help with it.

SebastianPfliegel:
Took some time to teach my IDE (VS Code) to run GDB as root. Is there actual a way to use the device configs from the local development directory while debugging instead of /usr/share/libratbag/?

These lines cause the message:

https://github.com/libratbag/libratbag/blob/109cc69da9a1920414bf267ecf2d5e2749344fd1/src/libratbag.c#L279-L283

https://github.com/libratbag/libratbag/blob/109cc69da9a1920414bf267ecf2d5e2749344fd1/src/libratbag.c#L1171-L1172

whot:
we have `ratbagd.devel` and `ratbagctl.devel` that use the files from the build tree (and create custom dbus interfaces so they don't interfere with a running ratbagd). And there's `ratbag-command` that skips the dbus bits and interacts with the library directly. See also the wiki here https://github.com/libratbag/libratbag/wiki/Developers, not that there's a lot of information there...

[...]

PR #682, opened 2019-03-24
Support for Roccat Kone Pure Optical Black/White Edition

hipnoizz:
WARNING: Do not merge, it will break support for Roccat Kone XTD.

I'm working on adding support for Roccat Kone Pure, specifically Optical Black/White Edition. Before I proceed further I would like some advice in a few areas.

Generally I wanted to take this approach:
1. Change `roccat` driver so it can handle both Kone XTD and Kone Pure (is there anyone that can validate my changes using XTD?).
2. Migrate the driver to new structure with `commit()` function.
3. Add support for additional features already supported by ratbag (e.g. LEDs, as I have a working prototype already).

The draft pull request visualizes changes that I needed to make to handle Kone Pure mouse. On one hand it borks existing support for Kone XTD. On the other hand it shows how both these mice differ.

1. My first question is how to tell the driver which mouse model he needs to communicate with. I saw two examples in the existing codebase - the `steelseries` driver uses `DeviceVersion` property (1, 2, 3...) and the `hidpp10` driver uses `ProfileType` property (e.g. G700). I'm not sure why there are already have two different properties already. Unfortunately I would prefer a property named `DeviceModel` as `ProfileType` is confusing due to the primary meaning of profiles as a set of configuration values. Thoughts?

2. Some differences between Pure and XTD could be described in the device files, e.g. `DpiRange` (100:5000@50 vs 200:8200@50), maybe number of buttons as well.

3. More serious are remaining differences, namely a) different data structures (USB HID GET/SET_REPORT) and button mappings. One way to handle that is to introduce additional structures (e.g. rename `roccat_settings_report` to `roccat_settings_report_kone_xtd` and then add a slightly different `roccat_settings_report_kone_pure`) and use conditions (if-oriented-programming ;-)) to handle different devices - there are like around 10 places like that. I would probably introduce several helper functions as some existing functions like `roccat_read_profile()` are already pretty lengthy. Any better ideas? Note that adding support for more Roccat devices would probably require to push further in that direction, as even XTD Optical seems to differ from 'common' XTD (https://github.com/libratbag/libratbag/issues/421).

I started documenting Kone Pure protocol here: https://github.com/hipnoizz/libratbag/wiki/Roccat-Kone-Pure if someone is interested. Some of this found it's place in the codebase, but not everything, so I think it is a worthwhile addition anyway.

### DIFF
--- a/data/devices/roccat-kone-pure.device
+++ b/data/devices/roccat-kone-pure.device
@@ -1,0 +1,4 @@
+[Device]
+Name=Roccat Kone Pure
+DeviceMatch=usb:1e7d:2dc2
+Driver=roccat
--- a/meson.build
+++ b/meson.build
@@ -308,6 +308,7 @@ data_files = files(
 	'data/devices/logitech-MX-Anywhere2.device',
 	'data/devices/logitech-MX-Anywhere2S.device',
 	'data/devices/roccat-kone-xtd.device',
+	'data/devices/roccat-kone-pure.device',
 	'data/devices/steelseries-kinzu-v2.device',
 	'data/devices/steelseries-rival-310.device',
 	'data/devices/steelseries-rival-600.device',
--- a/src/driver-roccat.c
+++ b/src/driver-roccat.c
@@ -34,7 +34,7 @@
 #include "libratbag-hidraw.h"
 
 #define ROCCAT_PROFILE_MAX			4
-#define ROCCAT_BUTTON_MAX			23
+#define ROCCAT_BUTTON_MAX			17
 #define ROCCAT_NUM_DPI				5
 #define ROCCAT_LED_MAX				0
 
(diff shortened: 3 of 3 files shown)


whot:
> is there anyone that can validate my changes using XTD

I have an XTD, ping me if you need specific testing. I'm also severely ETIME all the time though...

> how to tell the driver which mouse model he needs to communicate with.

Exactly like you describe. Those properties are purely implementation-defined and we can change them at will, whenever we feel like it. Pick what you feel comfortable with. `DeviceModel` can be iffy if you have multiple models sharing the same firmware, but as I said, we can change it anytime anyway.

> Some differences between Pure and XTD could be described in the device files

Yep, that would make sense

>  I would probably introduce several helper functions [...]

What we did for HID++ was to write protocol-only files (`hidpp10.c` and `hidpp20.c`) and then a separate driver that uses that protocol (`driver-hidpp10.c` and `driver-hidpp20.c`). Since not all devices support all bits of the protocol, this makes the driver easier to read. I think this approach should work for roccat devices too.

bentiss:
In addition to what @whot said, there is no need to over engineer the device file and try to make it really generic.
We do have generic Logitech device files, because we know well their protocols and we can write one properly. For device we feel uncomfortable having a common driver, there is nothing wrong adding a new one. For example, the Logitech G300 and G600 are similar but completely separated. This avoids having to make device comparisons and keep the drivers simpler.

So one solution could be to rename `driver-roccat.c` into `driver-roccat-kone-xtd.c` and have your as `driver-roccat-kone-pure.c`. This way you could have your driver quicker upstream and collect all of the fame and bug reports from various users :)

And just to add more water here, the drivers and the device files are pure internal component, we do not care about ABI and we can rename whatever we want. So feel free to use the simpler solution for you, and refine it later when time permit.

stephanlachnit:
@whot @FFY00 Since this is bascially merged and there's is no development here I can be closed I guess. But maybe @hipnoizz's wiki link should be noted somewhere? Could be either in the wiki or in the driver itself as a comment.

FFY00:
I think we can come back here if needed.

PR #685, opened 2019-04-05
test: disable tests relying on assert if NDEBUG is defined

whot:
This happens when building a 'release' build with meson or manually specifying
-Db_ndebug=true.

This just fixes the symptom, we rely too much on assert to always work. Which
isn't good.

### DIFF
--- a/test/test-context.c
+++ b/test/test-context.c
@@ -144,12 +144,14 @@ test_context_suite(bool using_valgrind)
 
 	s = suite_create("context");
 	tc = tcase_create("init");
+#ifndef NDEBUG
 	if (!using_valgrind) {
 		tcase_add_test_raise_signal(tc, context_init_NULL, SIGABRT);
 		tcase_add_test_raise_signal(tc, context_init_bad_iface, SIGABRT);
 		tcase_add_test_raise_signal(tc, context_init_bad_iface_open, SIGABRT);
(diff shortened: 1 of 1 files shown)


bentiss:
I don't really understand.

Looking at the downstream bug: https://bugzilla.redhat.com/show_bug.cgi?id=1696264 it seems that a meson change disabled the asserts?

How can this be even possible?

I am not happy adding such paper over the fact that there is a meson bug and we can't check if our program works as expected.

bentiss:
Looking at the assert man page:

> If the macro NDEBUG is defined at the moment <assert.h> was last included, the macro assert() generates no code, and hence does nothing at all.  It is not recommended to define NDEBUG  if  using assert() to detect error conditions since the software may behave non-deterministically.

We are in a case where we use assert to detect an error condition, and I do not see why meson would force us to change our program.

whot:
There are two parts to the story here. One is the changed default in meson in Fedora, now reverted. The other one is that someone may build libratbag with extra flags like `-DNDEBUG`, thus breaking things. The question is whether we want to handle this case, and if so, just block it? A three-line snippet would be enough here:
```
#ifndef NDEBUG
#error Do not disable NDEBUG
#endif 
```

bentiss:
I'd go for the snippet above, yes

whot:
Closing in favour of #700

PR #693, opened 2019-04-19
added notice about dependencies in README

Simon-Laux:


Hjdskes:
We purposely do not list dependencies in the README, because these lists tend to get stale and outdated. If needed, you can find a list in our CI configuration: https://github.com/libratbag/libratbag/blob/master/circle.yml

Issue #692, opened 2019-04-15
Low poll/refresh-rates not avaliable


Riesi:
I know this sounds like a silly issue. I mean who wants to use a lower poll-rate? But here is the point. Wine has [a problem with high mouse poll-rates](https://bugs.winehq.org/show_bug.cgi?id=33643) in multiple titles and the only fix for now is to set the poll-rate to 125Hz. Which is not possible with libratbag at least on the mice a friend of mine and I own.

Logitech G203 Prodigy
Logitech G700
Both mice support poll-rates of 125, 250, 500 and 1000Hz on Windows, but the lower two (125 & 250) are not available in libratbag+piper.

I am not sure if its possible to enable the low poll-rates for every mouse. But it would be great to maybe support setting the supported refresh-rates in the device config?

thanks in advance,
Riesi

mpipo:
I just got a G203 today and while ratbagctl seems to be letting me "change" the rate, ```evhz``` still reports 1000Hz and ```/sys/kernel/debug/usb/devices``` ```Ivl=1ms```

I can't even get it down to 500Hz. It's a shame Wine has this problem in 2019.

soredake:
@FFY00 can you please add 125 and 250 poll-rate to rival 310?

FFY00:
They are there:
```
$ ratbagctl hollering-marmot rate get-all
125 250 500 1000
```

We are, however, not printing the correct values when using the `get` command.
```
$ ratbagctl hollering-marmot rate set 125
$ ratbagctl hollering-marmot rate get
100
$ ratbagctl hollering-marmot rate set 250
$ ratbagctl hollering-marmot rate get
200
$ ratbagctl hollering-marmot rate set 500
$ ratbagctl hollering-marmot rate get
500
$ ratbagctl hollering-marmot rate set 1000
$ ratbagctl hollering-marmot rate get
1000
```

Can you reproduce this? Which version of libratbag are you using?

If you can't reproduce and you are not using an outdated version please open a new issue.

soredake:
I see, but from piper i see only 500/1000 rates, maybe report to the piper repo?

whot:
piper is hardcoded to only show the 500/1000 rate, that was a conscious decision on the basis that was a combination of "meh, do we need more?" and "it looks nicer this way" :)

FFY00:
But with wireless mice being more widespread it might make sense to change this behavior as lower polling rates have lower power consumption.

whot:
[...]

PR #697, opened 2019-04-24
G502 fixes (more likely HID++ 2.0)

bentiss:
I pulled out my G502 and found a few issues today.

I think it might solve #692 but I am not entirely sure if this is everything that needs to be done.

Anyway, feedback welcome :)

### DIFF
--- a/src/driver-hidpp20.c
+++ b/src/driver-hidpp20.c
@@ -1254,9 +1254,14 @@ hidpp20drv_commit(struct ratbag_device *device)
 			return RATBAG_ERROR_DEVICE;
 
 		list_for_each(profile, &device->profiles, link) {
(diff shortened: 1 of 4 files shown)


mpipo:
Tested on my G203 and can confirm it does indeed change the polling rate :smile: You, sir, are awesome.

There are two issues remaining (which probably aren't related to this PR) though

- The polling rate can't be changed from Piper, it has to be done through ```ratbagctl```

- ```ratbagctl G203 rate get-all``` reports ```125 250 500 1000```, but really maps to ```100 200 500 1000``` when set. Tested with ```evhz```.

whot:
@mpipo best to file separate issues for those two so we can address them now that this one is in

@bentiss thanks!

PR #701, opened 2019-05-07
Update circleci to ubuntu 19.04

whot:


whot:
yeah, it's bloody annoying. 17.10 seems to have dead repos now, 18.04 has the same timeout issue as this one. Needs to be analysed except that I need to find the time for it first.

whot:
Ok, found the issue, see the latest commit. `software-properties-common` installs `libnss-systemd` which is used by `dbus-daemon` on startup. Since systemd isn't available everything times out slowly (multiple minutes). And that timeout is also triggered after we install the `ratbagd.devel`-specific dbus configuration file. DBus doesn't respond during that timeout, so our tests fail waiting for that.

Simply removing `libnss-systemd` is sufficient to get this working again.

PR #703, opened 2019-05-16
Add battery voltage reporting for newer mice

pedrovanzella:
Newer mice don't report their battery levels. Instead, they report the voltage read by the SOC. The lghub app does some math with it to figure out the percentage, using some tables shipped with it, but for our purposes I believe having the battery voltage (in millivolts) and the status (charging, wireless charging, charged, discharging...) should suffice.

This is related to [this piper issue](https://github.com/libratbag/piper/issues/222) and the discussion there.

### DIFF
--- a/src/driver-hidpp20.c
+++ b/src/driver-hidpp20.c
@@ -56,6 +56,7 @@
 #define HIDPP_CAP_ONBOARD_PROFILES_8100			(1 << 6)
 #define HIDPP_CAP_LED_SW_CONTROL_1300			(1 << 7)
 #define HIDPP_CAP_ADJUSTIBLE_REPORT_RATE_8060		(1 << 8)
+#define HIDPP_CAP_BATTERY_VOLTAGE_1001			(1 << 9)
 
 struct hidpp20drv_data {
 	struct hidpp20_device *dev;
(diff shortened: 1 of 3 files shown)


pedrovanzella:
Should I include the battery data and math to turn the voltage into a percentage in this PR? There is also a matter of libratbag actually doing something with this information, as right now it's just setting the capabilities flag (both for this feature and the older `0x1000` feature).

FFY00:
What do you think @bentiss?

bentiss:
I think there has already been some discussion in #222.

So basically, thanks a lot for the code, but I am not sure we want to merge this in libratbag. Battery is a feature that needs to be monitored all the time, and having it in libratbag is not the place.

We should really port this code to the kernel, and thanks to @jwrdegoede we have the code merged in v5.2, so adding those tiny bits in the kernel should be easier now (at least over the lightspeed connector).

So keeping this opened, for references, but I am not sure we should merge the code now, as this would be just for documentation purposes.

bentiss:
After second thoughts, and after a chat with @whot, I decided to merge this. This will give us a way to easily test if the battery is properly handled.

PR #706, opened 2019-05-21
ci: add arch

FFY00:


whot:
I just fixed the ubuntu failure, please rebase on that. May also help you here - and it looks like you don't have a dbus-daemon running. See the ubuntu commands, we start it manually there too.

PR #707, opened 2019-05-23
meson: add 'sphinx-build' to sphinx executable name list

FFY00:
We currently assume the user is using Python 2 as default (over Python 3) and so they renamed the Sphinx Python 3 executable to `sphinx-build{3,-3}`. The default name is `sphinx-build` and is usually kept when using Python 3 as the default Python version. With Python 2 dying in 7 months (https://pythonclock.org/) and Sphinx officially dropping Python 2 (https://github.com/sphinx-doc/sphinx/commit/9412bd76b7f5fdf0adc231ef8130e45bda130164) this seems safe to merge.

### DIFF
--- a/doc/meson.build
+++ b/doc/meson.build
@@ -1,6 +1,6 @@
 if get_option('documentation')
 
-cmd_sphinx = find_program('sphinx-build-3', 'sphinx-build3')
+cmd_sphinx = find_program('sphinx-build-3', 'sphinx-build3', 'sphinx-build')
 
 doc_config = configuration_data()
 doc_config.set('PACKAGE_NAME', meson.project_name())


whot:
ubuntu 17.10 is dead it seems, so let's merge this

FFY00:
I wonder if we shouldn't remove the other file names and leave just `sphinx-build`. Sphinx officially dropped support for Python 2 so the distributions should start naming the Python 3 binary `sphinx-build` soon. I would make sense to change this more or less at same time as Sphinx.

whot:
>  so the distributions should start

Fedora 30 names it `sphinx-build` for the python3 version, but Fedora 29 still uses `sphinx-build` for the python2 one. That has around 6months left. Not sure about other distros. 

But unless all major distributions have *finished* renaming it, we'll just introduce weird errors, especially in the documentation that our users don't care about. So I'd say we leave it as-is and think about it again in a year or two.

PR #708, opened 2019-05-23
circleci: update to fedora 30

whot:


FFY00:
Looks like the `dbus-daemon` binary was renamed or it wasn't pulled as a dependency.

whot:
yup, f30 replaced the defaults with dbus-broker but I haven't figured out yet how to start this from the commandline instead of as a dbus service. Meanwhile, we can get by with the old `dbus-daemon`, good enough until that gets fully retired.

circle.yml
```diff
@@ -82,8 +82,8 @@ fedora_install: &fedora_install
     command: |
         echo keepcache=1 >> /etc/dnf/dnf.conf
         dnf upgrade -y libsolv
-        dnf install -y tree git gcc gcc-c++ meson check-devel libudev-devel libevdev-devel valgrind python3-gobject python3-evdev glib2-devel python3-lxml python3-devel swig libunistring-devel
-
+        dnf install -y tree git gcc gcc-c++ meson check-devel libudev-devel libevdev-devel valgrind python3-gobject python3-evdev glib2-devel python3-lxml python3-devel swig libunistring-devel dbus-daemon
+        sed -i 's/systemd//' /etc/nsswitch.conf
```
FFY00: Why is this?

whot: see https://github.com/libratbag/libratbag/commit/c4257802206186b891bea8d055c14b03772a64bf, https://github.com/libratbag/libratbag/pull/701#issuecomment-495420829 has a summary too. Without this, we're just failing with dbus-daemon related timeouts

PR #709, opened 2019-05-26
 steelseries: fix bug where we were wrongly dellocating dpi_{range,list} 

FFY00:
`_cleanup_(freep)` would deallocate the pointer returned by `ratbag_device_data_steelseries_get_dpi_range(device->data)` which resulted in a wrong step value. The step count sent to the device was then wrongly calculated due to this erroneous step value.

What about `dpi_list`? This is still an issue there, right?


whot:
yeah, `dpi_list` seems to suffer from the same issue here. can you please amend and push so that both are fixed? also - typo in the commit message -> `deallocated` needs two l.

PR #728, opened 2019-07-04
hidpp10: fix segfault for the c531 (and possibly other 27MHz receivers)

FFY00:
hidpp10_read_memory will return the HID++ error if the request fails.
right now, the error code from hidpp10_read_memory is transparently
passed through hidpp10_onboard_profiles_read_macro. This is problematic
because hidpp10_onboard_profiles_parse_macro only aborts on rc < 0 and
HID++ error are > 0.

---

Is this the right approach. Do we want `hidpp10_onboard_profiles_read_macro` to keep `hidpp10_read_memory`'s error code? Or should I overwrite `rc` when we see that `hidpp10_read_memory` failed?


whot:
good enough, thanks. there's bound to be a few more of those btw, I remember fixing a similar one somewhere too.

PR #731, opened 2019-07-07
data: add Logitech G Pro Hero

andyklimczak:
Running into some issues building libratbag/piper locally to test

### DIFF
--- a/data/devices/logitech-g-pro-hero.device
+++ b/data/devices/logitech-g-pro-hero.device
@@ -1,0 +1,5 @@
+[Device]
+Name=Logitech Gaming Mouse G Pro Hero
+DeviceMatch=usb:046d:c08c
+Driver=hidpp20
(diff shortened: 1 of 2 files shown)


andyklimczak:
Nevermind this PR, I see that this pid/vid is already covered in the [`.device` file](https://github.com/libratbag/libratbag/blob/master/data/devices/logitech-g-pro.device#L3) for the g pro:

>DeviceMatch=usb:046d:c085;usb:046d:c08c

andyklimczak:
And the reason my G Pro Hero isn't being recognized is because the second ID for the hero isn't in the latest release for my distro (arch) I think.

Just making sure it wasn't something wrong with libratbag

FFY00:
Yes, the last release of libratbag doesn't include the missing ID for the G Pro yet. As soon as a new version is released the arch package will be updated.

Issue #729, opened 2019-07-05
(logitech) Bad CRC when reading onboard memory profiles (8100)


FFY00:
I haven't been able to check this on older devices but at least in the newer ones (tested on G Pro Wireless, G603, MX518) we get an invalid CRC when reading the onboard memory profiles (feature 8100). Maybe the memory map changed and we are reading in the wrong place. I need some confirmation on how libratbag behaves with older devices.

aTosser:
This is definatly the case for G Pro Wireless, spend many hours in windows with wireshark and LGS. 
Beginning few lines for ratbagd trying to read profiles:
`ratbag raw: hidpp write: 10 01 09 08 00 00 00`
`ratbag raw: hidpp read:  11 01 09 08 01 03 01 05 01 08 10 00 ff 0a 04 00 00 00 00 00`
`ratbag raw: hidpp write: 10 01 09 28 00 00 00`
`ratbag raw: hidpp read:  11 01 09 28 01 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00`

Same lines of LGS reading the profiles:
`Leftover Capture Data: 1001094e000000`
`Leftover Capture Data: 1101094e00010000000000000000000000000000`
`Leftover Capture Data: 1101095e00010000000000000000000000000000`
`Leftover Capture Data: 1101095e01000064009600c800fa002c01ffffff`

LGS is reading my DPI correctly from the profile.. Therefore we are reading the wrong place. At least for newer devices.

ratbagd full init:
http://dpaste.com/2K1NR5W
LGS-profile full read:
http://dpaste.com/27YFJZM

Full capture of device being plugged in and LGS querying it:
[Initial Start.pcapng.gz](https://github.com/libratbag/libratbag/files/3364234/Initial.Start.pcapng.gz)

Please let me know if there is anything I can do to help. My C is very rusty but I can take a whack at it if someone would be willing to review some of the codebase with me.

FFY00:
I just confirmed this doesn't happen due to different memory mappings. The memory mapping is the same between all devices. What did however changed is the profile structure but it was done in such way that newer versions just extend the older ones, so that wouldn't cause this issue either.

Bypassing the CRC check fixes everything, so we have a faulty CRC check routine.

FFY00:
Finally found the issue :dizzy_face:. The last byte of the CRC isn't being properly read.

FFY00:
So, the newer devices report a 255 sector size, but we need to read 256 bytes.

[...]

PR #732, opened 2019-07-08
hidpp20: fix reading onboard profiles (8100)

FFY00:
when we read the memory from the device we already align the offset so
that it matches the end of the sector. for ex, for a 255 sector size the
offset will be set to 255-16.
when we copy the memory (using memcpy) we just copy the bytes that will
finish filling the buffer, ignoring that the offset was aligned. we need
to copy the full 16 bytes, otherwise it will fail when sector_size does
not align with 16.

### DIFF
--- a/src/hidpp20.c
+++ b/src/hidpp20.c
@@ -1558,7 +1558,7 @@ hidpp20_onboard_profiles_read_sector(struct hidpp20_device *device,
 			return rc;
 
 		/* msg.msg.parameters is guaranteed to have a size >= 16 */
-		memcpy(data + offset, buf.msg.parameters, min(count, 16));
+		memcpy(data + offset, buf.msg.parameters, 16);
 
 		/*
 		 * no need to check for count >= 0:


FFY00:
Fixes #729.

whot:
punting to @bentiss because he wrote that bit and the comments are fairly specific about how we may get invalid arguments here.

FFY00:
Sure. But basically the 8100 feature doesn't allow us to read past the end of the sector, so we position the offset so that it doesn't go past it (we read 16 bytes at a time so we position it at `sector_size - 16`).

We are then are are using `min(count, 16)` to calculate the size which would be valid if we didn't align the offset. But as we align the offset we need to copy the full 16 bytes to the memory.

Example (to also use as documentation):
Given a `sector_size` of 64
```
Read 16 bytes at 0x00 (total 16 bytes read)
Read 16 bytes at 0x10 (total 32 bytes read)
Read 16 bytes at 0x20 (total 48 bytes read)
Read 16 bytes at 0x30 (total 64 bytes read)
```

Given a `sector_size` of 70
```
Read 16 bytes at 0x00 (total 16 bytes read)
Read 16 bytes at 0x10 (total 32 bytes read)
Read 16 bytes at 0x20 (total 48 bytes read)
Read 16 bytes at 0x30 (total 64 bytes read)
Read 16 bytes at 0x36 (total 70 bytes read)
```

```0x36 = 0x46 (70 dec) - 0x10 (26 dec)```

PR #733, opened 2019-07-08
hidpp20: fall back to firmware profiles on a profile by profile basis

FFY00:
currently if any use profile is not valid we fallback to the firmware
profiles. this is quite annoying because the user might have only one
valid profile. this patch will keep valid profiles and only fallback to
firmware profiles on a profile by profile basis.

### DIFF
--- a/src/hidpp20.c
+++ b/src/hidpp20.c
@@ -1464,6 +1464,14 @@ int hidpp20_adjustable_report_rate_get_report_rate_list(struct hidpp20_device *d
 #define HIDPP20_ONBOARD_PROFILES_PROFILE_TYPE_G900	0x03
 #define HIDPP20_ONBOARD_PROFILES_MACRO_TYPE_G402	0x01
 
+#define HIDPP20_USER_PROFILES_G402			0x0000
+#define HIDPP20_ROM_PROFILES_G402			0x0100
+
+#define HIDPP20_PROFILE_DIR_END1			0
+#define HIDPP20_PROFILE_DIR_END2			1
+#define HIDPP20_PROFILE_DIR_END_VALUE			0xFF
+#define HIDPP20_PROFILE_DIR_ENABLED			2
+
 union hidpp20_internal_profile {
 	uint8_t data[HIDPP20_PROFILE_SIZE];
 	struct {
@@ -2384,16 +2392,17 @@ hidpp20_onboard_profiles_parse_profile(struct hidpp20_device *device,
 	return 0;
 }
 
-static int
-hidpp20_onboard_profiles_parse(struct hidpp20_device *device,
-			      struct hidpp20_profiles *profiles,
-			      uint16_t dict_address,
-			      bool check_crc)
+int
+hidpp20_onboard_profiles_initialize(struct hidpp20_device *device,
+				    struct hidpp20_profiles *profiles)
 {
 	_cleanup_free_ uint8_t *data = NULL;
 	int rc;
 	unsigned i;
 	bool crc_valid;
+	bool read_userdata = true;
+
+	assert(profiles);
 
 	for (i = 0; i < profiles->num_profiles; i++) {
 		profiles->profiles[i].address = 0;
@@ -2403,7 +2412,7 @@ hidpp20_onboard_profiles_parse(struct hidpp20_device *device,
 	data = hidpp20_onboard_profiles_allocate_sector(profiles);
 
 	rc = hidpp20_onboard_profiles_read_sector(device,
-						  dict_address,
+						  HIDPP20_USER_PROFILES_G402,
 						  profiles->sector_size,
 						  data);
 	if (rc < 0)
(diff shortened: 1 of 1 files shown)


src/hidpp20.c
```diff
@@ -2458,6 +2458,17 @@ hidpp20_onboard_profiles_parse(struct hidpp20_device *device,
...
 							    check_crc);
+
+		/* on fail to read the user profile fallback to the firmware profile */
+		if(rc && profiles->profiles[i].address == i + 1)
+		{
+			profiles->profiles[i].address += 0x0100;
+			rc = hidpp20_onboard_profiles_parse_profile(device,
+							    profiles,
+							    i,
+							    check_crc);
```
bentiss: Don't  you need to call check_crc with false here? On the mice I have, the ROM profiles do not have a valid CRC, so this call would fail.

src/hidpp20.c
```diff
@@ -2458,6 +2458,17 @@ hidpp20_onboard_profiles_parse(struct hidpp20_device *device,
...
+		}
+
 		if (rc == -EAGAIN)
 			crc_valid = false;
```
bentiss: This makes me think we should rethink the calls instead of just adding a band aid.

src/hidpp20.c
FFY00: TBH I didn't really understand what you were doing in the loop above. `hidpp20_onboard_profiles_read_sector` is only called one time and we are reading multiple profiles?? Doesn't each profile have its own sector?

src/hidpp20.c
FFY00: Yes, we should also probably replace `hidpp20_onboard_profiles_initialize`'s fallback mechanic for this one.

src/hidpp20.c
bentiss: The loop before the reading of the profiles is to read the dictionary of the device, where the addresses of the profiles are stored and where we know if a profile is enabled or not.

FFY00: Yeah nvm, you are just reading the profile directory, that's fine. What do you think about disabling the profile here and in the initial caller trying to enable it again with rom values?

bentiss: We always wanted to have a non intrusive approach. So if the device works, we should not tamper with it. Especially if there is a bug like the other one you just solved where libratbag consider a profile invalid when it is not and you are resetting the mouse for users.

TL;DR: please do not edit the dictionary there

FFY00: I was thinking about disabling the profile inside libratbag but as you point out the internal state in libratbag is tied to the state in the hardware. What do you propose then? How can we see if the profile as a valid CRC in the initial caller?

bentiss: basically unfold the calls to `hidpp20_onboard_profiles_parse()` in `hidpp20_onboard_profiles_initialize()`, and from there change the whole logic.

FFY00: That part I understand but how are you thinking about changing the whole logic? I have been trying to reformulate the logic in my head and I don't see how that would be substantially different from what I am doing here.

bentiss: I am not saying we should rethink the whole commands. The logic is fine.

But in the light of the new logic, the code should be reflecting that new logic only:
For each profile, we try to fetch the in-flash profile, if it fails, try to read the ROM one.

Then done.

src/hidpp20.c
```diff
@@ -2412,92 +2421,67 @@ hidpp20_onboard_profiles_parse(struct hidpp20_device *device,
 	crc_valid = hidpp20_onboard_profiles_is_sector_valid(device,
```
bentiss: 1/3: If the crc is not valid (corrupted dictionary)..

src/hidpp20.c
```diff
@@ -2412,92 +2421,67 @@ hidpp20_onboard_profiles_parse(struct hidpp20_device *device,
...
+					       i + 1,
+					       profiles->profiles[i].address,
+					       HIDPP20_USER_PROFILES_G402 | (i + 1));
+
+			profiles->profiles[i].enabled = !!d[HIDPP20_PROFILE_DIR_ENABLED];
 		}
+	} else {
+		hidpp_log_debug(device, "Profile directory has an invalid CRC... Reading ROM profiles.\n");
+
+		read_userdata = false;
```
bentiss: 2/3: ... we are accessing the ROM profiles, which is good, ...

src/hidpp20.c
```diff
@@ -2412,92 +2421,67 @@ hidpp20_onboard_profiles_parse(struct hidpp20_device *device,
...
+							profiles,
+							i,
+							false);
+
 		if (rc == -EAGAIN)
 			crc_valid = false;
 	}
 
-	if (check_crc && !crc_valid)
+	if (!crc_valid)
```
bentiss: 3/3: ... but `crc` is still invalid, meaning we are returning `-EAGAIN` here

I think you should keep the valid crc only for the dictionary, but remove any assignments to it when reading the profiles.

src/hidpp20.c
```diff
@@ -2412,92 +2421,67 @@ hidpp20_onboard_profiles_parse(struct hidpp20_device *device,
...
 		rc = hidpp20_onboard_profiles_parse_profile(device,
-							    profiles,
-							    i,
-							    check_crc);
+							profiles,
+							i,
+							false);
+
 		if (rc == -EAGAIN)
 			crc_valid = false;
```
bentiss: What's the point of setting the CRC to invalid here when we know the ROM profile doesn't have a valid CRC, meaning the `rc` error we have here is likely unrelated?

src/hidpp20.c
FFY00: Why I don't understand? What do you want me to do?

src/hidpp20.c
FFY00: Same. I don't understand what you want me to do :)

src/hidpp20.c
bentiss: The three comments 1/3 2/3 and 3/3 are supposed to be considered as a whole :)

FFY00: So the new changes I pushed are fine?

src/hidpp20.c
```diff
@@ -2412,91 +2421,60 @@ hidpp20_onboard_profiles_parse(struct hidpp20_device *device,
...
+		/* the number of rom profiles can be different than the number of user profiles
+		   so we if there are not enough rom profiles to populate all the user profiles
+		   we just use the first rom profile */
+		if (i + 1 > profiles->num_rom_profiles)
+			profiles->profiles[i].address = HIDPP20_ROM_PROFILES_G402 + 1;
+		else
+			profiles->profiles[i].address = HIDPP20_ROM_PROFILES_G402 + i + 1;
 
-	hidpp_log_debug(&device->base, "ROM settings loaded\n");
+		rc = hidpp20_onboard_profiles_parse_profile(device,
```
bentiss: The value of rc is now never checked.
We might want to have a `if (rc <0) return rc;` here

FFY00:
@bentiss

src/hidpp20.c
```diff
@@ -2412,91 +2421,61 @@ hidpp20_onboard_profiles_parse(struct hidpp20_device *device,
...
-					profiles->profiles[i].address = dict_address | (i + 1);
-				}
-				break;
-			default:
-				hidpp_log_error(&device->base,
-						"this should never happen: dict address is 0x%04x\n",
-						dict_address);
-				return -EINVAL;
-			}
+			profiles->profiles[i].enabled = !!d[HIDPP20_PROFILE_DIR_ENABLED];
```
bentiss: Didn't you said that we should not try to read the disabled profiles?

src/hidpp20.c
```diff
@@ -2412,91 +2421,61 @@ hidpp20_onboard_profiles_parse(struct hidpp20_device *device,
...
+			profiles->profiles[i].address = HIDPP20_ROM_PROFILES_G402 + 1;
+		else
+			profiles->profiles[i].address = HIDPP20_ROM_PROFILES_G402 + i + 1;
 
-	hidpp_log_debug(&device->base, "ROM settings loaded\n");
+		rc = hidpp20_onboard_profiles_parse_profile(device,
+							profiles,
+							i,
+							false);
+		if (rc < 0) return rc;
```
bentiss: Maybe go to the next line before the `return rc` :)

src/hidpp20.c
FFY00: Yeah but nevermind. We should still read disabled profiles. If a user wishes to enable the profile we need to have the data there.

We can either read them here or when they are enabled, I would prefer to read them here.

FFY00:
@bentiss

bentiss:
As mentioned in my previous comment:
> OK, getting close, please fix the nitpicks and merge.

so please merge it now that the nitpicks are fixed.

PR #739, opened 2019-07-17
Use python3-embed as dependency on Python 3.8

whot:
Python 3.8 no longer links to libpython, resulting in undefined references to
Python API.

But it provides a python3-embed.pc now, so let's try using that first and only
fall back on the normal case if that fails.

See https://github.com/mesonbuild/meson/issues/5629

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -573,8 +573,14 @@ swig_gen = generator(
                 '@INPUT@'],
 )
 
+# From python 3.8 we neeed python3-embed
+dep_python3 = dependency('python3-embed', required: false)
+if not dep_python3.found()
+	dep_python3 = dependency('python3')
+endif
+
 wrapper_deps = [
-    dependency('python3'),
+    dep_python3,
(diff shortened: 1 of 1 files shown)


muzena:
Is this dependency changes result in PPA failed build?
[Log file](https://launchpadlibrarian.net/433381570/buildlog_ubuntu-bionic-amd64.libratbag_0.9.905-1-201907170001~ubuntu18.04.1_BUILDING.txt.gz)

whot:
The error there is 
```../src/hidpp20.c:2465:5: error: format not a string literal and no format arguments [-Werror=format-security]```
which is a simple printf-related fix, but unrelated.

whot:
@muzena I think @FFY00 already fixed that one in 7a14026f2dfb1b607d0d0353fcb4155bca6f511e

FFY00:
Yup, but after the release :facepalm:. #745 should prevents this from happening again.

PR #764, opened 2019-07-25
Drop the custom git sha for ratbagd.devel

whot:
After spending way too much time working around the git sha instead of getting useful stuff done, I've decided to drop it. Our `.devel` tools are supposed to be for developers but right now they get in the way more than anything - try to debug why ratbagd crashes when you're running a ratbagctl.test, it's insane.

Anyway, first bunch of patches replaces the `org.freedesktop.ratbag_devel1_$gitsha` with just `org.freedesktop.ratbag_devel1` as busname. Nothing too exciting here. Second-to-last patch moves installing the dbus file from the tool box to `ratbagd.devel` directly, this way it can just be started as-is.

Last patch is just a cosmetic fix unrelated, because we can't be pretty enough.


FFY00:
> try to debug why ratbagd crashes when you're running a ratbagctl.test, it's insane

I just run `ratbagd` directly.

meson.build
```diff
@@ -409,21 +417,35 @@ executable('ratbagd',
 install_man('ratbagd/ratbagd.8')
 
 #### ratbagd_devel ####
+#
+# A development ratbagd server that owns a different name on the bus (named
+# after the git sha). Do not use directly, this server is used by
```
FFY00: Not anymore :laughing:.

FFY00:
> try to debug why ratbagd crashes when you're running a ratbagctl.test, it's insane

I just run `ratbagd` directly.

whot:
`ratbagd` doesn't use the data files from the source repo but loads it from the prefix. And it doesn't initialize the test device, which is needed when you're testing random things. Those two are the main motivation for `ratbagctl.devel`.

meson.build
whot: fixed! thanks for the review. I also squashed in the removal of the git-sha-generating script.

PR #768, opened 2019-07-26
meson.build: fix builds with --buildtype=plain

singron:
When building with --buildtype=plain, the build fails with the following
error:

    cc1: error: ‘-Wformat-security’ ignored without ‘-Wformat’ [-Werror=format-security]

-Werror=format-security was recently added in 43def1b3a5a by #745


whot:
thanks!

PR #773, opened 2019-07-27
Add the ability to load test devices via JSON

whot:
Right now, we have a hardcoded test device with "magic" values that the tests just have to know. This makes it hard to expand the tests, because it's not obvious which value can be changed without breaking tests or making things worse.

This PR replaces that test device with a new DBus call to load a test device based on a JSON file. That 

most basic one (1 profile, 1 resolution, etc.) and adds a new devel DBus API to load a JSON file with the data to merge into the test device. This makes the test more flexible and more obvious too - it's clear why we can test for say dpi 1300.

I'm planning to squash the test changes into one commit but for easier reviewing it's separated out.

The JSON parsing doesn't have a lot of error checking and is a bit hacky, but since this is a developer-only testing interface anyway crashing or asserting isn't the worst of all things.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -66,7 +66,7 @@ fedora_prep_cache: &fedora_prep_cache
   steps:
     - run:
         name: Initializing Fedora dnf cache
-        command: dnf install -y --downloadonly libsolv tree git gcc gcc-c++ meson check-devel libudev-devel libevdev-devel valgrind python3-gobject python3-evdev glib2-devel python3-lxml libunistring-devel dbus-daemon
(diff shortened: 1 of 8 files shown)


whot:
ready for review!

bentiss:
Seems good from a rough point of view (I'll need to read it more carefully later today).

The one thing @FFY00 is working on in https://github.com/libratbag/ratbag-emu is to have the same JSON description of a test device so he can make sure to have a known state before starting the CI on the specific REST endpoint.

So I made 1+1 in my head and thought that we might be able to use the same Json description so we can test the dbus API and the ratbag basic test driver with or without `ratbag-emu` (how wonderful would that be?)

whot:
yeah, that works for me. the JSON is the simplest format I've come up with and extensible enough. The only thing that would be nice is strings for the capabilities instead of integers, but that largely matters for human-written JSON files.

whot:
Ok, even easier then! :)

PR #786, opened 2019-07-31
Revert "add quirk system & full support for the G305"

whot:
Reverts libratbag/libratbag#784

Sorry, need to revert. This fails to compile, there's a `;` missing after the enum declaration. which makes me wonder why circleci didn't run on this PR.

And looking at the git log (reviewed the whole diff before and didn't notice): 
```
+enum {
+       HIDPP20_QUIRK_NONE,
+}
+#define HIDPP20_QUIRK_G305     
```
That's fixed in the second patch so I guess that's a rebase gone wrong?

And the test suite now fails with `AssertionError: Quirk must be in ['DeviceIndex']`. Also something that should've been picked up by the CI which for some reason didn't run.

### DIFF
--- a/data/devices/logitech-g305.device
+++ b/data/devices/logitech-g305.device
@@ -1,7 +1,0 @@
-[Device]
-Name=Logitech Gaming Mouse G305
-DeviceMatch=usb:046d:4074
-Driver=hidpp20
-
-[Driver/hidpp20]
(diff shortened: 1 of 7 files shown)


whot:
Ok, circleci is broken right now which is why it didn't run, see #787

PR #788, opened 2019-07-31
Update to Circle CI 2

whot:


whot:
Looks like we were already on 2.0 anyway except for the config location that needed to be moved. And breakage introduced in #780 which I reverted in this branch.

PR #790, opened 2019-08-02
Release script fixes (for piper, mainly)

whot:
hand-composing stuff is boring, let's re-use the script

### DIFF
--- a/release.sh
+++ b/release.sh
@@ -314,11 +314,11 @@ process_module() {
     pkg_version=$($GREP -m 1 '^	version : ' meson.build | sed 's|	version : \([^\s,]*\).*|\1|' | sed "s/'//g")
     tar_name="$pkg_name-$pkg_version"
     targz=$tar_name.tar.gz
-    tag_name=$(echo "v$pkg_version" | sed 's|.0$||')
+    tag_name=$(echo "v$pkg_version" | sed 's|\.0$||')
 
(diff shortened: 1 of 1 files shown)


whot:
FTR, this is a copy of the xorg `release.sh` script and it's only run by maintainers for releases. I'd say that as long as it's good enough it'll do the job, but if you do make changes, making it for the upstream script too would be very appreciated. Also note, the upstream script at least has to run on different platforms which may explain a few things...

https://gitlab.freedesktop.org/xorg/util/modular/

FFY00:
You use `/bin/bash` ion the shebang instead of `/bin/sh` so we can make use of bash specific features. Can I assume GNU's coreutils or is it supposed to be full posix compliant?

whot:
for libratbag? linux-only, basically, so assume whatever a sensible baseline is (we use glib and libsystemd already, so...)

for the upstream one? I think that one has to work on openbsd in bad weather when the moon is in the house of aquarius. Which is basically why no-one wants to touch it.

FFY00:
Then perhaps you should drop the bash dependency and make it full posix compliant, shell + commands.

whot:
lol. I just went "what, this uses bash??" and then found that it was me who changed it to bash 5 years ago, with exactly the argument of "bugger that, too complicated to support sh".

So go for it, add whatever bashism you want. GNU coreutils - don't know, but let's see if anyone complains.

PR #791, opened 2019-08-02
add quirk system & full support for the G305

FFY00:
#784 had a bad rebase.

### DIFF
--- a/src/hidpp20.c
+++ b/src/hidpp20.c
@@ -48,7 +48,6 @@ hidpp20_feature_get_name(uint16_t feature)
 {
 	static char numeric[8];
 	const char *str;
-#define CASE_RETURN_STRING(a) case a: return #a; break
 
 	switch(feature)
 	{
@@ -99,9 +98,19 @@ hidpp20_sw_led_control_get_mode_string(const enum hidpp20_led_sw_ctrl_led_mode m
 		str = numeric;
 		break;
 	}
-#undef CASE_RETURN_STRING
 
 	return str;
+}
+
+const char*
+hidpp20_get_quirk_string(enum hidpp20_quirk quirk)
+{
+	switch (quirk) {
+	CASE_RETURN_STRING(HIDPP20_QUIRK_NONE);
+	CASE_RETURN_STRING(HIDPP20_QUIRK_G305);
+	}
+
+	abort();
 }
 
 static int
--- a/src/hidpp20.c
+++ b/src/hidpp20.c
@@ -2611,6 +2611,15 @@ hidpp20_onboard_profiles_initialize(struct hidpp20_device *device,
 						  HIDPP20_USER_PROFILES_G402,
 						  profiles->sector_size,
 						  data);
+
+	if (rc && device->quirk != HIDPP20_QUIRK_G305) {
+		/* The G305 has a bug where it throws an ERR_INVALID_ARGUMENT
+		   if the sector has not been written to yet. If this happens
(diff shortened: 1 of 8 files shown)


src/hidpp20.h
```diff
@@ -83,6 +83,7 @@ hidpp20_device_destroy(struct hidpp20_device *device);
 
 enum {
```
whot: can you make this a named enum place (e.g. `enum hiddp20_quirks`), that way you can use it as argument type.

src/hidpp20.c
```diff
@@ -99,6 +99,24 @@ hidpp20_sw_led_control_get_mode_string(const enum hidpp20_led_sw_ctrl_led_mode m
 		str = numeric;
 		break;
 	}
+
+	return str;
+}
+
+const char*
+hidpp20_get_quirk_string(unsigned quirk)
```
whot: if you use a named enum you can use it as argument type here, which means ...

src/hidpp20.c
```diff
@@ -99,6 +99,24 @@ hidpp20_sw_led_control_get_mode_string(const enum hidpp20_led_sw_ctrl_led_mode m
...
+{
+	static char numeric[8];
+	const char* str = NULL;
+	switch (quirk)
+	{
+	CASE_RETURN_STRING(HIDPP20_QUIRK_NONE);
+	CASE_RETURN_STRING(HIDPP20_QUIRK_G305);
+	default:
```
whot: ... that you can skip the `default` case and have the compiler warn you about any missing `case:` statements for this enum, and ...

src/hidpp20.c
```diff
@@ -99,6 +99,24 @@ hidpp20_sw_led_control_get_mode_string(const enum hidpp20_led_sw_ctrl_led_mode m
...
+		sprintf_safe(numeric, "%#4x", quirk);
+		str = numeric;
+		break;
+	}
 #undef CASE_RETURN_STRING
 
 	return str;
```
whot: you can just call `abort()` here, because this is internal API and if someone passes an invalid quirk we should blow up early to make sure this bug cannot ever be shipped.

src/hidpp20.c
```diff
@@ -99,11 +99,23 @@ hidpp20_sw_led_control_get_mode_string(const enum hidpp20_led_sw_ctrl_led_mode m
...
 
+const char*
+hidpp20_get_quirk_string(enum hidpp20_quirk quirk)
+{
+	switch (quirk)
+	{
+	CASE_RETURN_STRING(HIDPP20_QUIRK_NONE);
+	CASE_RETURN_STRING(HIDPP20_QUIRK_G305);
+	}
+#undef CASE_RETURN_STRING
```
whot: two things here: `{` needs to be on the same line as the `switch`. And instead of `undef` here, just move the `CASE_RETURN_STRING` into the top-level, i.e. below the `#include` statements so we don't need to undef it.

Having a `define` in one function but the `undef` in another is bad form.

src/libratbag-data.c
```diff
@@ -160,6 +163,9 @@ init_data_hidpp20(struct ratbag *ratbag,
 		data->hidpp20.index = num;
 	if (error)
 		g_error_free(error);
+
+	str = g_key_file_get_string(keyfile, group, "Quirk", NULL);
```
whot: move that to the next commit please, the compiler should complain about this being unused.

PR #810, opened 2019-09-09
hidpp20: add suport for the G602

FFY00:

### DIFF
--- a/src/driver-hidpp20.c
+++ b/src/driver-hidpp20.c
@@ -859,7 +859,8 @@ hidpp20drv_read_resolution_dpi(struct ratbag_profile *profile)
 	unsigned int default_rate = 500;
 	int rc;
 
-	if (drv_data->capabilities & HIDPP_CAP_RESOLUTION_2200) {
+	if ((drv_data->capabilities & HIDPP_CAP_RESOLUTION_2200) &&
+	    (drv_data->capabilities & HIDPP_CAP_SWITCHABLE_RESOLUTION_2201) == 0) {
 		uint16_t resolution;
 		uint8_t flags;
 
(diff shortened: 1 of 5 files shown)


src/driver-hidpp20.c
```diff
@@ -859,7 +859,8 @@ hidpp20drv_read_resolution_dpi(struct ratbag_profile *profile)
 	unsigned int default_rate = 500;
 	int rc;
 
-	if (drv_data->capabilities & HIDPP_CAP_RESOLUTION_2200) {
+	if (drv_data->capabilities & HIDPP_CAP_RESOLUTION_2200 &&
```
whot: doesn't the compiler complain about this? either way, please add `()` around this bit too.

src/driver-hidpp20.c
```diff
@@ -1305,6 +1306,9 @@ hidpp20drv_init_feature(struct ratbag_device *device, uint16_t feature)
 		break;
 	}
 	case HIDPP_PAGE_COLOR_LED_EFFECTS: {
+		if (ratbag_device_data_hidpp20_get_quirk(device->data) == HIDPP20_QUIRK_G602)
```
whot: needs a short comment explaining why this needs to be done, if for no other reason to make it obvious whether the code does what we intend do do.

src/driver-hidpp20.c
FFY00: Not really, it will do the & first and then the &&. Either way, I added the parenthesis.

whot: weird, I was certain that the compiler warned about this. Anyway, just fwiw, this is a general style issue because the parens make the code immediately obvious and there's less chance of a bug being introduced. Technically not needed, humanly it usually is :)

k5nn:
Hello I'm using a G602 and found that for some reason restarting stoping and restarting ratbagd and then restarting piper makes piper detect the mouse but for some reason one of the keys is missing from the mapping

[https://imgur.com/a/m78K3hd](url)

PR #813, opened 2019-09-10
Emulation tests

FFY00:
~~@bentiss I didn't do some of the things you asked me:~~
  - ~~Start `ratbag-emu` using `Popen`: this is tricky to integrate with pytest, it's the same issue I had in the ratbag-emu test suite.  If you think it's worth to spend more time making this work let me know.~~
  - ~~Use `ratbag-command`: I was having some trouble using it so I changed to `ratbagctl` and it worked perfectly. I don't really think it matters for this simple POC.~~

~~I just want some feedback right now, I can deal with the two points above next, if you still think they are relevant.~~

@whot @bentiss could you have a look at the structure of the tests I implemented and tell me if it's fine?

Instructions:
  - Start `ratbagd`
  - Start ratbag-emu (`src/main.py`)
  - Invoke `pytest` on libratbag's root directrectory

### DIFF
--- a/test/test_emulation/__init__.py
+++ b/test/test_emulation/__init__.py
@@ -15,9 +15,8 @@ MM_TO_INCH = 0.0393700787
 
 
 class RatbagemuClient(object):
-    def __init__(self, url='http://localhost:8080'):
-        self.url = url
-
+    def __init__(self, url='http://localhost', port=8080):
+        self.url = f'{url}:{port}'
 
     # Internal wrappers around the http requests
     def _get(self, path):
--- a/test/test_emulation/__init__.py
+++ b/test/test_emulation/__init__.py
@@ -8,7 +8,7 @@ import pytest
 import threading
 
 from pathlib import Path
-from time import sleep
+from time import sleep, strftime, time
 
 
 MM_TO_INCH = 0.0393700787
@@ -90,20 +90,17 @@ class TestDevice(object):
     @pytest.fixture(autouse=True, scope='session')
     def server(self):
         p = None
-        try:
-            stdout = open('ratbag-emu-log-stdout.txt', 'w')
-            stderr = open('ratbag-emu-log-stderr.txt', 'w')
-
-            p = subprocess.Popen(['/usr/bin/env', 'python3', '-m', 'ratbag_emu'], stdout=stdout, stderr=stderr)
-            sleep(2)
-            yield
-        finally:
-            if p:
-                p.kill()
-            if stdout:
-                stdout.close()
-            if stderr:
-                stderr.close()
+        logs = f'ratbag-emu-log-{strftime("%Y-%m-%d_%H:%M")}'
+        with open(f'{logs}-stdout.txt', 'w') as stdout, \
+             open(f'{logs}-stderr.txt', 'w') as stderr:
+            try:
+                args = ['/usr/bin/env', 'python3', '-m', 'ratbag_emu']
+                p = subprocess.Popen(args, stdout=stdout, stderr=stderr)
+                sleep(2)
+                yield
+            finally:
+                if p:
+                    p.kill()
 
     @pytest.fixture(autouse=True, scope='session')
     @pytest.mark.usesfixtures('server')
--- a/test/test_emulation/__init__.py
+++ b/test/test_emulation/__init__.py
@@ -88,6 +88,23 @@ class TestDevice(object):
             self.reload_udev_rules()
 
     @pytest.fixture(autouse=True, scope='session')
+    def ratbagd_src(self):
+        ratbagd_src = Path('tools/ratbagd.py')
+        ratbagd_dest = Path('test/test_emulation/ratbagd.py')
+        shutil.copyfile(ratbagd_src, ratbagd_dest)
+
+        yield
+
+        if ratbagd_dest.is_file():
+            ratbagd_dest.unlink()
+
+    @pytest.fixture(autouse=True)
+    @pytest.mark.usesfixtures(['ratbagd_src', 'id'])
+    def ratbagd(self):
+        import test_emulation.ratbagd
+        yield ratbagd.Ratbagd(api_version=1)
+
+    @pytest.fixture(autouse=True, scope='session')
     def server(self):
         p = None
         logs = f'ratbag-emu-log-{strftime("%Y-%m-%d_%H:%M")}'
@@ -107,7 +124,8 @@ class TestDevice(object):
     def client(self):
         yield RatbagemuClient()
 
-    @pytest.fixture
+    @pytest.fixture(autouse=True)
+    @pytest.mark.first
     def id(self, client):
         id = client.create(self.shortname)
 
--- a/test/test_emulation/__init__.py
+++ b/test/test_emulation/__init__.py
@@ -88,7 +88,7 @@ class TestDevice(object):
             rules_dest.unlink()
             self.reload_udev_rules()
 
-    @pytest.fixture(autouse=True)
+    @pytest.fixture(autouse=True, scope='session')
     def client(self):
         yield RatbagemuClient()
 
--- a/test/test_emulation/__init__.py
+++ b/test/test_emulation/__init__.py
@@ -92,12 +92,19 @@ class TestDevice(object):
     def server(self):
         p = None
         try:
-            p = subprocess.Popen('python -m ratbag_emu 2&>1 >/dev/null', shell=True)
+            stdout = open('ratbag-emu-log-stdout.txt', 'w')
+            stderr = open('ratbag-emu-log-stderr.txt', 'w')
+
+            p = subprocess.Popen(['/usr/bin/env', 'python3', '-m', 'ratbag_emu'], stdout=stdout, stderr=stderr)
             sleep(2)
             yield
         finally:
             if p:
                 p.kill()
+            if stdout:
+                stdout.close()
+            if stderr:
+                stderr.close()
 
     @pytest.fixture(autouse=True, scope='session')
     @pytest.mark.usesfixtures('server')
--- a/test/test_emulation/__init__.py
+++ b/test/test_emulation/__init__.py
@@ -92,6 +92,14 @@ class TestDevice(object):
     def client(self):
         yield RatbagemuClient()
 
+    @pytest.fixture
+    def id(self, client):
+        id = client.create(self.shortname)
+
+        yield id
+
+        client.delete(id)
+
     def _simulate(self, client, id, data):
         input_nodes = client.get_input_nodes(id)
 
--- a/test/test_emulation/__init__.py
+++ b/test/test_emulation/__init__.py
@@ -89,6 +89,18 @@ class TestDevice(object):
             self.reload_udev_rules()
 
     @pytest.fixture(autouse=True, scope='session')
+    def server(self):
+        p = None
+        try:
+            p = subprocess.Popen('python -m ratbag_emu 2&>1 >/dev/null', shell=True)
+            sleep(2)
+            yield
+        finally:
+            if p:
+                p.kill()
+
+    @pytest.fixture(autouse=True, scope='session')
+    @pytest.mark.usesfixtures('server')
     def client(self):
         yield RatbagemuClient()
 
--- a/test/test_emulation/__init__.py
+++ b/test/test_emulation/__init__.py
@@ -34,10 +34,12 @@ class RatbagemuClient(object):
     '''
     ratbag-emu functions
     '''
-    def create(self, shortname):
+    def create(self, shortname, initial_state=None):
         data = {
             'shortname': shortname
         }
+        if initial_state:
+            data['initial_state'] =  initial_state
         response = self._post('/devices/add', json=data)
         assert response.status_code == 201
         return response.json()['id']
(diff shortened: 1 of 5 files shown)


FFY00:
Sorry, forgot to select "Draft pull-request" in the dropdown.

test/test_emulation/test_steelseries.py
```diff
@@ -0,0 +1,29 @@
...
+
+    def _test_dpi(self, client, dpi_id):
+        id = client.create('steelseries-rival310')
+
+        old_dpi = 500
+        new_dpi = 1000
+
+        client.set_dpi(id, dpi_id, old_dpi)
+        subprocess.run(f'ratbagctl ratbag-emu resolution {str(dpi_id)} dpi set {str(new_dpi)}', shell=True)
+        assert client.get_dpi(id, dpi_id)
```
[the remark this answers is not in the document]
FFY00: Could you check the new code I pushed?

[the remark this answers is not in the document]
bentiss: Yep, looks better.

I just realized that the code in TestSteelseriesDevice2 is *very* generic and could use subclassing. How about you create a TestBaseDevice with the content of your current class, and subclass it in `test_steelseries.py` so you can just have the same tests for free on Logitech devices?

test/test_emulation/test_steelseries.py
```diff
@@ -0,0 +1,39 @@
...
+        client.delete(id)
+
+    def test_dpi(self, client, dpi_id=0):
+        id = client.create('steelseries-rival310')
+
+        old_dpi = 500
+        new_dpi = 1000
+
+        client.set_dpi(id, dpi_id, old_dpi)
+        subprocess.run(f'ratbagctl ratbag-emu resolution {str(dpi_id)} dpi set {str(new_dpi)}', shell=True)
```
bentiss: Also, this calls `ratbagctl` with `ratbag-emu` as device parameter. This works OK in the case you have only one test running, but will break when several are running in parallel.

A solution could be to generate a random device name and use it there instead.

FFY00:
@bentiss I am now starting ratbag-emu from pytest, you might want to test that.

bentiss:
>  I am now starting ratbag-emu from pytest, you might want to test that.

Hmm, that's a progress, but there are at least 2 issues:
- hardcoding `python` won't work on Fedora, we need to call `python3` there
- we need to install `ratbag-emu` globally (or maybe adjust `$PYTHON_PATH`), so this is not very handy for development

IMO, we would need to use the various capabilities of `meson` to be able to work around those 2 issues. IOW, not something to tackle right now, so please add a `# TODO`

test/test_emulation/__init__.py
```diff
@@ -0,0 +1,168 @@
...
+from time import sleep, time
+
+
+class RatbagemuClient(object):
+    def __init__(self, url='http://localhost:8080'):
+        self.url = url
+
+    '''
+    Internal wrappers around the http requests
+    '''
```
whot: just a fwiw: AIUI any `'''` is a proper python docstring, which in this case doesn't make sense since anything underscore is pseudo-private in python. So for comments that merely explain structure of the code, use `#` insteadl.

test/test_emulation/__init__.py
```diff
@@ -0,0 +1,168 @@
...
+                             data=data,
+                             json=json)
+
+    def _delete(self, path):
+        return requests.delete(f'{self.url}{path}')
+
+    def _put(self, path, data=None, json=None):
+        return requests.put(f'{self.url}{path}',
+                            data=data,
+                            json=json)
```
whot: nickpick but no requirement: these are short enough to fit on one line

test/test_emulation/__init__.py
```diff
@@ -0,0 +1,168 @@
...
+    @pytest.fixture(scope='session', autouse=True)
+    def udev_rules(self):
+        rules_dir = Path('/run/udev/rules.d')
+        rules_dir.mkdir(exist_ok=True)
+        shutil.copyfile('test/test_emulation/rules.d/61-ratbag-emu-ignore-test-devices.rules', '/run/udev/rules.d/61-ratbag-emu-ignore-test-devices.rules')
+        self.reload_udev_rules()
+
+        yield
+
+        rules = rules_dir.joinpath('61-ratbag-emu-ignore-test-devices.rules')
```
whot: same string for the `.rules` file three times and you're re-typing the `rules.d` even though you already have `rules_dir`.
```
fname = '61-ratbag-....rules'
srcpath = Path('test/test_emulation/rules.d/') / fname
dstpath = rules_dir / fname
shutil.copyfile ...

yield

if dstpath.is_file():
 ...

```
wherever some function is unhappy about a `Path`, use `os.fspath(Path)`

test/test_emulation/__init__.py
```diff
@@ -0,0 +1,168 @@
...
+        if rules.is_file():
+            rules.unlink()
+            self.reload_udev_rules()
+
+    @pytest.fixture(autouse=True, scope='session')
+    def server(self):
+        p = None
+        try:
+            p = subprocess.Popen('python -m ratbag_emu 2&>1 >/dev/null', shell=True)
```
whot: see [popen](https://docs.python.org/3/library/subprocess.html#subprocess.Popen), you can route `stdin` etc to `subprocess.DEVNULL` rather than relying on a shell feature.

test/test_emulation/__init__.py
```diff
@@ -0,0 +1,168 @@
...
+        self.x = x
+        self.y = y
+
+    @staticmethod
+    def from_mm(dpi, x=0, y=0):
+        return MouseData(x=int(x * MM_TO_INCH * dpi),
+                         y=int(y * MM_TO_INCH * dpi))
+
+
+MM_TO_INCH = 0.0393700787
```
whot: maybe I'm just old-fashioned here but constants usually go at the top of the file :)

test/test_emulation/test_steelseries.py
```diff
@@ -0,0 +1,36 @@
...
+
+    def test_create(self, id, client):
+        subprocess.check_call(f"ratbagctl 'ratbag-emu test device #{id}' info >/dev/null", shell=True)
+
+    def test_dpi(self, id, client, dpi_id=0):
+        old_dpi = 500
+        new_dpi = 1000
+
+        client.set_dpi(id, dpi_id, old_dpi)
+        subprocess.check_call(f"ratbagctl 'ratbag-emu test device #{id}' resolution {str(dpi_id)} dpi set {str(new_dpi)}", shell=True)
```
whot: do we need the shell here?

whot:
> hardcoding python won't work on Fedora, we need to call python3 there

`/usr/bin/env python3` I think should work everywhere.

> we need to install ratbag-emu globally (or maybe adjust $PYTHON_PATH), so this is not very handy for development

if we call it as a module like we do here, just adjusting `$PYTHONPATH` should be enough.

test/test_emulation/__init__.py
```diff
@@ -0,0 +1,175 @@
...
+
+            p = subprocess.Popen(['/usr/bin/env', 'python3', '-m', 'ratbag_emu'], stdout=stdout, stderr=stderr)
+            sleep(2)
+            yield
+        finally:
+            if p:
+                p.kill()
+            if stdout:
+                stdout.close()
+            if stdout:
```
bentiss: typo: `s/stdout/stderr/`

test/test_emulation/__init__.py
```diff
@@ -92,12 +92,19 @@ def udev_rules(self):
     def server(self):
         p = None
         try:
-            p = subprocess.Popen('python -m ratbag_emu 2&>1 >/dev/null', shell=True)
+            stdout = open('ratbag-emu-log-stdout.txt', 'w')
+            stderr = open('ratbag-emu-log-stderr.txt', 'w')
+
+            p = subprocess.Popen(['/usr/bin/env', 'python3', '-m', 'ratbag_emu'], stdout=stdout, stderr=stderr)
```
whot: I think you may be able to hook up a logger for this too and then pass the logger through as stdout/stderr. Something to put onto the nice-to-have list.

test/test_emulation/__init__.py
```diff
@@ -92,12 +92,19 @@ def udev_rules(self):
...
             sleep(2)
             yield
         finally:
             if p:
                 p.kill()
+            if stdout:
+                stdout.close()
+            if stderr:
+                stderr.close()
```
whot: is there any case where stdout or stderr can be None?

test/test_emulation/__init__.py
```diff
@@ -92,12 +92,19 @@ def udev_rules(self):
     def server(self):
         p = None
         try:
-            p = subprocess.Popen('python -m ratbag_emu 2&>1 >/dev/null', shell=True)
+            stdout = open('ratbag-emu-log-stdout.txt', 'w')
+            stderr = open('ratbag-emu-log-stderr.txt', 'w')
```
whot: lesson learned a few times: always use something date/timestamped like `filename-2019-09-18-13:00:12`. Worst case you'd have to delete a bunch of files, best case you didn't just overwrite the last log that had important information in it.

Also, after opening this write the current date/time into the file too. That way it's preserved even where the filename changes (e.g. when you post on pastebin)

test/test_emulation/__init__.py
FFY00: Maybe? I am not sure.

bentiss: see https://github.com/libratbag/ratbag-emu/commit/6f6a21ed6947b7268b954bd4a1d78eaa9faec0a6 with a `with` statement you don't need to check and close the file, the context manager will do it for you.

FFY00: Sure.

FFY00:
Needs https://github.com/libratbag/ratbag-emu/pull/21

FFY00:
IIRC the only thing missing is adding an `initial_state` parameter in ratbag-emu to replace `client.set_dpi(id, dpi_id, old_dpi)`.

test/test_emulation/__init__.py
```diff
@@ -90,20 +90,17 @@ def udev_rules(self):
...
-            sleep(2)
-            yield
-        finally:
-            if p:
-                p.kill()
-            if stdout:
-                stdout.close()
-            if stderr:
-                stderr.close()
+        logs = f'ratbag-emu-log-{strftime("%Y-%m-%d_%H-%M")}'
```
whot: use `%H:%M` because that's what we'd expect for time (well, at least the people who can handle 24h notation ;)

test/test_emulation/__init__.py
```diff
@@ -87,6 +87,23 @@ def udev_rules(self):
...
+        yield
+
+        if ratbagd_dest.is_file():
+            ratbagd_dest.unlink()
+
+    @pytest.fixture(autouse=True)
+    @pytest.mark.usesfixtures(['ratbagd_src', 'id'])
+    def ratbagd(self):
+        import test_emulation.ratbagd
+        yield ratbagd.Ratbagd(1)
```
whot: I actually had to look up what the 1 refers to. Please change this to

```
yield ratbagd.Ratbagd(api_version=1)
```
Code is only written once, but read a thousand times. So you always optimise for the reader, not the writer.

test/test_emulation/test_steelseries.py
```diff
@@ -1,20 +1,26 @@
...
         old_dpi = 500
         new_dpi = 1000
 
         client.set_dpi(id, dpi_id, old_dpi)
-        subprocess.check_call(f"ratbagctl 'ratbag-emu {id}' resolution {str(dpi_id)} dpi set {str(new_dpi)}", shell=True)
+
+        device = ratbagd.find_device(f'ratbag-emu {id}')
+        device.active_profile.resolutions[dpi_id].resolution = (new_dpi,)
+        device.commit()
+        sleep(0.1)
```
whot: libratbag note: this indicates that we need a DBus API to tell us when everything succeeded. Some devices take much longer than 100ms to commit (which is why we have the async commit API) and hoping that a sleep is long enough isn't ideal here.

cc @bentiss

test/test_emulation/test_steelseries.py
```diff
@@ -7,12 +7,19 @@
...
+        ],
+        'active_dpi': 1
+    }
 
     def test_create(self, id, ratbagd, client):
         assert ratbagd.find_device(f'ratbag-emu {id}')
 
     def test_dpi(self, id, ratbagd, client, dpi_id=0):
-        old_dpi = 500
+        old_dpi = self.initial_state['dpi'][self.initial_state['active_dpi']]
```
FFY00: This doesn't seem great. Suggestions?

FFY00:
This seems to be ready!

FFY00:
@whot @bentiss

bentiss:
I'd like to take a few more days before we hit the merge button on this PR. I really don't have the time right now to review the code, and I wonder if we should not have a more mature api on ratbag-emu first.

Don't take it personally, sending this PR means you achieved one big chunk of your internship. But OTOH, I would like to take a little bit more time to think about it before we introduce a dependency on libratbag to a distinct project.

FFY00:
Sure, I totally understand, and I agree.

FFY00:
Closing until the ratbag-emu work gets finished.

PR #817, opened 2019-09-16
Create logitech-g502-spectrum.device

matt24smith:
Adding support for my mouse

### DIFF
--- a/data/devices/logitech-g502-spectrum.device
+++ b/data/devices/logitech-g502-spectrum.device
@@ -1,0 +1,4 @@
+[Device]
+Name=Logitech G502 Spectrum
+DeviceMatch=usb:04d9:0169
+Driver=hidpp20


FFY00:
Please add it to `meson.build` too.

cvuchener:
g502 proteus spectrum is here: https://github.com/libratbag/libratbag/blob/master/data/devices/logitech-g502-proteus-spectrum.device

Yours has a Holtek vendor id. What is it?

FFY00:
Hum, what? A knockoff?

cvuchener:
I think it may be a keyboard:

https://antoinealb.net/2014/12/05/archlinux-install.html

> To do this, first use lsusb to find the device ID. For example for my WASD keyboard the relevant line is :
>
> > Bus 001 Device 006: ID 04d9:0169 Holtek Semiconductor, Inc.
>
> and the device ID here is 04d9:0169.

FFY00:
Yeah, takes makes more sense. @matt24smith you probably got the wrong id. Are you sure the latest release doesn't work?

matt24smith:
Yes you're right actually. The library wasn't working for me until after I created the file so I mistakenly thought I had fixed the issue with it. It appears to be working correctly now without the file I created. My mistake. 
Thanks for the quick response!

PR #823, opened 2019-09-25
swig: generate python helpers to use hidpp C library

FFY00:
```
$> sudo python3
Python 3.7.1 (default, Nov 23 2018, 10:01:49)
[GCC 8.2.1 20181105 (Red Hat 8.2.1-5)] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import hidpp
>>> f = open('/dev/hidraw1', 'r+b')
>>> base = hidpp.hidpp_device()
>>> hidpp.hidpp_device_init(base, f.fileno())
>>> dev = hidpp.hidpp20_device_new(base, 0xff)
>>> hidpp.hidpp20_feature_get_name(0x8100)
'HIDPP_PAGE_ONBOARD_PROFILES'
>>> hidpp.new_uint8_tp()
<Swig Object of type 'uint8_t *' at 0x7f865198fb70>
>>> feature_index = hidpp.new_uint8_tp()
>>> feature_type = hidpp.new_uint8_tp()
>>> feature_version = hidpp.new_uint8_tp()
>>> hidpp.hidpp_root_get_feature(dev, 0x8100, feature_index, feature_type, feature_version)
0
>>> hidpp.uint8_tp_value(feature_index)
14
```

### DIFF
--- a/src/hidpp.i
+++ b/src/hidpp.i
@@ -1,0 +1,23 @@
+/* wrapping libratbag functions from hidpp*.h using SWIG. */
+
+%module hidpp
+%{
+	/* the resulting C file should be built as a python extension */
+	#define SWIG_FILE_WITH_INIT
+	/*  Includes the header in the wrapper code */
+	#include <shared.h>
+	#include <hidpp-generic.h>
+	#include <hidpp20.h>
+%}
+#define __attribute__(x)
+#define _Static_assert(a,b)
+
+
+/*  Parse the header file to generate wrappers */
+%include "stdint.i"
+%include "cpointer.i"
+%pointer_functions(int, intp);
+%pointer_functions(uint8_t, uint8_tp);
+%pointer_functions(uint16_t, uint16_tp);
+%include "hidpp-generic.h"
+%include "hidpp20.h"
--- a/meson.build
+++ b/meson.build
@@ -618,6 +618,19 @@ shared_library(
    install: false,
 )
 
+i_source = join_paths(meson.source_root(), 'src', 'hidpp.i')
+c_source = swig_gen.process(i_source)
+shared_library(
+   'hidpp',
+   c_source,
+   name_prefix: '_',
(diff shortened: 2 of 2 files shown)


src/hidpp.i
```diff
@@ -0,0 +1,26 @@
...
+
+/*  Parse the header file to generate wrappers */
+%include "stdint.i"
+%include "cpointer.i"
+%pointer_functions(int, intp);
+%pointer_functions(uint8_t, uint8_tp);
+%pointer_functions(uint16_t, uint16_tp);
+/* include "libratbag.h"
+include "libratbag-enums.h"
+include "shared.h" */
```
bentiss: we likely need to remove this comments

whot:
I understand the basic need for it but whoah, that is one awful python API. I think it might be worth writing the proper python wrapper around it at the same time (á la python-libevdev), ideally before too much code relies on the swig API.

bentiss:
I understand the need for a proper python API, but OTOH, the idea is just to provide access to the C code of the hidpp.c files. What we want to achieve here is to ensure that each HID++ feature implemented in `ratbag-emu` can be trusted.

Ideally, I think we should just copy/paste the c files in a pytest fixture and generate the wrapper there, so that it keeps getting internal to ratabg-emu and we do not export it. Problem is, `hidpp*.[ch]` indirectly depend on `libratbag-private.h`.

Given that the `.so` is not installed, as long as we put the proper warnings, we should be safe.

Side note: I really wish we started libratabg as a python library, that would have saved a bunch of time :(

whot:
yeah, hindsight is always wiser but I doubt we'll find the time to rewrite it.

I had a look at the `libratbag-private.h` dependency and it falls into the same category. Disentangling is possible but effort for little gain other than correctness. So... not ideal either, let's just go with the swig bindings as you have them here.

PR #828, opened 2019-10-14
devices: add Logitech G604

### LINKED ISSUE #827: Logitech G604 Lightspeed
Hello libratbag Team, 

thanks a lot for the great work. Is there already any approaches for the new Logitech G604 Lightspeed? I own this mice and would like to use it in full on linux. So is there a way i can help you out?

Kind Regards

FFY00:
Fixes #827


FFY00:
It would be great if we could merge #829 before this.

whot:
> It would be great if we could merge #829 before this.

that one still has work pending, doesn't it?

FFY00:
Yes, we can merge this however.

PR #829, opened 2019-10-14
hidpp20: check supported report types

FFY00:


FFY00:
Ex of broken device: G604

bentiss:
I am pretty sure there are devices that need short commands instead of long. IIRC we had the case in the kernel and this backfired (or maybe this was only for HID++ 1.0 devices).

One thing the kernel is about to do: https://patchwork.kernel.org/patch/11184749/ (the second hunk) -> if the device doesn't export a HID++ short report ID, we can detect it and convert every request to use longs instead.

FFY00:
According to the HID++ 2.0 spec all devices should tolerate long reports. This might not be true in the real world though.

What we should actually do is abstract the report type away and replace with a smart function that will only use long reports when necessary and handle the cases where the devices don't support a certain report type.

bentiss:
> What we should actually do is abstract the report type away and replace with a smart function that will only use long reports when necessary and handle the cases where the devices don't support a certain report type.

looks good to me. This is similar to what the kernel is working on.

FFY00:
I am going to drop the optimal approach and just make a sanity check here.

whot:
> I am going to drop the optimal approach and just make a sanity check here.

sorry, EPARSE - what does that mean?

FFY00:
I am just going to add a check to see if the device supports certain report types and overwrite the report type in the message before sending if nor supported.

FFY00:
This is ready for review. @bentiss @whot 

Summary: Some devices don't support short report types. This patch introduces `hidpp_get_supported_report_types` which will check the supported report types. `hidpp20_request_command_allow_error` was changed to adjust the report type in case it is not supported.

FFY00:
Actually no, this has an issue. Please don't merge.

FFY00:
Sorted. It's ready now.

src/hidpp-generic.c
```diff
@@ -306,6 +306,63 @@ hidpp_read_response(struct hidpp_device *dev, uint8_t *buf, size_t size)
...
+
+	return 1;
+}
+
+void
+hidpp_get_supported_report_types(struct hidpp_device *dev, uint8_t device_index)
+{
+	dev->supported_report_types &= (0xff << 2); /* reset the bits we are gonna check */
+
+	if (hidpp_test_report(dev, device_index, REPORT_ID_SHORT, SHORT_MESSAGE_LENGTH))
```
bentiss: I get that we might need to ensure that a device exporting a report type might not answer it, but shouldn't we also check for the `ratbag_hid_report` presence of the report ID before even trying to communicate with the device?

That might even simplify the logic of hidpp_test_report() with the timeouts and retries.

src/hidpp20.c
```diff
@@ -133,6 +133,16 @@ hidpp20_request_command_allow_error(struct hidpp20_device *device, union hidpp20
 	}
 	msg->msg.address |= DEVICE_SW_ID;
 
+	/* some mice don't support short reports */
+	if (msg->msg.report_id == REPORT_ID_SHORT && !(device->base.supported_report_types & HIDPP_REPORT_SHORT))
+		msg->msg.report_id =  REPORT_ID_LONG;
```
bentiss: nitpick: extra space before `REPORT_ID_LONG`

src/hidpp20.c
```diff
@@ -2884,6 +2894,8 @@ hidpp20_device_new(const struct hidpp_device *base, unsigned int idx)
 	dev->proto_major = 1;
 	dev->proto_minor = 0;
 
+	hidpp_get_supported_report_types(&(dev->base), idx);
```
bentiss: we should ensure we have at least the long or short reports available here.

src/hidpp-generic.c
FFY00: What do you mean exactly? I should repeat the test from `hidpp_write_command`?

```
int
hidpp_write_command(struct hidpp_device *dev, uint8_t *cmd, int size)
{
	int fd = dev->hidraw_fd;
	int res;

	if (size < 1 || !cmd || fd < 0)
		return -EINVAL;
```

src/hidpp20.c
FFY00: I am not sure about that. There are a lot of features that don't require short reports. It would be sane for a device to not have short reports. Do you still want the check here, as opposed to being in `hidpp20_request_command_allow_error`?

src/hidpp-generic.c
bentiss: nope.

`struct ratbag_hidraw` has a `reports` field which contains `struct ratbag_hid_report`. The first field in this last struct is the report ID, so I guess just by reading the static device hid definition you can see if some reports are supported or not.

src/hidpp20.c
bentiss: No, I mean, a device not supporting both short and long reports is not suppoorted, correct? So if `->report_suported == 0`, then we should bail out.

FFY00: Nevermind, I misread your first reply :disappointed:.

src/hidpp-generic.c
FFY00: Thanks for the clarification, I will look into it.

FFY00: It does indeed solve the issue in a much better way! I didn't know we had this information available :) Thanks!

src/hidpp-generic.c
```diff
@@ -306,6 +306,33 @@ hidpp_read_response(struct hidpp_device *dev, uint8_t *buf, size_t size)
...
+		return;
+	}
+
+	/* reset the bits we are gonna check */
+	dev->supported_report_types &= (0xff & ~HIDPP_REPORT_SHORT);
+	dev->supported_report_types &= (0xff & ~HIDPP_REPORT_LONG);
+
+	for (unsigned i = 0; i < dev->num_hid_reports; i++)
+		if (dev->hid_reports[i].usage_page >= 0xff00 &&
+		    dev->hid_reports[i].usage_page <= 0xffff) /* vendor defined usage page (0xff00-0xffff) */
```
bentiss: nitpick: `(dev->hid_reports[i].usage_page & 0xff00) == 0xff00`

src/driver-hidpp20.c
```diff
@@ -1561,6 +1561,8 @@ hidpp20drv_probe(struct ratbag_device *device)
 	drv_data = zalloc(sizeof(*drv_data));
 	ratbag_set_drv_data(device, drv_data);
 	hidpp_device_init(&base, device->hidraw[0].fd);
+	base.hid_reports = (struct hidpp_hid_report*) device->hidraw[0].reports;
+	base.num_hid_reports = device->hidraw[0].num_reports;
```
bentiss: It doesn't make much sense to copy the variable in the struct when they are only used in one single function (`hidpp_get_supported_report_types()`). Please use local variables in `hidpp_get_supported_report_types()` instead if you don't like accessing `device->hidraw[0]` everywhere.

src/hidpp-generic.c
FFY00: I was going to do that but I didn't because I thought this would be a little more readable. I am going to change it.

src/driver-hidpp20.c
FFY00: Done.

PR #836, opened 2019-11-02
data: add Steelseries Kinzu V3

tomodachi:
Added support for the SteelSeries Kinzu V3


FFY00:
You also need to add the .device file in `meson.build`.

PR #838, opened 2019-11-04
Updated G602 device with the correct USB id

dhollinger:
Updated the G602 device id with the correct id found on the libratbag Wiki and the output I see from `lsusb` and `usbguard --list-devices`

### DIFF
--- a/data/devices/logitech-g602.device
+++ b/data/devices/logitech-g602.device
@@ -1,7 +1,7 @@
 # Logitech G602 over wireless USB
 [Device]
 Name=Logitech G602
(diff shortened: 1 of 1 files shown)


whot:
I'm pretty sure `c537` is a receiver, not the actual mouse. @FFY00?

we should have a CI test that just greps for the known receiver IDs and complains where they're added to a `.device` file :)

FFY00:
The current ID is correct. The ID you are using is the ID of the receiver. Since Linux 5.2 the device ID will be exported to the userspace. There might be other devices with the same receiver.

Issue #802, opened 2019-08-18
ratbagd API version is -1


Pshemas:
I've cleaned the meson build of ratbag and piper as there are up to date version in Opensuse repos.
But I'm having an issue that makes me scratch my head - when I run as root everything works, but when I try to run as regular user I get:

Traceback (most recent call last):
  File "/usr/bin/ratbagctl", line 2468, in <module>
    main(sys.argv[1:])
  File "/usr/bin/ratbagctl", line 2452, in main
    _r = open_ratbagd(verbose=cmd.verbose)
  File "/usr/bin/ratbagctl", line 2418, in open_ratbagd
    r = Ratbagd(1)
  File "/usr/bin/ratbagctl", line 266, in __init__
    raise RatbagdIncompatible(self.api_version or -1, api_version)
__main__.RatbagdIncompatible: ratbagd API version is -1 but we require 1

What have I messed up?

FFY00:
No, this is a version mismatch between piper and ratbagd.

Pshemas:
but why it works when I launch it as root?
Checked on a laptop (clean Opensuse Leap 15.1 install) and I get the same problem.

ratbagctl --version
0.10

sudo ratbagd --version
0.10

Piper at 0.3

whot:
I suspect that somehow the ratbagd that's running (possibly started by dbus activation) is an older version that's left over from somewhere. and there's some `PATH` setting that messes things up.

try running everything with the full path, i.e. `sudo /usr/bin/ratbagctl` and check if the error occurs - if so, then we have the culprit.

Pshemas:
as mentioned I've tried on a clean system (Opensuse too) and got the same.
Using full path doesn't help - if I run as regular user ( /usr/bin/ratbagctl) I get API error, but when I do (sudo /usr/bin/ratbagctl) it starts (and shows connected device).

whot:
run `d-feet` (a gnome dbus debugging GUI) and search for the `org.freedesktop.ratbagd1` bus, then go into the `/org/freedesktop/ratbagd1/` path and click on the `org.freedesktop.ratbagd1.Manager` interface. That should have a `APIVersion` property, what does it report when you double-click it?

lavadrop:
I have the same problem.
I have ratbag installed from the Ubuntu 19.04 repositories and Piper installed from Flathub.
This is what d-feet shows me:

⏷/org/freedesktop/ratbagd1/
    ∟⏷**Interfaces**
[...]

PR #840, opened 2019-11-11
Give a more specific error message if user cannot reach ratbagd

theMarix:
If access failed, e.g. due to insufficient permissions, don't confuse
the user by reporting an API mismatch.

This is a minimal improvement to target issue #802. As, if access is
restricted to a group, that group is known in Meson, it would even be
possible to report that group in the error message. But, that would
require converting ratbagd.py to a template or finding some other way to
inject that information at build time.

### DIFF
--- a/tools/ratbagd.py
+++ b/tools/ratbagd.py
@@ -252,13 +252,13 @@ class Ratbagd(_RatbagdDBus):
 
     def __init__(self, api_version):
         super().__init__("Manager", None)
-        result = self._get_dbus_property("Devices") or []
-        self._devices = [RatbagdDevice(objpath) for objpath in result]
-        self._proxy.connect("notify::g-name-owner", self._on_name_owner_changed)
-        if self.api_version == None and not self._proxy.get_cached_property_names():
+        result = self._get_dbus_property("Devices")
+        if result is None and not self._proxy.get_cached_property_names():
             raise RatbagdUnavailable("Make sure it is running and your user is in the required groups.")
         if self.api_version != api_version:
             raise RatbagdIncompatible(self.api_version or -1, api_version)
+        self._devices = [RatbagdDevice(objpath) for objpath in result or []]
+        self._proxy.connect("notify::g-name-owner", self._on_name_owner_changed)
 
     def _on_name_owner_changed(self, *kwargs):
         self.emit("daemon-disappeared")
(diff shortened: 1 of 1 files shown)


tools/ratbagd.py
```diff
@@ -255,6 +255,10 @@ def __init__(self, api_version):
         result = self._get_dbus_property("Devices") or []
         self._devices = [RatbagdDevice(objpath) for objpath in result]
         self._proxy.connect("notify::g-name-owner", self._on_name_owner_changed)
+        if self.api_version == None and not self._proxy.get_cached_property_names():
+            raise RatbagdUnavailable(
+                "Make sure it is running and your user is in the required groups."
+            )
```
FFY00: Same line.

tools/ratbagd.py
```diff
@@ -255,6 +255,10 @@ def __init__(self, api_version):
         result = self._get_dbus_property("Devices") or []
         self._devices = [RatbagdDevice(objpath) for objpath in result]
         self._proxy.connect("notify::g-name-owner", self._on_name_owner_changed)
+        if self.api_version == None and not self._proxy.get_cached_property_names():
```
whot: this seems a big backwards, even in the current implementation. we check for `Devices` first, then fail on the API version later. looks like instead of just checking the property names it makes more sense to check `Devices` for `None` and use that before assigning `self._devices`. IOW, reshuffling the order of `__init__` would be a good tidyup.

tools/ratbagd.py
```diff
@@ -252,11 +252,13 @@ class Ratbagd(_RatbagdDBus):
 
     def __init__(self, api_version):
         super().__init__("Manager", None)
-        result = self._get_dbus_property("Devices") or []
-        self._devices = [RatbagdDevice(objpath) for objpath in result]
-        self._proxy.connect("notify::g-name-owner", self._on_name_owner_changed)
+        result = self._get_dbus_property("Devices")
+        if result is None and not self._proxy.get_cached_property_names():
```
FFY00: `result is None` > `not result` unless you have a specific reason.

theMarix: I actually wanted to exclude the case where the list of devices is simply empty. Thus the explicit comparison with `None`. However, in that case the second condition will never trigger. Thus, it actually makes sense to use the simpler comparison.

FFY00: My bad then. I think it's best to leave it as it was if you have the time to change it again. We can wait for @whot though.

theMarix: Undid it for now. :)

FFY00: Thanks!

theMarix:
Fix the spelling errors in the commit message. Kind of you to talk only about two of them…

PR #841, opened 2019-11-16
Filename package cleanup for #800.

### LINKED ISSUE #800: Don't ship filenames with `[]` characters, please
Some characters are very difficult for packaging tools. You're shipping 2 files with `[]` characters and we have to manually rename them for including them in `clearlinux`. This will affect all RPM based distributions as far as I can see. I'd appreciate if these files can omit those characters so we don't need special handling.

lrenaud:
Renamed filenames to avoid brackets, and added a check to the meson.build
script to throw a build error when we find an illegal character in a filename.

Should close #800 

The error in the build script could become a warn, but that pushes up the meson version required to 0.44. There is already a build warning suggesting the minimum required is 0.50, but I don't know if that warning fails gracefully on older versions or not.


meson.build
```diff
@@ -338,6 +338,21 @@ data_files = files(
...
+	data_file_str = '@0@'.format(data_file)
+	foreach illegal_character : illegal_filename_characters
+		if data_file_str.contains(illegal_character)
+			error('illegal character "@1@" found in data file "@0@"'.format(
+				data_file_str, illegal_character
+			))
+			break
+		endif
+	endforeach
+endforeach
```
FFY00: This should be in a unit test, not the meson file.

Can you move it to `data/devices/data-parse-test.py`?

lrenaud: Easy enough. Any complaints if I use a regular expression to do that test?

FFY00: It should be fine :grin:.

lrenaud: I've moved it out of the build file, and squashed the commits.

data/devices/data-parse-test.py
```diff
@@ -190,6 +191,16 @@ def check_section_driver(driver, section):
...
 
 
+def validate_data_file_name(path):
+    # Matching any of the characters in the regular expression will throw an
+    # error. Currently only tests the square brackets [], parentheses, and curly
+    # braces.
+    illegal_characters_regex = '([\[\]\{\}\(\)])'
+    found_characters = re.findall(illegal_characters_regex, path)
+    if len(found_characters) > 0:
+        raise AssertionError('data file name \'{}\' contains illegal characters: \'{}\''.format(path, ''.join(found_characters)))
```
FFY00: I would use `if found_characters` and double quotes in the `AssertionError` to avoid having to escape the single quotes.

lrenaud: No sense in having me muck up something that would be cleaner otherwise. One more run on this guy then.

data/devices/data-parse-test.py
```diff
@@ -190,6 +191,16 @@ def check_section_driver(driver, section):
         assert('Unsupported driver section {}'.format(driver))
 
 
+def validate_data_file_name(path):
+    # Matching any of the characters in the regular expression will throw an
+    # error. Currently only tests the square brackets [], parentheses, and curly
+    # braces.
+    illegal_characters_regex = '([\[\]\{\}\(\)])'
+    found_characters = re.findall(illegal_characters_regex, path)
+    if len(found_characters) > 0:
```
FFY00: I think you missed this one. Should be `if found_characters:`, python will evaluate 0 to False and >=1 to True.

lrenaud: found_characters currently returns a list of matches. My goal was to spit out which characters were illegal later on in the request.

FFY00: Yes?

```python
if found_characters:
	raise AssertionError("data file name '{}' contains illegal characters: '{}'".format(path, ''.join(found_characters)))
```

An empty list will be evaluated to false (ignore my first comment about the 0 and >=1).

lrenaud: Ah! I didn't know a test on an empty list would evaluate to False.

FFY00: Yes, python is pretty smart when evaluating things!

lrenaud:
Looks like the build check failed on fedora due to something unrelated to this commit?

FFY00:
Yes, I restarted it.

PR #842, opened 2019-11-17
meson: fix minimum version

FFY00:
```
 * 0.46.0: {'Python Module'}
 * 0.50.0: {'install arg in configure_file'}
```

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -2,7 +2,7 @@ project('libratbag', 'c',
 	version : '0.11',
 	license : 'MIT/Expat',
 	default_options : [ 'c_std=gnu99', 'warning_level=2' ],
-	meson_version : '>= 0.40.0')
+	meson_version : '>= 0.50.0')
 
 # The DBus API version. Increase this every time the DBus API changes.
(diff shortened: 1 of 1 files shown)


FFY00:
Hum, maybe this was a false positive as it seemed to be working before?

whot:
meson has a bug where it still warns about these things even when the offending command is wrapped in meson version checks. e.g. we have
```
if meson.version().version_compare('<0.48.0')
 ....
```
and it still warns about the `import('python')` bit.

The second warning (install arg in configure file) is valid afaict but since it only affects the ratbagctl/lur-command man pages so I've mostly gone "meh" on that so far :)

FFY00:
Closing this then.

whot:
I switched the Fedora CI to F31 in #843 just now but for ubuntu I reckon it makes sense to keep 19.04 as a trigger for breaking things too hard. There's a large user-base on 19.04 still so let's not make our life harder than needed - asking them to install meson from pip is not helping our time bottlenecks.

PR #843, opened 2019-11-17
circle CI: switch to Fedora 31

whot:


whot:
sigh, I don't know why it was shown was 'all tests passed' but `diff` is missing and the CI fails right now. See #844 for the fix

PR #866, opened 2019-12-10
data: add a test for unifying receiver IDs

whot:
We don't want any device files that use the receiver itself as a match so
let's test for that. In virtually all cases this means the user must update
the kernel and hope support for that receiver is already present in that
version.

@bentiss, @FFY00 - what are the new receiver IDs again? I've only added `0xc52b` and `0xc532` here.

### DIFF
--- a/data/devices/receiver-check.py
+++ b/data/devices/receiver-check.py
@@ -1,0 +1,75 @@
+#!/usr/bin/env python3
+# vim: set expandtab shiftwidth=4 tabstop=4:
+#
+# Copyright © 2018 Red Hat, Inc.
+#
+# Permission is hereby granted, free of charge, to any person obtaining a
+# copy of this software and associated documentation files (the "Software"),
+# to deal in the Software without restriction, including without limitation
+# the rights to use, copy, modify, merge, publish, distribute, sublicense,
+# and/or sell copies of the Software, and to permit persons to whom the
+# Software is furnished to do so, subject to the following conditions:
+#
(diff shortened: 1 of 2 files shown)


FFY00:
Well, you can make the check not trigger the regex `c5*` as that's the receiver range.

whot:
all of that range? we have two devices within that range - g7 and g700-wireless with `c51a` and `c531`, respectively.

FFY00:
Yes, they are receivers. We don't have those devices in the kernel driver yet, we need testing.

Maybe it's better to just blacklist the receivers in the kernel. You can find them here:
https://github.com/torvalds/linux/blob/master/drivers/hid/hid-logitech-dj.c#L1826
https://github.com/torvalds/linux/blob/master/drivers/hid/hid-ids.h#L772

whot:
done!

data/devices/receiver-check.py
```diff
@@ -0,0 +1,74 @@
...
+# DEALINGS IN THE SOFTWARE.
+#
+
+import argparse
+import os
+import sys
+import configparser
+
+# see the IDs from around
+# https://github.com/torvalds/linux/blob/master/drivers/hid/hid-ids.h#L772
```
FFY00: Maybe we should also add https://github.com/torvalds/linux/blob/master/drivers/hid/hid-logitech-dj.c#L1826 here as that list the actual enabled receivers.

whot: you mean add that link? That I can do, done now :)

Issue #846, opened 2019-11-26
G305 not working


czrcbl:
Hi,

My G305 has just arrived, but Piper-libratbag does not detect it. I also have a G600 and it is normally detected by libratbag. I run `lsusb` and realized that the entry that corresponds to the G305 is `Bus 001 Device 008: ID 046d:c53f Logitech, Inc. CSR8510 A10`. It is different from the id that appears on the config file (`usb:046d:4074`). So I edited the config file and added my G305 ID, changing the line to `DeviceMatch=usb:046d:c53f;usb:046d:4074`. I have reboot but the device is still not detected.

FFY00:
It should be working out of the box in Linux 5.4, can you confirm?

czrcbl:
I am running Ubuntu 19.10, it will probably not update to 5.4. I also tried to update the kernel by force, but the gnome display manager wouldn't start, so I cannot get into the desktop to test it right now.

banditopazzo:
Hi, I'm on Arch with kernel `5.4.2-arch1-1` and I can confirm the G305 is not working.

This is the output when I plug in the USB receiver:
```
ratbag debug: New device: Logitech USB Receiver
ratbag debug: Using data directory '/usr/share/libratbag'
ratbag debug: No data file found for 046d:c53f
ratbag debug: New device: Logitech USB Receiver
ratbag debug: Using data directory '/usr/share/libratbag'
ratbag debug: No data file found for 046d:c53f
ratbag debug: New device: Logitech USB Receiver
ratbag debug: Using data directory '/usr/share/libratbag'
ratbag debug: No data file found for 046d:c53f
ratbag debug: New device: Logitech USB Receiver
ratbag debug: Using data directory '/usr/share/libratbag'
ratbag debug: No data file found for 046d:c53f
ratbag debug: New device: Logitech USB Receiver
ratbag debug: Using data directory '/usr/share/libratbag'
ratbag debug: No data file found for 046d:c53f
ratbag debug: New device: Logitech USB Receiver
ratbag debug: Using data directory '/usr/share/libratbag'
ratbag debug: No data file found for 046d:c53f
ratbag debug: New device: Logitech G305
ratbag debug: Using data directory '/usr/share/libratbag'
ratbag debug: device assigned driver hidpp20
ratbag debug: hidraw info: bus 0x03 vendor 0x46d product 0xffffc53f
ratbag debug: hidraw info: bus 0x03 vendor 0x46d product 0xffffc53f
[...]

Issue #833, opened 2019-10-27
G304 not supported? even though it is exactly the same as G305 which is supported


yashvone:

FFY00:
Probably just needs the a .device file for the device ID to be identified.

yashvone:
so after adding .device file would i need to build it and piper from source?

yashvone:
> Probably just needs the a .device file for the device ID to be identified.

so after adding .device file would i need to build it and piper from source?

FFY00:
No, you should be able to just drop the file in your filesystem. However, if you want the correct image to show up in piper you will need to add an entry in the resources file and rebuild it.

yashvone:
I tried getting the device id, but it only shows the usb reciever id. It's basically the same device as g305 but for non US regions. Am i doing something wrong? I tried "lsusb" and "ratbagctl"

FFY00:
Can you post the output of `dmesg | grep logitech`?

yashvone:
~$ dmesg | grep Logitech
[20412.460085] usb 1-6: Manufacturer: Logitech
[20412.466069] input: Logitech USB Receiver as /devices/pci0000:00/0000:00:14.0/usb1/1-6/1-6:1.0/0003:046D:C53F.0004/input/input23
[20412.527308] hid-generic 0003:046D:C53F.0004: input,hidraw0: USB HID v1.11 Keyboard [Logitech USB Receiver] on usb-0000:00:14.0-6/input0
[20412.531266] input: Logitech USB Receiver Mouse as /devices/pci0000:00/0000:00:14.0/usb1/1-6/1-6:1.1/0003:046D:C53F.0005/input/input24
[20412.589238] input: Logitech USB Receiver Consumer Control as /devices/pci0000:00/0000:00:14.0/usb1/1-6/1-6:1.1/0003:046D:C53F.0005/input/input25
[20412.589603] input: Logitech USB Receiver System Control as /devices/pci0000:00/0000:00:14.0/usb1/1-6/1-6:1.1/0003:046D:C53F.0005/input/input26
[20412.590222] hid-generic 0003:046D:C53F.0005: input,hiddev0,hidraw1: USB HID v1.11 Mouse [Logitech USB Receiver] on usb-0000:00:14.0-6/input1
[20412.594536] hid-generic 0003:046D:C53F.0006: hiddev1,hidraw2: USB HID v1.11 Device [Logitech USB Receiver] on usb-0000:00:14.0-6/input2
[...]

PR #868, opened 2019-12-12
drivers: fix G305 quirk detection

crab2313:
Fixes: #833 
Fixes: #846 

The second patch is not well tested. But I think it should work. Linux kernel version 5.4 causes serious regression of my system and force me to switch back to an old kernel.

### DIFF
--- a/src/driver-hidpp20.c
+++ b/src/driver-hidpp20.c
@@ -1509,7 +1509,6 @@ hidpp20drv_remove(struct ratbag_device *device)
 	if (drv_data->profiles)
 		hidpp20_onboard_profiles_destroy(drv_data->profiles);
 	free(drv_data->led_infos.color_leds_8070);
-	free(drv_data->led_infos.color_leds_8071);
 	free(drv_data->controls);
 	free(drv_data->sensors);
 	if (drv_data->dev)
--- a/src/hidpp20.c
+++ b/src/hidpp20.c
@@ -2627,7 +2627,7 @@ hidpp20_onboard_profiles_initialize(struct hidpp20_device *device,
 						  profiles->sector_size,
 						  data);
 
-	if (rc && device->quirk != HIDPP20_QUIRK_G305) {
+	if (rc && device->quirk == HIDPP20_QUIRK_G305) {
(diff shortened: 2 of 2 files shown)


src/driver-hidpp20.c
```diff
@@ -1509,7 +1509,6 @@ hidpp20drv_remove(struct ratbag_device *device)
 	if (drv_data->profiles)
 		hidpp20_onboard_profiles_destroy(drv_data->profiles);
 	free(drv_data->led_infos.color_leds_8070);
-	free(drv_data->led_infos.color_leds_8071);
 	free(drv_data->controls);
```
whot: whoah, yeah, that's a double free, that info is in a union. Though really we need to wrap the `free()` calls into a capability check, similar to `hidpp20drv_read_color_leds()`.

src/hidpp20.c
```diff
@@ -2627,7 +2627,7 @@ hidpp20_onboard_profiles_initialize(struct hidpp20_device *device,
 						  profiles->sector_size,
 						  data);
 
-	if (rc && device->quirk != HIDPP20_QUIRK_G305) {
+	if (rc && device->quirk == HIDPP20_QUIRK_G305) {
```
whot: @FFY00 - typo, right? this patch is correct?

FFY00: Yes!

src/driver-hidpp20.c
FFY00: We can make a macro to free and make the value NULL.

crab2313: @FFY00  I will remove the second commit if you decide to fix it later.

whot: > We can make a macro to free and make the value NULL.

we'd still rely on this struct to be free-able with a single call. better to be correct here and add the `if` so we don't introduce new bugs if we ever extend this.

FFY00:
Can you split this in two commits? One fixing the G305 quirk and one fixing the double free. Also, if you could use the `hidpp20` prefix instead of `drivers` it would be great!

crab2313:
@FFY00 Already two commits.

PR #874, opened 2019-12-15
HID++2.0 Keyboards

FFY00:


FFY00:
Please only merge after #829. I don't want this included in the new release.

src/driver-hidpp20.c
```diff
@@ -1179,6 +1179,34 @@ hidpp20drv_read_profile_8100(struct ratbag_profile *profile)
...
+
+	rc = hidpp20_onboard_profiles_initialize(drv_data->dev, drv_data->profiles);
+	if (rc < 0)
+		return rc;
+
+	drv_data->num_profiles = drv_data->profiles->num_profiles;
+	drv_data->num_buttons = drv_data->profiles->num_buttons;
+
+	if (drv_data->capabilities & HIDPP_CAP_SWITCHABLE_RESOLUTION_2201)
+		drv_data->num_resolutions = drv_data->profiles->num_modes;
```
FFY00: To be clearer, I added this check to the original code. This check needs the capabilities to be populated first. The original code was being run in the loop where we parse the features and populate the capabilities. It would work if the `AdjustableDpi` feature was found before `OnboardMemoryProfiles`.

PR #875, opened 2019-12-15
Add support for the Logitech G915

FFY00:
Needs #874


stephanlachnit:
Is it really a good idea to keep all device file in one directory, especially if we don't have a `DeviceType` property?

FFY00:
Yes? Each device file contains the driver name (`Driver=hidpp20`) and all the required configurations to work properly with that driver. I am not sure I follow.

stephanlachnit:
> Yes? Each device file contains the driver name (`Driver=hidpp20`) and all the required configurations to work properly with that driver. I am not sure I follow.

I think I was more thinking about piper, because here the UI handling would be different but it's not directly related to libratbag.

FFY00:
> I think I was more thinking about piper, because here the UI handling would be different but it's not directly related to libratbag.

Well, kind of. The SVGs are bigger, we would need to adapt piper for that. But right now it works fine with the default (fallback) SVG.

FFY00:
At least for me, the main thing is the ability to configure the leds, and that is working just fine. Button remapping is not great, but it was not great for mice either way.

whot:
ftr, piper only needs the files for "this device needs that SVG". whereas the libratbag files are "here  are the custom quirks for this device", including things like the driver name.  both files are internal API, so If we ever need a device type property, we can add it then.

PR #882, opened 2019-12-22
docs: adjust description to say we support other devices than mice

FFY00:
`... a DBus daemon to configure gaming mice.` > `... a DBus daemon to configure input devices, mainly gaming mice`


FFY00:
Since we now support keyboards, and we are likely to support the Logitech G Powerplay gamepad in the future.

PR #885, opened 2019-12-23
Use EditorConfig instead of vim comments

stephanlachnit:
Removed all `# vim:` lines, EditorConfig is available not just for [vim](https://github.com/editorconfig/editorconfig-vim), but basically any IDE: https://editorconfig.org/#download

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,20 @@
+root = true
+
+[*]
+charset = utf-8
+end_of_line = lf
+trim_trailing_whitespace = true
+insert_final_newline = true
+
+[*.{c,h}]
+indent_style = tab
+tab_width = 8
+trim_trailing_whitespace = true
+
+[{*.{py,py.in},tools/ratbagctl.*.in}]
+indent_style = space
+indent_size = 4
+
+[*.{yml,xsl}]
+indent_style = space
+indent_size = 2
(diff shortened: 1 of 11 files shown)


FFY00:
Will the vim plugin find `.editorconfig` if we are not in the root directory? Either way, I think it's better if we leave the original vim comments, can you remove that from the commit?

FFY00:
Also, since you are adding support for editorconfig we might as well just run the tests in the CI. But that could be done in a separate PR if you want.

stephanlachnit:
> Will the vim plugin find .editorconfig if we are not in the root directory?

Yes, it will! That's why there is the `root = true` line. Once the plugin found that, it stops searching.


> Also, since you are adding support for editorconfig we might as well just run the tests in the CI. But that could be done in a separate PR if you want.

Separate PR, no idea how to run tests for that.

FFY00:
> Separate PR, no idea how to run tests for that.

I can do that then.

.editorconfig
```diff
@@ -0,0 +1,20 @@
+root = true
+
+[*]
+charset = utf-8
+end_of_line = lf
```
whot: out of interest - is this really needed or isn't this the default anyway?

stephanlachnit: Don't thinks it needed since we all use a sane platform, but I think if it's not there it just won't be given to the editor, meaning system defaults. Can remove it.

whot: nah, not worth the churn, just leave it. Thanks!

PR #886, opened 2019-12-23
Update CircleCI config

stephanlachnit:
List dependencies separately.
Uninstall check merged in upstream meson, so it can be simplified.
`libsolv` doesn't need to be updated any more with Fedora 31.


stephanlachnit:
> This looks fine. Can you please just rename the commit?

Done.

Issue #716, opened 2019-06-15
MX Master settings not applying.


PixlOne:
Nothing happens when I change on of the libratbag settings on my MX Master.

ratbagctl returns no errors when I set an option but the new changes do not take effect.

When I try changing it in Piper, it says 'Something went wrong. The device has been reset to a previous state.'

whot:
start with `ratbagd --verbose` and try it then, do you see any error messages?

PixlOne:
Here's what I get with `ratbagd --verbose`:
`ratbag debug: New device: Sony PLAYSTATION(R)3 Controller
ratbag debug: Using data directory '/usr/local/share/libratbag'
ratbag debug: No data file found for 054c:0268
ratbag debug: New device: HID 04b4:1007
ratbag debug: Using data directory '/usr/local/share/libratbag'
ratbag debug: No data file found for 04b4:1007
ratbag debug: New device: HID 04b4:1007
ratbag debug: Using data directory '/usr/local/share/libratbag'
ratbag debug: No data file found for 04b4:1007
ratbag debug: New device: Logitech MX Master
ratbag debug: Using data directory '/usr/local/share/libratbag'
ratbag debug: device assigned driver hidpp20
ratbag debug: hidraw info: bus 0x03 vendor 0x46d product 0x4069
ratbag debug: Logitech MX Master is device '/dev/hidraw4'.
ratbag debug: HID report ID 01
ratbag debug: HID report ID 0e
ratbag debug: HID report ID 02
ratbag debug: HID report ID 03
ratbag debug: HID report ID 04
ratbag debug: HID report ID 10
ratbag debug: HID report ID 11
ratbag debug: HID report ID 20
ratbag debug: HID report ID 21
ratbag debug: 'Logitech MX Master' is using protocol v4.5
ratbag debug: device battery level is 90% (next 50%), status 0 
ratbag debug: device has programmable keys/buttons
ratbag debug: device has 8 buttons
ratbag debug: device has adjustable dpi
ratbag debug: device is at 1000 dpi (variable between 400 and 1600).
ratbag debug: device is at 1000 dpi (variable between 400 and 1600).
ratbag debug: device has 8 buttons
ratbag debug: driver match found: Logitech HID++2.0
hidraw4: "Logitech MX Master", 1 profiles
ratbag debug: New device: Logitech USB Receiver
ratbag debug: Using data directory '/usr/local/share/libratbag'
ratbag debug: No data file found for 046d:c52b
`

[...]

PR #890, opened 2019-12-26
hidpp20: implement dirty flag for report rate

FFY00:


FFY00:
This should fix #716 and #834.

src/driver-hidpp20.c
```diff
@@ -1423,7 +1423,7 @@ hidpp20drv_commit(struct ratbag_device *device)
 			continue;
 
 		rc = hidpp20drv_update_report_rate(profile, profile->hz);
-		if (rc)
+		if (rc && rc != -ENOTSUP)
```
whot: doesn't this mean we lose the information that the current report rate couldn't be set? e.g. if I want to change the rate it would now show as success rather than error? Do we need some extra check for whether it differs from the current rate or something?

FFY00: In #891 I added errors to the log to for this cases. Do we have a better way to deal with this? Well, actually I think we could have a dirty flag.

src/libratbag-private.h
```diff
@@ -295,6 +295,7 @@ struct ratbag_profile {
 	unsigned int hz;	/**< report rate in Hz */
 	unsigned int rates[8];	/**< report rates available */
 	size_t nrates;		/**< number of entries in rates */
+	bool rate_dirty;
```
whot: Experimental option (just typed out here):

```
struct ratbag_profile {
    [...]
    union {
        unsigned int is_dirty;
        struct {
            bool rate:1;
            bool resolutions:1;
        } dirty;
    };
};
```
This uses a C11 anonymous union and bitfields for the various dirty bits. And because it's a union we only need to check one field to check all of them.
Which would make the code easy to read and easy to check:
```
dev->dirty.rate = true;
if (dev->is_dirty) { ... }
```
as long as `sizeof(is_dirty)` is greater than all bits for the fields, we're good. Could even put a static assert for that in.

But really, I just like having all the dirty bits grouped in a struct to force the code to be readable.

How does that sound?

FFY00: Sure. We can do that in other PR, is that fine?

whot: sure, shouldn't block this one.

PR #891, opened 2019-12-26
Better errors

FFY00:
This could have helped spot #716/#834 much easier.

### DIFF
--- a/ratbagd/ratbagd-device.c
+++ b/ratbagd/ratbagd-device.c
@@ -172,6 +172,8 @@ static void ratbagd_device_commit_pending(void *data)
 	int r;
 
 	r = ratbag_device_commit(device->lib_device);
+	if (r)
+		log_error("error committing device (%d)\n", r);
 	if (r < 0)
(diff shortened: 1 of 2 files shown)


FFY00:
Rebased on top of #890

ratbagd/ratbagd-device.c
```diff
@@ -172,6 +172,8 @@ static void ratbagd_device_commit_pending(void *data)
 	int r;
 
 	r = ratbag_device_commit(device->lib_device);
+	if (r)
+		log_error("error committing device (%d)\n", r);
```
stephanlachnit: Shouldn't it be `log_error(ctx,"...")`?

FFY00: Ah, yeah. Thanks for catching that.

FFY00: Actually no, not here. I just checked and we only use that for the drivers.

stephanlachnit: Whoops yes, think I mixed something up.

PR #895, opened 2020-01-03
HID++ 2.0 `0x8060` support

FFY00:
Rebased on top of #890.

### DIFF
--- a/src/driver-hidpp20.c
+++ b/src/driver-hidpp20.c
@@ -822,9 +822,12 @@ hidpp20drv_read_report_rate_8060(struct ratbag_device *device)
 {
 	struct hidpp20drv_data *drv_data = ratbag_get_drv_data(device);
 	struct ratbag *ratbag = device->ratbag;
+	struct ratbag_profile *profile;
 	uint8_t bitflags_ms;
 	int nrates = 0;
 	int rc;
+	uint8_t rate_ms;
+	unsigned rate_hz;
 
 	rc = hidpp20_adjustable_report_rate_get_report_rate_list(drv_data->dev,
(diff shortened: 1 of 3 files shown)


FFY00:
No. A report rate of 0Hz will never be written to anywhere. Bellow we have

```
if (rate_hz) {
	log_debug(ratbag, "report rate is %u\n", rate_hz);
	ratbag_device_for_each_profile(device, profile)
		profile->hz = rate_hz;
} 
```

`if (rate_hz)` is the same as `if (rate_hz != 0)`. We set the `rate_hz` to 0 if no valid report rate is found and then only act on valid values.

src/driver-hidpp20.c
```diff
@@ -1189,12 +1189,12 @@ hidpp20drv_init_leds_8070_8071(struct ratbag_device *device)
 	if (!(drv_data->capabilities & HIDPP_CAP_ONBOARD_PROFILES_8100)) {
 		if (drv_data->capabilities & HIDPP_CAP_COLOR_LED_EFFECTS_8070) {
 			log_debug(ratbag, "disabling 0x8070 (Color Leds Effects) feature because the device doesn't have 0x8100 (Onboard Memory Profiles)\n");
-			drv_data->capabilities &= !HIDPP_CAP_COLOR_LED_EFFECTS_8070;
+			drv_data->capabilities &= ~HIDPP_CAP_COLOR_LED_EFFECTS_8070;
```
whot: general comment: i think a `set_cap` and `clear_cap` macro would be useful here.

FFY00: Yes. Will do in a separate PR.

PR #896, opened 2020-01-03
HID++ 2.0 Cleanup

FFY00:
We now have a routine to identify the supported report types so it is safe to go ahead and use short reports when possible.


FFY00:
When I was going over the functions I noticed one had an incorrect report type, I included the fix here.

PR #897, opened 2020-01-03
Silence Warnings

FFY00:
The first commit adds `-Wno-address-of-packed-member` which is triggered by accessing `metadata->lgs02.*` in `hidpp10.c`, example:

https://github.com/libratbag/libratbag/blob/master/src/hidpp10.c#L1413

This warnings spit out a lot of spam making it more difficult to identify other ones.


whot:
hmm, after reading about this warning, I think it may be more prudent to just go and fix those instances instead of papering over it. While not an immediate problem, it could become an issue if these are ever run on ARM, or some ARMs anyway.

FFY00:
We are working with strings, which are just char arrays, so the warning does not apply here. The problem would be if we were taking, for example, the pointer of a 64bit int. The compiler could pass that pointer directly to an ARM instruction which was expecting the pointer to be aligned. As we are working with bytes arrays there's nothing to worry about. There isn't really anything to fix, we need to access the data, there's no way of accessing that data without triggering the warning. The alternative to not having `-Wno-address-of-packed-member` would be to put `#pragma`s in the code ignoring the warning on the places we access the string.

FFY00:
Hum, I stand corrected. We are declaring `name` and `macro_names` as `uint16_t` arrays, which could cause trouble.

staticssleever668:
I don't think it's good to ignore the warning, as it does indeed show a potential error.
However, I'd consider this low-priority as to my knowledge on x86 (and ARM, but I'm less sure about that) it would've given a hardware exception.
Ignored warnings are extreme noise, but we should really just properly fix this.

PR #898, opened 2020-01-07
ratbagd: switch Profile.Enabled to Profile.Disabled

stephanlachnit:
Rebased single commit from whot/libratbag/wip/custom-vtables against master. Goal is to merge this before #870.
That being said, I think we should not just update the DBus API but also all the drivers, meaning replacing `ratbag_profile_set_enabled` with `ratbag_profile_set_enabled` so we don't have even more  confusion with negations. But for now this should be enough, maybe will do all the driver work in the future.

### DIFF
--- a/doc/dbus.rst
+++ b/doc/dbus.rst
@@ -206,12 +206,12 @@ org.freedesktop.ratbag1.Profile
         The name of this profile. If the name is the empty string, the
         profile name cannot be changed.
 
-.. attribute:: Enabled
+.. attribute:: Disabled
 
         :type: b
         :flags: read-write, mutable
 
-        True if this is the profile is enabled, false otherwise.
+        True if this is the profile is disabled, false otherwise.
 
         Note that a disabled profile might not have correct bindings, so it's
         a good thing to rebind everything before calling
--- a/ratbagd/ratbagd-profile.c
+++ b/ratbagd/ratbagd-profile.c
@@ -301,27 +301,27 @@ static int ratbagd_profile_set_active(sd_bus_message *m,
 }
 
 static int
-ratbagd_profile_set_enabled(sd_bus *bus,
-			    const char *path,
-			    const char *interface,
-			    const char *property,
-			    sd_bus_message *m,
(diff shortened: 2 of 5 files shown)


stephanlachnit:
> we'll have to bump the API version as well to make sure piper complains appropriately.

Ah yep. I guess it makes sense to do this with #870 together then, that should need a bump as well.

FFY00:
Well, I think we just need to worry about bumping it again after a release. So we could merge it now and bump the version and then merge #870 without bumping anything. This is just a minor comment, do as you please :)

stephanlachnit:
> Well, I think we just need to worry about bumping it again after a release. So we could merge it now and bump the version and then merge #870 without bumping anything. This is just a minor comment, do as you please :)

Ah true, I'll let @whot decide, it's his commit after all.

whot:
unless you have a very specific schedule that you can keep and you merge both within a day or two, it's better to bump every time. Otherwise the whole point of the version making sure both ends support it correctly is moot. And it's just an integer, we're not going to run out of those for ... a while.

Bumping the version is a lot quicker than having to deal with bugs when the two repos are out of sync despite the error message telling users to update both.

stephanlachnit:
True, libratbag isn't really regular with it's realeases. Speaking of that: I asked @FFY00 to do a release before the Ubuntu 20.04 import freeze to get the latest devices in this LTS releases which probably a lot of people will use, so if he wants to merge this he could add a version bump there, else I'm gonna close it again and add a version bump in #870 when I have time to rebase it.

FFY00:
> True, libratbag isn't really regular with it's realeases

Well, we don't have a regular development cycle, this is just done as best effort.

> Speaking of that: I asked @FFY00 to do a release before the Ubuntu 20.04 import freeze to get the latest devices in this LTS releases which probably a lot of people will use, so if he wants to merge this he could add a version bump there, else I'm gonna close it again and add a version bump in #870 when I have time to rebase it.

I think it's maybe best if we leave it out since I don't have much time to look at this and it's not that important.

PR #899, opened 2020-01-07
Minor hidpp20 driver work

stephanlachnit:
Removes useless setting of the dpi index when updating resolutions.
Removes weird empty line.


stephanlachnit:
Also added the commit that replaces the gotos (was previously in #870).

PR #906, opened 2020-01-29
Declare a bunch of structs as extern

whot:
gcc 10 complains about multiple definitions of these when linking.

```
/usr/bin/ld: libhidpp.a(src_hidpp-generic.c.o):/builddir/build/BUILD/libratbag-0.12/x86_64-redhat-linux-gnu/../src/libratbag-private.h:588:
multiple definition of `etekcity_driver';
libutil.a(src_libratbag-util.c.o):/builddir/build/BUILD/libratbag-0.12/x86_64-redhat-linux-gnu/../src/libratbag-private.h:588:
first defined here
```

Special case here is the test_driver struct, without the extern gcc seems to
think it's a new struct with all zeros, causing a segfault when we then try to
strcmp the test driver name.

### DIFF
--- a/src/libratbag-private.h
+++ b/src/libratbag-private.h
@@ -585,14 +585,14 @@ ratbag_led_set_mode_capability(struct ratbag_led *led,
 }
 
 /* list of all supported drivers */
-struct ratbag_driver etekcity_driver;
-struct ratbag_driver hidpp20_driver;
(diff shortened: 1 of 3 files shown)


FFY00:
Did you go through the headers to see if there are more cases of this issue?

whot:
I didn't, I figured if there are, gcc will scream at me in the future again :)

PR #908, opened 2020-01-29
data: add support for Logitech G815 keyboard

tfylling:


FFY00:
Do the leds change the moment after you apply changes? Here they don't, I need to press one of the M keys. Maybe I am in a funny firmware.

tfylling:
> Do the leds change the moment after you apply changes? Here they don't, I need to press one of the M keys. Maybe I am in a funny firmware.

They do not change for me either until I've changed to a different profile and back.

tfylling:
> Looks good to me. Did you test it?
> Also, you should squash the commits into one (they will be squashed when merging anyway, but just for the future).

Noted on the squashing.

It works as well as can be expected. Obviously it doesn't add support for individual key leds or anything like that, but the keboard is recognized and configurable with ratbagctl, and I can turn that horrible default rainbow cycling off!

Also, of course, it's not fully working in piper, as it's just showing the 404-rat and button 0 and 1 (G1 and G2) only have options for left or right handed, but I guess that's beyond the scope of both libratbag and this PR?

stephanlachnit:
Piper doesn't know that we added keyboards to libratbag, so this is expected.

FFY00:
> They do not change for me either until I've changed to a different profile and back. I'm guessing piper and ratbagctl aren't or can't send commands to reload the profile after committing it to the keyboard?

The daemon can send that command, however I tried it and it still does not trigger the LED change. I tried several different things but none of them seem to update the LEDs. I will query Logitech about this.

>It works as well as can be expected.

I also tried it and I can change the LEDs and remap the G keys.

> Obviously it doesn't add support for individual key leds or anything like that, but the keboard is recognized and configurable with ratbagctl, and I can turn that horrible default rainbow cycling off! Only thing I've found so far that's not working is detailed in issue #909 but I can't see that that's anything to do with the device file.

Yes, out of scope for now.

PR #910, opened 2020-01-29
add LED for Logitech g9x

heimer16:


stephanlachnit:
If we change the g9x anyway, we could also merge the g9x and the g9x CoD device files, they are the same anyway (except for the usb id). If you want you can do it, else I would add it.

stephanlachnit:
See #913.

PR #912, opened 2020-01-30
update and merge logitech g9x

heimer16:
Merged G9x with Call of Duty edition.
Updated with DPI LED
also updated min DPI (200) and increments (50), although I can't confirm if CoD edition is the same.
updated support links


stephanlachnit:
In case you forgot something, you can amend something to a commit in git, or run `git reset HEAD^`, or rebase your branch. If you search online you will surely find simple tutorials. tldr: you don't need to create a new pull request in case you want to change something.

PR #919, opened 2020-02-08
Misc fixes for HID++ 2.0 keyboards

### LINKED ISSUE #903: Logitech G815 keyboard support.
Today I bought a new logitech G815 keyboard from the store. Since I've had a lot of luck so far with my linux installation I didn't think anything of it until I tried downloading Logitech G hub to set the colors which of course is only supported on Windows and Mac.
After some googeling I found libratbag which I tried to use but sadly does not seem to support my keyboard.

I would love it if libratbag could support my logitech G815 keyboard. 
Ive tried to use the g810-led package in hopes it would maybe support the 815 as well but could not get it working. 

I of course understand this is a community effort and would love to help with proggraming / debugging / testing for this feature. My C is quite rusty but with some guidance I think it could be fun.

FFY00:
Now we should be ready for a new release. This fixes the error described in #903 and makes sure we don't overwrite the effects saved in the keyboard if they're not supported.

### DIFF
--- a/src/driver-hidpp20.c
+++ b/src/driver-hidpp20.c
@@ -1482,7 +1482,6 @@ hidpp20drv_commit(struct ratbag_device *device)
 	struct ratbag_led *led;
 	struct ratbag_resolution *resolution;
 	int rc;
-	unsigned int dpi_index = 0;
 
 	list_for_each(profile, &device->profiles, link) {
 		if (!profile->dirty)
@@ -1544,10 +1543,9 @@ hidpp20drv_commit(struct ratbag_device *device)
 			if (profile->is_active) {
 				ratbag_profile_for_each_resolution(profile, resolution) {
 					if (resolution->is_active)
-						dpi_index = resolution->index;
+						hidpp20_onboard_profiles_set_current_dpi_index(drv_data->dev,
+											       resolution->index);
 				}
-				hidpp20_onboard_profiles_set_current_dpi_index(drv_data->dev,
-									       dpi_index);
 			}
 		}
 	}
--- a/src/hidpp20.c
+++ b/src/hidpp20.c
@@ -2664,12 +2664,20 @@ hidpp20_onboard_profiles_read_led(struct hidpp20_led *led,
 		if (brightness == 0)
 			brightness = 100;
 		break;
+	case HIDPP20_LED_STARLIGHT:
+		led->color = internal_led.effect.starlight.color_sky;
+		led->extra_color = internal_led.effect.starlight.color_star;
+		break;
 	case HIDPP20_LED_BREATHING:
 		period = hidpp_be_u16_to_cpu(internal_led.effect.breath.period_or_speed);
 		brightness = internal_led.effect.breath.intensity;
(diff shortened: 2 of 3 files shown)


FFY00:
@bentiss @whot, @stephanlachnit wanted me to release this weekend so that the new release makes its way into Ubuntu 18.04 LTS so if I could get an ACK on this patch soon it would be great.

src/driver-hidpp20.c
```diff
@@ -1546,8 +1546,9 @@ hidpp20drv_commit(struct ratbag_device *device)
 					if (resolution->is_active)
 						dpi_index = resolution->index;
 				}
-				hidpp20_onboard_profiles_set_current_dpi_index(drv_data->dev,
-									       dpi_index);
+				if (dpi_index != 0xff)
```
stephanlachnit: Couldn't we do instead of this (in my opinion a little bit hacky) solution do simply:

```C
if (resolution->is_active)
	hidpp20_onboard_profiles_set_current_dpi_index(drv_data->dev,
	                                               resolution->index);
```

And even remove the variable.

FFY00: I think we had that for previously but for some reason removed it. I am not home right now, could you look at the git blame?

stephanlachnit: True, commit is 563550c473ff29849560007eec6808f2995152c2. If I get it right the problem was that `dpi_index` was used before in the commit loop and that could be a problem depending on the order of the profiles somehow. However, `dpi_index` is not used before anymore, so I think it should be fine to remove it.

FFY00: Thanks for looking into it, I am going to change it when I get home.

src/hidpp20.c
```diff
@@ -2572,6 +2572,7 @@ hidpp20_buttons_to_cpu(struct hidpp20_device *device,
 			break;
 		case HIDPP20_BUTTON_SPECIAL:
 			button->special.special = b->special.special;
+			button->special.profile = b->special.profile;
```
whot: Can you expand the commit message a bit here? I think I know what it does but not having to guess would be preferable ;)

Remember not everyone has the specs or remembers what they say, so a good commit message can make having to look them up unnecessary.

src/driver-hidpp20.c
```diff
@@ -1482,7 +1482,7 @@ hidpp20drv_commit(struct ratbag_device *device)
 	struct ratbag_led *led;
 	struct ratbag_resolution *resolution;
 	int rc;
-	unsigned int dpi_index = 0;
+	unsigned int dpi_index = 0xff;
```
whot: AIUI the `dpi_index` is only used in that inner loop, so best to move the declaration to there too so it becomes better contained. And then the `0xff` doesn't look quite as confusing as it does now.

src/driver-hidpp20.c
whot: agree with @stephanlachnit, whatever the code was before the current flow allows it to be called on `is_active` which makes the whole thing less confusing, IMO.

PR #920, opened 2020-02-08
data: add the Logitech G Powerplay

FFY00:


FFY00:
Yes, but the kernel bits are not upstreamed yet.

See https://lkml.org/lkml/2020/1/15/1042

PR #945, opened 2020-04-04
update gpw profile for chinese version

j0x7c4:
it fails to recognize my gpw at the very beginning.
Maybe it has different device id in different country version. 
Using lsusb to check my gpw device id


FFY00:
That is the ID of the receiver, not the device. Due to some changes to the kernel driver you need at least Linux 5.5 for the device to be recognized.

PR #946, opened 2020-04-05
Added G304 device

adarshsingh87:


FFY00:
That ID is the receiver, not the device. Please make sure you are running at least Linux 5.5 and see if it's working, if it not we will need to add a device file for the device.

PR #947, opened 2020-04-06
fixed G304

adarshsingh87:


FFY00:
You also have to have an entry in the `meson.build` file.

FFY00:
Looks like the G304 is the same as the G305. We already have a device file for this ID, is it not showing in your machine?

adarshsingh87:
it shows up as g305 rather than g304

FFY00:
The devices have the same ID, and seem the same. We can only have one name.

PR #960, opened 2020-04-16
ci: add arch

FFY00:


FFY00:
@whot turns out the issue was really simple, `/etc/dbus-1/system.d` did not exist. I blame bad debug tooling! The local circleci runner now works fine, so I was able to track this down.

stephanlachnit:
The config looks fine, but do we really want to add arch linux? Since it's a rolling release distro it could potentially break at any point, for the same reason we removed Fedora Rawhide in  769289fe01b1bccb67caa40033a0b32652dffcc6.

FFY00:
Then we know in advance if something breaks. Arch Linux itself is pretty stable (not sure if it's the same with rawhide?), it is just that we use the latest upstream version. We don't package unstable releases/pre-releases. We also keep as close as the upstream as possible, this means that we don't add any extra patches, like Debian does, not sure if it's the same for Fedora. From my understanding Fedora rawhide is a development version of Fedora, this means the packages themselves are not stable, contrary to Arch Linux. If I am wrong please let me know.

TLDR: If something breaks for us with Arch Linux, almost all the time it's the upstream's fault (API changes, etc.). That means we will have to deal with it sooner or later, I prefer sooner.

whot:
@stephanlachnit fwiw, I've been running libinput builds against arch for ages now and it's always been stable with I think only one build error in two years or so. so the distro is in a good shape.

FFY00:
But the cache persists between jobs. Arch Linux requires a system upgrade when installing new packages (partial upgrades are not supported, this means if we update the repo databases we must do a system upgrade before/when we install new packages) and having the packages cached saved us from re-downloading them, which saves quite a bit of time.

whot:
between jobs or between pipelines (or "workflows") as circle CI calls them. It needs to be the latter to work but [the circle CI docs](https://circleci.com/docs/2.0/configuration-reference/#persist_to_workspace) suggest it's the wrong tool for the job here. Looks like we need [caching](https://circleci.com/docs/2.0/caching/) instead (technically for the fedora bits too).

FFY00:
It turns out using `caching` simplifies things, as we don't need a separate prep_cache step. We already have the cache from the previous job, we install the packages and update the cache.

FFY00:
Don't merge yet.

FFY00:
Okay, great! Everything is working fine as we can see here: https://circleci.com/gh/libratbag/libratbag/4992

Issue #966, opened 2020-04-25
Mouse without support for profile (e.g. Logitech G5) crashes ratbagd


danlehmann:
**Information**
- `ratbagd` version (`ratbagd --version`): 0.13
- Distribution: Ubuntu 20.04
- Kernel version (ex. `uname -srmo`): `Linux 5.4.0-26-generic x86_64 GNU/Linux`

**Describe the bug**
Using a Logitech G5 crashes ratbagd.

I got the source and debugged. There are three issues that I can see:
1. In cases where ratbag doesn't initialize correctly, a half-initialized object is free'd, crashing the daemon. There's already a pull request open that fixes the issue for me: https://github.com/libratbag/libratbag/pull/954 . I pulled this patch and it fixes the crash for me, so I hope this will be merged soon. However, even with it in the mouse won't be enumerated properly.

I isolated two more issues and have a WIP patch to address those:
2. There's an error enumerating profiles. I found some documentation on this (https://github.com/cvuchener/g500/blob/master/doc/commands/0f-profile.md) and it seems that the G5 doesn't support this query. I made hidpp10_get_current_profile more resilient and it then keeps going.
3. With that in, I am hitting another issue: querying resolutions. In the ".device" file for the mouse I see a DpiList specified, but it is never read unless I also specify "ProfileType=G500".

A patch that addresses two issues is here. However, I am new to libratbag, so I am sure there things I am not aware of. Please let me know what needs to be done so that it can be merged (assuming I'm on the right track).
https://github.com/danlehmann/libratbag/commit/2e84dcc57f6c7eac268bddc375d1793e8f4beefc

The DpiList fix in the patch also seems like it would affect other mice. If a DpiList is specified in the .device file, shouldn't it always be read, even if the profile is not G500?

**To Reproduce**
1. Plugin a G5
2. Launch ratbagd. Notice that the daemon crashes.
3. Pull my patch above
4. Mouse is enumerated successfully.

**Logs**
I dumped the log while my patch is applied. Otherwise, due to the error crash there isn't much in the log.

```
$ ratbagd --verbose=raw
Initializing libratbag
ratbag debug: New device: Logitech K830
ratbag debug: Using data directory '/usr/local/share/libratbag'
ratbag debug: No data file found for 046d:4032
ratbag debug: New device: Logitech MX Master 3
ratbag debug: Using data directory '/usr/local/share/libratbag'
ratbag debug: device assigned driver hidpp20
[...]

PR #967, opened 2020-04-26
Fix enumeration of Logitech G5 (2007)

danlehmann:
Before this patch, ratbagd would abort initialization of the device because
no profiles are returned.

Once that was handled more gracefully, ratbagd would fail because it doesn't
read the DpiList from its own config file, as no ProfileType is specified.
Specifying G500 as the closest match.


danlehmann:
This is a proposed fix for https://github.com/libratbag/libratbag/issues/966

I added an explanation on the issue.

FFY00:
I am really sorry for the delay!

PR #980, opened 2020-05-11
meson: make finding python more robust

dcbaker:
This code assumes that there is a python3 or python3-embed pkg-config
file on the system, which may or may not be true. On Gentoo for example,
only python-X.Y pkg-config files are available and this will fail to
build.

This makes use of the new py_installation.dependency(embed : true) if
possible, and falls back to trying to find a python that matches the
version of python found by the python3 or python module.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -538,12 +538,28 @@ configure_file(input : 'dbus/org.freedesktop.ratbag1.conf.in',
 
 #### tools ####
 
+# This is a special idiomatic construct in meson to get a fast dependency that
+# doesn't exist and isn't logged as "not found".
+dep_python3 = dependency('', required : false)
 if meson.version().version_compare('<0.48.0')
   python = import('python3')
   py3 = python.find_python()
 else
   pymod = import('python')
   py3 = pymod.find_installation()
+  if meson.version().version_compare('>= 0.53.0')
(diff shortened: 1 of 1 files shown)


meson.build
```diff
@@ -538,12 +538,31 @@ configure_file(input : 'dbus/org.freedesktop.ratbag1.conf.in',
 
 #### tools ####
 
+dep_python3 = dependency('', required : false)
```
FFY00, line 541: `''` is this a typo?

dcbaker: no, this is the idiomatic "not found" dependency. This creates a dependency object that will never be found, but allows you to make sure that dep_python3 is always a dependency object

dcbaker: I've added a comment explaining this, and also opened an upstream documentation addition about this: https://github.com/mesonbuild/meson/pull/7122

FFY00: Okay, I still need to test but this looks fine. Please fix the indentation, you are using tabs while we use 2 spaces.

dcbaker:
I've simplified the logic a bit, and added a comment about the not-found dependency.

meson.build
dcbaker: The latest version should be using spaces. I got confused because the meson.build has both tabs and spaces in it.

PR #989, opened 2020-05-20
ci: arch: install git

FFY00:
CircleCI complains:

Either git or ssh (required by git to clone through SSH) is not installed in the
image. Falling back to CircleCI's native git client but the behavior may be
different from official git. If this is an issue, please use an image that has
official git and ssh installed.

Using arch's git client would be better.


FFY00:
Weird, it still complains.

FFY00:
Nevermind, git was already installed, CircleCI is being weird.

PR #1002, opened 2020-06-07
ci: bump Ubuntu to 20.04

stephanlachnit:
20.04 is an LTS release, so we should use it, and hopefully we don't need to bump it until 22.04 LTS^^

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -156,10 +156,7 @@ ubuntu_settings: &ubuntu_settings
     - run:
         name: Install prerequisites
         command: |
-          apt-get update
-          apt-get install -y software-properties-common
-          apt-get remove -y libnss-systemd
-          add-apt-repository universe
+          export DEBIAN_FRONTEND=noninteractive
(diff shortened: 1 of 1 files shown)


stephanlachnit:
Apparently it's still buggy (already tried it once it was released).

.circleci/config.yml
```diff
@@ -156,10 +156,7 @@ ubuntu_settings: &ubuntu_settings
     - run:
         name: Install prerequisites
         command: |
-          apt-get update
-          apt-get install -y software-properties-common
-          apt-get remove -y libnss-systemd
```
whot, line 161: I'd prefer those removed as a separate commit, or at least specifically mentioned in the commit message as no longer needed. Took us quite some time to figure it out in the first place.

stephanlachnit:
> look good, thanks, but please do mention the libnss-system explicity as no longer needed (with a ref to the sha that introduced it) so we can easily trace this back if we get the hangs again.

Good point, I updated the description.

PR #1016, opened 2020-06-21
Add USB device ID + fix device model no.

robfrancken:


data/devices/logitech-g102-g203.device
```diff
@@ -1,6 +1,6 @@
 # G102 and G203 (USB)
 [Device]
-Name=Logitech Gaming Mouse G102/G103
-DeviceMatch=usb:046d:c084
+Name=Logitech Gaming Mouse G102/G203
```
FFY00, line 3: Please keep the `G103`, it's a different variant only sold in China IIRC.

```
G102/G103/G203
```

FFY00:
The arch CI seems to be broken, it doesn't have anything to do with your change. I'll look into it later.

stephanlachnit:
> The arch CI seems to be broken, it doesn't have anything to do with your change. I'll look into it later.

Nothing to do with Arch, happens to all distros on CircleCI. It's this error:
```
gi.repository.GLib.GError: g-io-error-quark: Could not connect: Connection refused (39)
```
Just restarted the Arch CI, it usually doesn't happen twice in a row.

stephanlachnit:
This is already done in https://github.com/libratbag/libratbag/commit/02226d04280964f69fe5b58bc88232ec697abc70? The difference is the name change, which would need to be resolved @robfrancken.

FFY00:
Yeah, the name can be added.

robfrancken:
I thought the file name contained a typo, since the device file is named logitech-g102-g**2**03.device, but in the file the name was specified as G102/G**1**03.

stephanlachnit:
Already addressed in https://github.com/libratbag/libratbag/commit/1e72ebaa7ba848be8685cd14c70ba5d4e7f489b8

Issue #1007, opened 2020-06-10
Glorious Model O support


aindriu80:
Can you please add support for the Glorious Model O mouse

Glorious Model O Website:
https://www.pcgamingrace.com/products/glorious-model-o-black

aindriu80:
oh ok.  I have it in front of me.  What needs to be done with it?

stephanlachnit:
> oh ok. I have it in front of me. What needs to be done with it?

You need to write a driver for it. Here are some things to help you get started:
https://github.com/libratbag/libratbag/wiki/Adding-a-device
https://www.youtube.com/watch?v=PSaRhC2hCO0
https://fosdem.org/2020/schedule/event/libratbag/

enkore:
I reverse engineered the protocol a while ago, some C code is here: https://github.com/enkore/gloriousctl

I'm probably maybe going to add it to libratbag.

stephanlachnit:
> I'm probably maybe going to add it to libratbag.

That would be pretty cool! If you need help somewhere, let me know it.

staticssleever668:
Issue is fixed and should be closed.

PR #1020, opened 2020-06-25
Add sinowealth driver for Glorious Model O/D

enkore:
RGB color picker doesn't quite work right, I suspect I slipped by a few bytes while writing the C struct based on my notes. Everything else just works as far as I can tell.

Implements #1007

### DIFF
--- a/data/devices/glorious-model-d.device
+++ b/data/devices/glorious-model-d.device
@@ -1,5 +1,5 @@
 [Device]
-Name=SINOWEALTH Wired Gaming Mouse
+Name=Glorious Model D
 DeviceMatch=usb:258a:0033
 Driver=sinowealth
 
--- a/data/devices/glorious-model-d.device
+++ b/data/devices/glorious-model-d.device
@@ -1,0 +1,5 @@
+[Device]
+Name=SINOWEALTH Wired Gaming Mouse
+DeviceMatch=usb:258a:0033
+Driver=sinowealth
+
--- a/data/devices/glorious-model-o.device
+++ b/data/devices/glorious-model-o.device
@@ -1,0 +1,5 @@
+[Device]
+Name=SINOWEALTH Wired Gaming Mouse
+DeviceMatch=usb:258a:0036
+Driver=sinowealth
+
--- a/data/devices/glorious-model-o.device
+++ b/data/devices/glorious-model-o.device
@@ -1,5 +1,5 @@
 [Device]
-Name=SINOWEALTH Wired Gaming Mouse
+Name=Glorious Model O
 DeviceMatch=usb:258a:0036
 Driver=sinowealth
 
--- a/src/driver-sinowealth.c
+++ b/src/driver-sinowealth.c
@@ -186,8 +186,15 @@ sinowealth_rgb_mode_to_brightness(uint8_t mode)
 	return max((mode >> 4) * 64, 255);
 }
 
+static uint8_t
+sinowealth_brightness_to_rgb_mode(int brightness)
+{
+	/* convert 0-255 to 0-4 in the upper nibble of mode */
+	return ((brightness + 1) / 64) << 4;
+}
+
 static int
-sinowealth_rgb_mode_to_speed(uint8_t mode)
+sinowealth_rgb_mode_to_duration(uint8_t mode)
 {
 	switch (mode & SINOWEALTH_RGB_SPEED_BITS) {
 	case 0: return 10000; /* static: does not translate to duration */
@@ -199,17 +206,32 @@ sinowealth_rgb_mode_to_speed(uint8_t mode)
 }
 
 static uint8_t
-sinowealth_to_rgb_mode(int brightness, int speed_ms)
+sinowealth_duration_to_rgb_mode(int duration_ms)
 {
-	/* convert 0-255 to 0-4 in the upper nibble of mode */
-	uint8_t mode = ((brightness + 1) / 64) << 4;
-	if (speed_ms <= 500) {
+	uint8_t mode = 0;
+	if (duration_ms <= 500) {
 		mode |= 3;
-	} else if (speed_ms <= 1000) {
+	} else if (duration_ms <= 1000) {
 		mode |= 2;
 	} else {
 		mode |= 1;
 	}
+	return mode;
+}
+
+static void
+sinowealth_set_led_from_rgb_mode(struct ratbag_led *led, uint8_t mode)
+{
+	led->brightness = sinowealth_rgb_mode_to_brightness(mode);
+	led->ms = sinowealth_rgb_mode_to_duration(mode);
+}
+
+static uint8_t
+sinowealth_led_to_rgb_mode(const struct ratbag_led *led)
+{
+	uint8_t mode;
+	mode = sinowealth_brightness_to_rgb_mode(led->brightness);
+	mode |= sinowealth_duration_to_rgb_mode(led->ms);
 	return mode;
 }
 
@@ -278,14 +300,12 @@ sinowealth_read_profile(struct ratbag_profile *profile)
 	case RGB_RAVE:
 	case RGB_WAVE:
 		led->mode = RATBAG_LED_CYCLE;
-		led->brightness = sinowealth_rgb_mode_to_brightness(config->glorious_mode);
-		led->ms = sinowealth_rgb_mode_to_speed(config->glorious_mode);
+		sinowealth_set_led_from_rgb_mode(led, config->glorious_mode);
 		break;
 	case RGB_BREATHING1:
 		led->mode = RATBAG_LED_BREATHING;
 		led->color = sinowealth_rbg_to_color(config->breathing1_color);
-		led->brightness = sinowealth_rgb_mode_to_brightness(config->breathing1_mode);
-		led->ms = sinowealth_rgb_mode_to_speed(config->breathing1_mode);
+		sinowealth_set_led_from_rgb_mode(led, config->breathing1_mode);
 		break;
 	}
 	ratbag_led_unref(led);
@@ -439,12 +459,12 @@ sinowealth_commit(struct ratbag_device *device)
 		break;
 	case RATBAG_LED_CYCLE:
 		config->rgb_effect = RGB_GLORIOUS;
-		config->glorious_mode = sinowealth_to_rgb_mode(led->brightness, led->ms);
+		config->glorious_mode = sinowealth_led_to_rgb_mode(led);
 		break;
 	case RATBAG_LED_BREATHING:
 		config->rgb_effect = RGB_BREATHING1;
 		config->breathing1_color = sinowealth_color_to_rbg(led->color);
-		config->breathing1_mode = sinowealth_to_rgb_mode(led->brightness, led->ms);
+		config->breathing1_mode = sinowealth_led_to_rgb_mode(led);
 		break;
 	}
 	ratbag_led_unref(led);
--- a/src/driver-sinowealth.c
+++ b/src/driver-sinowealth.c
@@ -29,6 +29,7 @@
 #define SINOWEALTH_CMD_FIRMWARE_VERSION 0x1
 #define SINOWEALTH_CMD_GET_CONFIG 0x11
 #define SINOWEALTH_CONFIG_SIZE 520
+#define SINOWEALTH_CONFIG_SIZE_USED 131
 
 #define SINOWEALTH_XY_INDEPENDENT 0x80
 
@@ -43,39 +44,39 @@
 #define SINOWEALTH_RGB_BRIGHTNESS_BITS 0xF0
 #define SINOWEALTH_RGB_SPEED_BITS 0x0F
 
-struct RGB8 {
+struct sinowealth_rgb8 {
 	uint8_t r, g, b;
 } __attribute__((packed));
 
-_Static_assert(sizeof(struct RGB8) == 3, "Invalid size");
+_Static_assert(sizeof(struct sinowealth_rgb8) == 3, "Invalid size");
 
-struct RBG8 {
+struct sinowealth_rbg8 {
 	uint8_t r, b, g;
 } __attribute__((packed));
 
-_Static_assert(sizeof(struct RBG8) == 3, "Invalid size");
+_Static_assert(sizeof(struct sinowealth_rbg8) == 3, "Invalid size");
 
 enum rgb_effect {
 	RGB_OFF = 0,
 	RGB_GLORIOUS = 0x1,   /* unicorn mode */
 	RGB_SINGLE = 0x2,     /* single constant color */
-	RGB_BREATHING = 0x5,  /* RGB breathing */
-	RGB_BREATHING7 = 0x3, /* breathing with seven colors */
+	RGB_BREATHING = 0x5,  /* Full RGB breathing */
+	RGB_BREATHING7 = 0x3, /* breathing with seven user-defined colors */
 	RGB_BREATHING1 = 0xa, /* single color breathing */
-	RGB_TAIL = 0x4,       /* idk what this is supposed to be */
-	RGB_RAVE = 0x7,       /* ig */
+	RGB_TAIL = 0x4,
+	RGB_RAVE = 0x7,
 	RGB_WAVE = 0x9
 };
 
 struct sinowealth_config_report {
 	uint8_t report_id; /* SINOWEALTH_REPORT_ID_CONFIG */
 	uint8_t command_id;
-	uint8_t unk1;
+	uint8_t unknown1;
 	uint8_t config_write;
 	/* always 0 when config is read from device,
 	 * has to be 0x7b when writing config to device
 	 */
-	uint8_t unk2[6];
+	uint8_t unknown2[6];
 	uint8_t config;
 	/* 0x80 - SINOWEALTH_XY_INDEPENDENT */
 	uint8_t dpi_count:4;
@@ -91,7 +92,7 @@ struct sinowealth_config_report {
 	 * If XY are identical, dpi[0-6] contain the sensitivities,
 	 * while in XY independent mode each entry takes two chars for X and Y.
 	 */
-	struct RGB8 dpi_color[8];
+	struct sinowealth_rgb8 dpi_color[8];
 	uint8_t rgb_effect;
 	/* see enum rgb_effect */
 	uint8_t glorious_mode;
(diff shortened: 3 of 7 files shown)


src/driver-sinowealth.c
```diff
@@ -0,0 +1,400 @@
...
+	char breathing1_mode;
+	/* 0x1/2/3 - speed */
+	RGB8 breathing1_color;
+
+	char unk4;
+	char lift_off_distance;
+	/* 0x1 - 2 mm
+	 * 0x2 - 3 mm
+	 */
+} __attribute__((packed));
```
FFY00, line 129: Can you add a static assert for the size? This should make sure we don't compile with compilers which do not pack properly.

Example: https://github.com/libratbag/libratbag/blob/02575a90083b6cc3d5dd1a2d510acc34c4c24443/src/hidpp20.h#L904

src/driver-sinowealth.c
```diff
@@ -0,0 +1,400 @@
...
+	 * but the glorious software only exposes six
+	 */
+	uint8_t dpi[16];
+	/* DPI/CPI is encoded in the way the PMW3360 sensor accepts it
+	 * value = (DPI - 100) / 100
+	 * If XY are identical, dpi[0-6] contain the sensitivities,
+	 * while in XY independent mode each entry takes two chars for X and Y.
+	 */
+	RGB8 dpi_color[8];
+
```
FFY00, line 84: Can you keep the style consistent here? Sometimes you add a newline between fields, sometimes you don't.

src/driver-sinowealth.c
```diff
@@ -0,0 +1,400 @@
...
+	struct sinowealth_config_report *config = &drv_data->config;
+	struct ratbag_resolution *resolution;
+	struct ratbag_led *led;
+	int num_dpis = (SINOWEALTH_DPI_MAX - SINOWEALTH_DPI_MIN) / SINOWEALTH_DPI_STEP + 1 + 1;
+	unsigned int dpis[num_dpis];
+	unsigned int hz = 1000; /* TODO */
+	int rc;
+	uint8_t *data;
+
+	uint8_t cmd[6] = {SINOWEALTH_REPORT_ID_CMD, SINOWEALTH_CMD_GET_CONFIG};
```
FFY00, line 175: I would expand this in multiple lines, it makes it easier to read :smile:.

src/driver-sinowealth.c
```diff
@@ -0,0 +1,400 @@
...
+	rc = ratbag_hidraw_get_feature_report(device, SINOWEALTH_REPORT_ID_CONFIG,
+					      data, SINOWEALTH_CONFIG_SIZE);
+	/* The GET_FEATURE report length has to be 520, but the actual data returned is less */
+	if (rc < sizeof(*config)) {
+		log_error(device->ratbag, "Could not read device configuration: %d\n", rc);
+		return -1;
+	}
+
+	memcpy(config, data, sizeof(*config));
+	free(data);
```
FFY00, line 192: Can you make `sinowealth_config_report` a union and avoid `memcpy` and friends?

Example: 
https://github.com/libratbag/libratbag/blob/02575a90083b6cc3d5dd1a2d510acc34c4c24443/src/hidpp20.h#L44
https://github.com/FFY00/wireshark/blob/385c4da96a6110baac5c6c6f512123837568945e/epan/dissectors/packet-usb-hid.c#L180

src/driver-sinowealth.c
```diff
@@ -0,0 +1,400 @@
...
+	}
+
+	/* Body lighting */
+	led = ratbag_profile_get_led(profile, 0);
+	led->type = RATBAG_LED_TYPE_SIDE;
+	led->colordepth = RATBAG_LED_COLORDEPTH_RGB_888;
+	ratbag_led_set_mode_capability(led, RATBAG_LED_OFF);
+	ratbag_led_set_mode_capability(led, RATBAG_LED_ON);
+	ratbag_led_set_mode_capability(led, RATBAG_LED_CYCLE);
+	ratbag_led_set_mode_capability(led, RATBAG_LED_BREATHING);
```
FFY00, line 230: I don't think this should be done in `read_profile`, it should probably be in `probe`.

Or do you have dynamic capabilities? I don't think piper is ready to handle that.

enkore: This function only gets called from probe. I understood the comments about ratbag_driver to mean that a new driver is only supposed to implement probe/commit/remove, since everything else is deprecated.

FFY00: Right, sorry. I still think it would be maybe better to have the capability setting logic unbundled from `read_profile` but that's just a nitpick.

enkore: Ah, ok. I split read_profile up and moved all the invariant stuff to a separate function (sinowealth_init_profile).

src/driver-sinowealth.c
enkore: I changed it a little bit, the struct now just has all the padding required to make it work. No memcpy. The firmware in these devices is a little bit weird; it requires the get/set feature reports to carry 520 bytes (including 1 byte for the report ID), but only 131 are actually used (and sent by the device). I assume this was originally just the entire 512 byte EEPROM (plus a header) of the controller they used, but as features got added they split it up into segments or something like that.

src/driver-sinowealth.c
enkore: Done

src/driver-sinowealth.c
```diff
@@ -0,0 +1,475 @@
...
+	uint8_t r, g, b;
+} RGB8;
+
+_Static_assert(sizeof(RGB8) == 3, "Invalid size");
+
+typedef struct __attribute__((packed)) {
+	uint8_t r, b, g;
+} RBG8;
+
+_Static_assert(sizeof(RBG8) == 3, "Invalid size");
```
FFY00, line 56: Sorry for being picky :smile:. We don't really typedef structures.

Example: https://github.com/libratbag/libratbag/blob/02575a90083b6cc3d5dd1a2d510acc34c4c24443/src/hidpp20.h#L891

enkore: No problemo, done :grin:

FFY00:
@whot can you have a look? Just to make sure everything is in order.

data/devices/glorious-model-d.device
```diff
@@ -0,0 +1,5 @@
+[Device]
+Name=SINOWEALTH Wired Gaming Mouse
```
stephanlachnit, line 2: Why don't you set the name to `Glorious Model D` here?

data/devices/glorious-model-o.device
```diff
@@ -0,0 +1,5 @@
+[Device]
+Name=SINOWEALTH Wired Gaming Mouse
```
stephanlachnit, line 2: same here

src/driver-sinowealth.c
```diff
@@ -0,0 +1,475 @@
+/*
+ * Copyright © 2020 Marian Beermann
```
stephanlachnit, line 2: Not sure if I like it but I guess it's fine (IMHO if you contribute of a FOSS project you might as well give them the copyright of your code, but I understand that not everyone sees it that way).

src/driver-sinowealth.c
```diff
@@ -0,0 +1,475 @@
...
+} __attribute__((packed));
+
+_Static_assert(sizeof(struct RBG8) == 3, "Invalid size");
+
+enum rgb_effect {
+	RGB_OFF = 0,
+	RGB_GLORIOUS = 0x1,   /* unicorn mode */
+	RGB_SINGLE = 0x2,     /* single constant color */
+	RGB_BREATHING = 0x5,  /* RGB breathing */
+	RGB_BREATHING7 = 0x3, /* breathing with seven colors */
```
stephanlachnit, line 63: What exactly is the difference between `RGB Breathing` and `RGB_BREATHING7`? Can you maybe extend the description a bit?

src/driver-sinowealth.c
```diff
@@ -0,0 +1,475 @@
...
+
+struct RBG8 {
+	uint8_t r, b, g;
+} __attribute__((packed));
+
+_Static_assert(sizeof(struct RBG8) == 3, "Invalid size");
+
+enum rgb_effect {
+	RGB_OFF = 0,
+	RGB_GLORIOUS = 0x1,   /* unicorn mode */
```
stephanlachnit, line 60: Maybe a better description here. If find neither `RGB_GLORIOUS` nor `unicorn mode` very descriptive.

src/driver-sinowealth.c
```diff
@@ -0,0 +1,475 @@
...
+	uint8_t unk2[6];
+	uint8_t config;
+	/* 0x80 - SINOWEALTH_XY_INDEPENDENT */
+	uint8_t dpi_count:4;
+	uint8_t active_dpi:4;
+	uint8_t dpi_enabled;
+	/* bit set: disabled, unset: enabled
+	 * this structure has support for eight DPI slots,
+	 * but the glorious software only exposes six
+	 */
```
stephanlachnit, line 87: Cool that this driver supports this, maybe consider adding support for https://github.com/libratbag/libratbag/pull/870 once that is merged.

src/driver-sinowealth.c
```diff
@@ -0,0 +1,475 @@
...
+
+static void
+sinowealth_remove(struct ratbag_device *device)
+{
+	ratbag_close_hidraw(device);
+	free(ratbag_get_drv_data(device));
+}
+
+struct ratbag_driver sinowealth_driver = {
+	.name = "Sinowealth Gaming Mouse",
```
stephanlachnit, line 470: Maybe just leave it as `Sinowealth` to make it more consistent with the other drivers (usually it's the vendor / protocol name, only for very device specific drivers the name is in there).

src/driver-sinowealth.c
FFY00: We have no CLA, if they write it, it's their copyright, but they're licensing it under MIT.

src/driver-sinowealth.c
FFY00: All these different mice have different color modes, I don't see the necessity to describe them all, especially since we don't support them.

src/driver-sinowealth.c
stephanlachnit: > We have no CLA, if they write it, it's their copyright, but they're licensing it under MIT.

I know that it isn't necessary. I still think it would be nice, it doesn't matter if you install it from source, but for example in Debian you have to mention _every_ copyright notice that's in the code if you want it in the official archive. Depending on the project, that list can grow pretty quickly and it gets hard to manage it.

src/driver-sinowealth.c
```diff
@@ -0,0 +1,475 @@
...
+	 */
+	uint8_t breathing1_mode;
+	/* 0x1/2/3 - speed */
+	struct RBG8 breathing1_color;
+	uint8_t unk5;
+	uint8_t lift_off_distance;
+	/* 0x1 - 2 mm
+	 * 0x2 - 3 mm
+	 */
+	uint8_t padding[520 - 131];
```
whot, line 133: `SINOWEALTH_CONFIG_SIZE` instead of a literal 520 here.

src/driver-sinowealth.c
```diff
@@ -0,0 +1,475 @@
...
+static struct RBG8
+sinowealth_color_to_rbg(struct ratbag_color color)
+{
+	return (struct RBG8) {.r = color.red, .g = color.green, .b = color.blue};
+}
+
+static int
+sinowealth_rgb_mode_to_brightness(uint8_t mode)
+{
+	return max(mode >> 4 << 6, 255);
```
whot, line 182: extra parenthesis please here to avoid accidental precedence misunderstandings. Also, tbh, I prefer `/` over `>>` except where we're actually dealing with bits. The compiler will know what to do anyway and IMO it's a lot easier to understand a `x * 64` than a `x << 6`.

src/driver-sinowealth.c
```diff
@@ -0,0 +1,475 @@
...
+	case 2: return 1000;
+	case 3: return 500;
+	default: return 0;
+	}
+}
+
+static uint8_t
+sinowealth_to_rgb_mode(int brightness, int speed_ms)
+{
+	uint8_t mode = (brightness + 1) >> 6 << 4;
```
whot, line 200: you should have the inverse helper to `sinowealth_rgb_mode_to_brightness` for this instead of open-coding it here, nicer for symmetry.

src/driver-sinowealth.c
```diff
@@ -0,0 +1,475 @@
...
+	return 0;
+}
+
+static void
+sinowealth_init_profile(struct ratbag_device *device)
+{
+	struct ratbag_profile *profile;
+	struct ratbag_resolution *resolution;
+	struct ratbag_led *led;
+	int num_dpis = (SINOWEALTH_DPI_MAX - SINOWEALTH_DPI_MIN) / SINOWEALTH_DPI_STEP + 1 + 1;
```
whot, line 307: `+1 + 1`? I've paged most of ratbag out by now but this looks odd to me.

src/driver-sinowealth.c
```diff
@@ -0,0 +1,475 @@
...
+	struct sinowealth_data *drv_data = device->drv_data;
+	struct sinowealth_config_report *config = &drv_data->config;
+	struct ratbag_resolution *resolution;
+	struct ratbag_led *led;
+	int rc;
+
+	config->config &= ~SINOWEALTH_XY_INDEPENDENT;
+	ratbag_profile_for_each_resolution(profile, resolution) {
+		if (resolution->dpi_x != resolution->dpi_y && resolution->dpi_x && resolution->dpi_y) {
+			config->config &= SINOWEALTH_XY_INDEPENDENT;
```
whot, line 400: should this be `|=` ?

src/driver-sinowealth.c
```diff
@@ -0,0 +1,475 @@
...
+	ratbag_profile_for_each_resolution(profile, resolution) {
+		if (config->config & SINOWEALTH_XY_INDEPENDENT) {
+			config->dpi[resolution->index * 2] = sinowealth_dpi_to_raw(resolution->dpi_x);
+			config->dpi[resolution->index * 2 + 1] = sinowealth_dpi_to_raw(resolution->dpi_y);
+		} else {
+			config->dpi[resolution->index] = sinowealth_dpi_to_raw(resolution->dpi_x);
+		}
+		if (resolution->dpi_x && resolution->dpi_y) {
+			/* enable DPI step (dpi_enabled is inverted) */
+			config->dpi_enabled &= ~(1<<resolution->index);
```
whot, line 415: spaces around `<<` for consistency

whot: I think it may be easier to understand if you OR the bits first and then invert the lot after the loop

src/driver-sinowealth.c
```diff
@@ -0,0 +1,475 @@
...
+		led = ratbag_profile_get_led(profile, i);
+		config->dpi_color[i - 1] = sinowealth_color_to_raw(led->color);
+		ratbag_led_unref(led);
+	}
+
+	config->config_write = 0x7b; /* magic */
+
+	rc = ratbag_hidraw_set_feature_report(device, SINOWEALTH_REPORT_ID_CONFIG,
+					      (uint8_t*) config, SINOWEALTH_CONFIG_SIZE);
+	if(rc != SINOWEALTH_CONFIG_SIZE) {
```
whot, line 452: space after `if`

data/devices/glorious-model-d.device
enkore: That's their vendor/product string, see [lsusb](https://github.com/enkore/gloriousctl/blob/latest/lsusb.txt#L8). I assume that's what's meant by "kernel name" in the example.

src/driver-sinowealth.c
enkore: "Glorious mode" is what their manual/software calls it. It looks like a rainbow cycling around the mouse.

src/driver-sinowealth.c
enkore: RGB_BREATHING uses all RGB colors, RGB_BREATHING7 uses the seven colors from breathing7_colors, RGB_BREATHING1 uses one color from breathing1_color; I can add that yeah.

src/driver-sinowealth.c
enkore: "+ 1" because (max-min)/step counts like an half-open interval, but we want both min and max included, another "+ 1" because we have "0" as an extra special value to signal a DPI step being disabled -- I'll add an comment and make it "+2".

src/driver-sinowealth.c
enkore: Agreed

src/driver-sinowealth.c
enkore: Yes, good catch!

data/devices/glorious-model-d.device
FFY00: There is no need to use the same name as shown in lsusb. I prefer to have names that better describe the device.

src/driver-sinowealth.c
enkore: Found some more spots of this and fixed them as well.

src/driver-sinowealth.c
enkore: Clarified the code and added a comment

src/driver-sinowealth.c
enkore: I added another #define for SINOWEALTH_CONFIG_SIZE_USED which made the code in read_profile code clearer as well.

src/driver-sinowealth.c
```diff
@@ -0,0 +1,475 @@
...
+#define SINOWEALTH_DPI_MAX 12000
+#define SINOWEALTH_DPI_STEP 100
+
+/* other models might have up to eight */
+#define SINOWEALTH_NUM_DPIS 6
+
+#define SINOWEALTH_RGB_BRIGHTNESS_BITS 0xF0
+#define SINOWEALTH_RGB_SPEED_BITS 0x0F
+
+struct RGB8 {
```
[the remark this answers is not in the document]
enkore: Done, also renamed to struct sinowealth_rgb8 etc. to be more in-line with the general style.

src/driver-sinowealth.c
enkore: I don't really agree on this one, because the usage in the code isn't symmetric, the helpers are asymmetric as well (bundling/unbundling multiple fields into a single field). I suppose if these really need to be symmetric they could be defined in terms of `struct ratbag_led`.

enkore: Like so:

```c
static int
sinowealth_rgb_mode_to_brightness(uint8_t mode)
{
	/* brightness is 0-4 in the upper nibble of mode,
	 * while libratbag uses 0-255.
	 */
	return max((mode >> 4) * 64, 255);
}

static int
sinowealth_rgb_mode_to_speed(uint8_t mode)
{
	switch (mode & SINOWEALTH_RGB_SPEED_BITS) {
	case 0: return 10000; /* static: does not translate to duration */
	case 1: return 1500;
	case 2: return 1000;
	case 3: return 500;
	default: return 0;
	}
}

static void
sinowealth_set_led_from_rgb_mode(const struct ratbag_led *led, uint8_t mode)
{
	led->brightness = sinowealth_rgb_mode_to_brightness(mode);
	led->ms = sinowealth_rgb_mode_to_speed(mode);
}

static uint8_t
sinowealth_led_to_rgb_mode(const struct ratbag_led *led)
{
	/* convert 0-255 to 0-4 in the upper nibble of mode */
	uint8_t mode = ((led->brightness + 1) / 64) << 4;
	if (led->ms <= 500) {
		mode |= 3;
	} else if (led->ms <= 1000) {
		mode |= 2;
	} else {
		mode |= 1;
	}
	return mode;
}
```

enkore:
Good to merge?

src/driver-sinowealth.c
```diff
@@ -0,0 +1,506 @@
...
+	return (struct sinowealth_rbg8) {.r = color.red, .g = color.green, .b = color.blue};
+}
+
+static int
+sinowealth_rgb_mode_to_brightness(uint8_t mode)
+{
+	/* brightness is 0-4 in the upper nibble of mode,
+	 * while libratbag uses 0-255.
+	 */
+	return max((mode >> 4) * 64, 255);
```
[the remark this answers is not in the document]
enkore: I actually just realized now that I never really tested this conversion, because it is only used during probe. Fixed, rechecked, works now.

enkore:
I don't know why the ubuntu_20_04 build failed, CircleCI doesn't want to tell me without having write access to all of my repositories...

stephanlachnit:
> I don't know why the ubuntu_20_04 build failed, CircleCI doesn't want to tell me without having write access to all of my repositories...

Nothing to do with the code, it's due to some weird glib error that happens on all distros from time to time. I restarted it, should be fine now.

whot:
I was happy with it last time I reviewed it, yes.

FFY00:
LGTM :+1:

compul:
I have the model o wireless. I added the device to /usr/share/libratbag, with the pid adjusted accordingly. It's not being recognized. Anything I can do?

staticssleever668:
@compul, you should open an issue about that.

garrett-hopper:
@compul, I opened #1172

PR #1035, opened 2020-07-25
g300: use the correct step size

TurkeyTheGobbler:
"Fixed" the dpi steps of the g300 to 50 to allow for finer adjustment.


FFY00:
Are you sure this is supported by the device?

TurkeyTheGobbler:
I'm allowed to do this in the Logitech Gaming Software on windows and it
works so I don't see why not.

PR #1043, opened 2020-08-16
Add driver for Nubwo X7 Spectrum

### LINKED ISSUE #1042: Support for Nubwo X7 Spectrum
**Is your feature request related to a problem? Please describe.**
I want to configure my mouse's functionality directly from Linux.

**Describe the solution you'd like**
A driver for my mouse

**Describe alternatives you've considered**
Buy a supported mouse. The mouse I have is too valuable ($30) to just throw away though.

**Additional context**

So I've already figured out some of the protocol and also written a partial driver for it.

The pull request should be coming soon.

saengowp:
Nubwo X7 Spectrum is a sinowealth-based gaming mouse.
The `sinowealth-nubwo` driver included in this commit provides

- LED support (except for color modes that are not available in libratbag)
- DPI and reporting support

The driver doesn't suppoty
- button mapping/macros

fixes #1042

### DIFF
--- a/src/driver-sinowealth-nubwo.c
+++ b/src/driver-sinowealth-nubwo.c
@@ -1,0 +1,315 @@
+/*
+ * Copyright © 2020 Pipat Saengow
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a
+ * copy of this software and associated documentation files (the "Software"),
+ * to deal in the Software without restriction, including without limitation
+ * the rights to use, copy, modify, merge, publish, distribute, sublicense,
+ * and/or sell copies of the Software, and to permit persons to whom the
+ * Software is furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice (including the next
+ * paragraph) shall be included in all copies or substantial portions of the
+ * Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.  IN NO EVENT SHALL
+ * THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+ * FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+ * DEALINGS IN THE SOFTWARE.
+ */
+
+#include "libratbag-enums.h"
+#include "libratbag-hidraw.h"
+#include "libratbag-private.h"
+#include "libratbag-util.h"
+#include "shared-macro.h"
+#include <stdint.h>
+#include <string.h>
+
+#define SINOWEALTHNUBWO_PERF_CMD_REPORTID 0x02
+#define SINOWEALTHNUBWO_AESTHETIC_CMD_REPORTID 0x03
+#define SINOWEALTHNUBWO_GET_FIRMWARE_CMD_REPORTID 0x04
+
+#define SINOWEALTHNUBWO_GET_FIRMWARE_MSGSIZE 256
+#define SINOWEALTHNUBWO_GET_FIRMWARE_MSGOFFSET 48
+
+#define SINOWEALTHNUBWO_PERF_CMD_MSGSIZE 16
+
+// Actually more but I only implemented one
+#define SINOWEALTHNUBWO_NUM_PROFILES 1
+#define SINOWEALTHNUBWO_NUM_RESOLUTIONS 1
+// Actually 8 but I am not going to implement macros.
+#define SINOWEALTHNUBWO_NUM_BUTTONS 0
+#define SINOWEALTHNUBWO_NUM_LEDS 1
+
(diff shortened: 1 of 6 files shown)


enkore:
I hoped that "most" Sinowealth mice would use a similar protocol to what I found in the Glorious mice, but it looks like they made a bunch of entirely different protocols over time.

src/driver-sinowealth-nubwo.c
```diff
@@ -0,0 +1,315 @@
...
+
+//Magic set_feature that must be called before requesting firmware string
+static uint8_t PREFIRMWARE_QUERY_MSG[] = {0x02, 0x01, 0x49, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00};
+
+static unsigned int REPORT_RATES[] = { 125, 250, 333, 500, 1000 };
+static uint8_t REPORT_RATES_ENCODED[] = {0x08, 0x04, 0x03, 0x02, 0x01};
+static uint8_t REPORT_RATES_CMD[] = {0x02, 0x06, 0xbb, 0xaa, 0x28, 0x00, 0x01, 0x00};
+
+static unsigned int DPILIST[] = { 1000, 2000, 3000, 5000, 15000};
+static uint8_t DPI_ENCODED[] = {0x04, 0x03, 0x02, 0x01, 0x00};
```
enkore, line 56: It's interesting that the sensor used in this mouse (PMW3310) supports setting DPI/CPI in increments of 50, but the firmware hides this. Unfortunate :(

Also 15000 is not a natively available setting on that sensor, so the firmware probably multiplies the sensor's output by 3 on this setting, which would result in missing codes (skipping).

src/driver-sinowealth-nubwo.c
```diff
@@ -0,0 +1,315 @@
...
+			resolution->dpi_x = resolution->dpi_y = DPILIST[0];
+			resolution->is_active = true;
+			resolution->is_default = true;
+		}
+		ratbag_profile_for_each_led(profile, led) {
+			led->mode = RATBAG_LED_OFF;
+			led->color.red = led->color.green = led->color.blue = 0;
+			led->colordepth = RATBAG_LED_COLORDEPTH_RGB_888;
+			led->type = RATBAG_LED_TYPE_SIDE;
+			led->color.red = led->color.green = led->color.blue = 0;
```
enkore, line 147: If I read your code correctly, then the code isn't reporting the current color of the LED back to ratbag, right? Is the sinowealthnubwo_aesthetic_report read-only?

src/driver-sinowealth-nubwo.c
```diff
@@ -0,0 +1,315 @@
...
+		if (REPORT_RATES[i] == reportrate)
+			return REPORT_RATES_ENCODED[i];
+	return REPORT_RATES_ENCODED[0];
+}
+
+static int
+sinowealthnubwo_set_dpi(struct ratbag_device *device, int dpi)
+{
+	uint8_t buf[SINOWEALTHNUBWO_PERF_CMD_MSGSIZE];
+	memset(buf, 0, SINOWEALTHNUBWO_PERF_CMD_MSGSIZE);
```
enkore, line 184: You can also write this as `uint8_t buf[...] = {0};` :)

src/driver-sinowealth-nubwo.c
```diff
@@ -0,0 +1,315 @@
...
+	uint8_t buff[SINOWEALTHNUBWO_GET_FIRMWARE_MSGSIZE];
+	int size;
+
+	size = ratbag_hidraw_set_feature_report(device, SINOWEALTHNUBWO_PERF_CMD_REPORTID, PREFIRMWARE_QUERY_MSG, ARRAY_LENGTH(PREFIRMWARE_QUERY_MSG));
+	if (size < 0)
+		log_error(device->ratbag, "Error while sending pre-firmware request message: %d\n", size);
+
+	size = ratbag_hidraw_get_feature_report(device, SINOWEALTHNUBWO_GET_FIRMWARE_CMD_REPORTID, buff, SINOWEALTHNUBWO_GET_FIRMWARE_MSGSIZE);
+	if (size != SINOWEALTHNUBWO_GET_FIRMWARE_MSGSIZE)
+		log_error(device->ratbag ,"Firmware report reply size mismatch expected %d got %d\n", SINOWEALTHNUBWO_GET_FIRMWARE_MSGSIZE, size);
```
enkore, line 106: In both of these `log_error` cases you might want to return a `strdup`'d constant, since log_error just emits the log message, it doesn't exit/panic the process or anything like that.

src/driver-sinowealth-nubwo.c
saengowp: That's correct. The official configuration interface can't read the config from the device. (except for dpi)

src/driver-sinowealth-nubwo.c
saengowp: TIL.

saengowp:
@enkore  . Thanks for a speedy review. I've refactored the code. Does it look good to you now?

saengowp:
> I hoped that "most" Sinowealth mice would use a similar protocol to what I found in the Glorious mice, but it looks like they made a bunch of entirely different protocols over time.

Yeah. I tried using the `sinowealth` driver but the mice just keep on reconnecting and exhibit strange behaviour. Not sure why they didn't just reuse the protocol. 🤷

enkore (approved):
lgtm like before. In case you already did the research on the "actually more supported" parts, it'd be great if you could add that information to the code (like for example what offset in DPI_CMD tells the mouse which DPI slot to update)

enkore:
No objections in three weeks... merging

Thanks for the contribution & patience, saengowp!

PR #1052, opened 2020-08-31
libratbag-data: log message if the device is a Logitech receivers

FFY00:
We have been receiving countless issues and PRs reporting or "fixing"
the device IDs because they think it's wrong. Hopefully this will give
them an hint that is intended behavior. If they look closely at the log,
on a recent enough kernel, they should see their device show up with the
internal ID.

### DIFF
--- a/src/libratbag-data.c
+++ b/src/libratbag-data.c
@@ -34,6 +34,7 @@
 #include "libratbag-private.h"
 #include "libratbag-data.h"
 #include "hidpp20.h"
+#include "usb-ids.h"
 
 #define GROUP_DEVICE "Device"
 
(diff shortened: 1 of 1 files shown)


FFY00:
@whot do you think it makes sense to do it here? Honestly I am not sure, I'd be incline to say that is fine, but I may be overlooking something.

src/libratbag-data.c
```diff
@@ -515,7 +515,9 @@ ratbag_device_data_new_for_id(struct ratbag *ratbag, const struct input_id *id)
 			goto out;
 	}
 
-	if (!data)
+	if (id->vendor == 0x046d && (id->product >> 8) == 0xc5)
```
whot, line 518: we have `USB_VENDOR_ID_LOGITECH` for the vendor ID which makes a little bit more obvious. Otherwise I'd probably go with this:

```c
/* The 0xc5* range is reserved for receivers */
if (... id->product & 0xff00 == 0xc500)
```
The only reason here is that 0xc5 is less obviously a 2-byte match for 0xc52b and others than 0xc500. But this is definitely personal preference, so feel free to go with what you think is better.

whot (approved):

PR #1062, opened 2020-09-22
sinowealth: remove per-resolution led

phomes:
The glorious mice uses a single profile. To differentiate between
resolutions each one has customizable color. A led on the bottom
of the mouse shows the active resolution color.

Libratbag does not support a per-resolution color.

To allow adjusting these colors the sinowealth driver defines a led
for each resolution. This is a misuse of the model and will cause
problems when the mouse is added to Piper.

This patch removes the per-resolution leds and hardcodes the led
count to one.

### DIFF
--- a/src/driver-sinowealth.c
+++ b/src/driver-sinowealth.c
@@ -309,14 +309,6 @@ sinowealth_read_profile(struct ratbag_profile *profile)
 	}
 	ratbag_led_unref(led);
 
-	/* DPI indicator LED */
-	for (int i = 1; i < SINOWEALTH_NUM_DPIS + 1; i++) {
-		led = ratbag_profile_get_led(profile, i);
-		led->mode = RATBAG_LED_ON;
-		led->color = sinowealth_raw_to_color(config->dpi_color[i - 1]);
-		ratbag_led_unref(led);
-	}
-
 	profile->is_active = true;
 
 	return 0;
@@ -335,7 +327,7 @@ sinowealth_init_profile(struct ratbag_device *device)
 	unsigned int dpis[num_dpis];
 
 	/* TODO: Button remapping */
-	ratbag_device_init_profiles(device, 1, SINOWEALTH_NUM_DPIS, 0, SINOWEALTH_NUM_DPIS + 1);
+	ratbag_device_init_profiles(device, 1, SINOWEALTH_NUM_DPIS, 0, 1);
 
 	profile = ratbag_device_get_profile(device, 0);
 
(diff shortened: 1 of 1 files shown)


phomes:
Note that I do not actually own a glorious mouse. I have only read/watched reviews of the mouse and its features. I do have some WIP for adding another sinowealth-based mouse to libratbag but that can go into the next release.

FFY00:
We could export a ratbag profile for each resolution. Every time someone changes one of the shared LEDs (so everything other than the resolution indicator) we would also propagate that for the other profiles. What do you think? Or am I misinterpreting something?

phomes:
Yes we could make the driver create a profile per resolution. As you say we would have to sync changes to the shared LED, buttons, report rate, etc between all profiles. Another option is to add an optional color field to resolution and api to change that.

This is yet another case where a corner case does not fit into our model. Should we add a hack like the fake profiles, or support the corner case in our model and take on the work of supporting it all the way in the API, ratbagctl and in Piper? Or just not support that corner case?

FFY00:
I don't think adding a color field to resolutions makes that much sense. There are definitely some other devices with profiles for only some fields, I think the best way to proceed is to make these fake-ish profiles as they solve the problem in a nice way, I think, and do not put much stress on the API. For eg. my friend has a mouse from Redragon and it only has profiles for the macros, this approach would also fit that. With that said, I don't see this as *that* much of a hack, it works pretty well IMO. What do you think?

We can merge the PR as is and make these changes later or wait, it's your call.

phomes:
I think that using fake profiles will break the users expectations. If I configure a profile in Piper then I would be surprised to see some of the changes also affecting the other profiles.

In my opinion we should extend our model to properly cover the functionality, or simply leave it unsupported. I would go for the latter in this case as I do not consider this feature critical to support.

enkore:
I noticed Piper doesn't seem to support more than three or four LEDs (it just doesn't display any controls for more LEDs, nothing bad happens), though I suspect people use only 2-3 DPI settings anyway. Personally I only use two, one for everything, and one with a super-low DPI for the odd console port with buggy mouse input.

Besides not all "fake LEDs" showing up, what problems are you thinking about?

I don't really have an opinion on this. It seems to me like extending the data model every time another oddball mouse comes around will result in a very complex data model, while adding fake profiles is probably rather confusing overall.

phomes:
The LED model is intended to be for separate physical LEDs on the device. In piper the LED will only show up if the SVG has a leader line pointing to its physical location on the device. Did you have 6 separate leader lines for the fake LEDs?

The problem is that we are exposing wrong data over our API, and we hope that the clients will then know how to deal with that. It will not be obvious in Piper that the 6 fake LEDs are actually different colors for resolutions that all belong to one single LED. We should not send fake data in the API hoping that the clients will deal with it.

It would be more correct to have a LED for the light effects within the mouse, and another for the resolution indicator on the bottom. But then we can only define a single color for the resolution...

I think that we should model the feature correctly if we think it is important, or otherwise leave it unsupported.

enkore:
> I think that we should model the feature correctly if we think it is important, or otherwise leave it unsupported.

I agree.

FFY00:
I am not proposing fake LEDs, I am proposing the physical LEDs that *do exist* on the device but do not have profiles to be exported in all profiles.

Let's take for example a mouse that has 4 LEDs but only one, say LED 0, is present on the profiles. This means if you change LEDs 1, 2 or 3 and change the profile, they will stay the same, because they are not configurable per profile. LED 0, however, is configured per profile.
What I am proposing here is to export ratbag profiles with 4 LEDs, but to match the device behavior, if you change LED 1, 2 or 3 on one profile, they are changed on all profiles.

I do not think ratbag profiles *need* to map exactly to how profiles are implemented in the device. We are abstracting the vendor API for all devices, I do not think it's wise to try to run after an API design that works perfectly for everything.

So, let's think about how this could be implemented? Adding a LED field to resolutions would solve this specific problem, but not similar issues when only some specific fields are configurable via profiles. What I think makes most sense is to allow extra fields independent of profiles.

With that said, I do understand and respect your opinion regarding the approach I am proposing.
I don't think we should introduce such big API changes, at least not for libratbag 1.0, maybe 2.0? So, for now, let's not support this use case.

FFY00:
Anyway, this can go in.

Issue #598, opened 2019-01-04
Second Mode/G-Shift support


CharoSW:
> Quick question about functionality support: is it a device limitation that I can't see a way to configure the "g-shift" (which is called second-mode by libratbag) functions, or is it a general libratbag problem? Happy to open a new issue if that's required.

I think this is something we never sorted out:
- you can assign a special key `second-mode` that triggers the G-shift
- you can not assign the buttons while in the G-Shift mode (at least not in ratbagctl)

IIRC, when working on the Roccat driver I duplicated every button, but that created a massive amount of visual data.

So there are 2 problems:
- one UI problem (how do we present these from the UI) -> that would be a piper issue
- one protocol problem: I am not sure how we can configure an alternate mapping for a button.

_Originally posted by @bentiss in https://github.com/_render_node/MDU6SXNzdWUzMDkzNjM3NTk=/issues/unread_timeline#issuecomment-451442215_

With that in mind are there any ideas around way that second-mode could be implemented effectively within the protocol.

BurgerJerk:
I came here to post this, as it stands now I have to run a windows VM, so I can bind to any of the " G-Shift " keys.

scj643:
This really needs to be implemented. I am fed up with the Logitech software being broken and not allowing me to unbind left and right click as the secondary binds.

whot:
@scj643 we'd really appreciate any help with the implementation. As it is, everyone is struggling for time.

grenzionky:
> I came here to post this, as it stands now I have to run a windows VM, so I can bind to any of the " G-Shift " keys.

how did u get the mouse to show up in the logitech gaming software within the vm?

mritunjaymusale:
Same, I have a G502 and I cannot use my binds which where on Gmode like I used to on windows

haroldma:
Same here, any way to do it via CLI?

Praxsidian:
> > I came here to post this, as it stands now I have to run a windows VM, so I can bind to any of the " G-Shift " keys.
> 
> @BurgerJerk how did u get the mouse to show up in the logitech gaming software within the vm?

I had two mice plugged in and then I passed through my logitech mouse to the VM.

FFY00:
Well, it seems we already have an api for `second-mode` so it shouldn't be that hard to add support.

Bernd-L:
> Well, it seems we already have an API for `second-mode` so it shouldn't be that hard to add support.

When do you expect progress in that regard?
[...]

PR #1064, opened 2020-09-25
Add support for G-Shift keys on g600

bluemner:
Code implements a possible solution to #598 

# Overview
Due to the locality of G-shift addresses in the memory of the mouse the G-Shift functions can be accessed via the normal button interface with a minor tweak of extending the count to 41. 20 buttons + 20 G-shift buttons + 1 color register.  
> Note:  The color register is between the two button 20 addresses so the code creates a "Phantom" button 20 which should not be used. 

In order to address this issue of the color address for the g-shift operations the code copies the led color from the profile into the g-shift color address. When pressing the g-shift button you get the same color as if you where not.

### DIFF
--- a/src/driver-logitech-g600.c
+++ b/src/driver-logitech-g600.c
@@ -34,7 +34,7 @@
 #include "libratbag-hidraw.h"
 
 #define LOGITECH_G600_NUM_PROFILES			3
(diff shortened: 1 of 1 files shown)


FFY00 (approved):
LGTM :+1:

@bentiss can you test?

PR #1065, opened 2020-09-28
WIP: sinowealth driver work for dm5

phomes:
This is a WIP pr for another device using the sinowealth driver.

Most things are compatible after I extracted device-specific features (number of buttons, number of leds, and sensor) out into the .device files.
Two difference remain. The color for the main LED seems to be read in RBG format in the current code, while the resolution color indicator is RGB. In the DM5 the main LED is RGB, but that can either be a device specific thing, or perhaps something that has changed in different version of the firmware.
The other difference is the LED modes, where most modes match but both devices has a few extra that the other does not.

In the WIP is also the start of my work on supporting button mapping.

It would be great to get feedback on any breakage on the glorious mice, and to know if the button reading code works there as well. Also any collaboration on the patches and missing parts are very welcome.

TODO for the pr:

- [ ] Button read completion (special keys, macros)
- [ ] Button write support
- [ ] Map extra LED modes


enkore:
Very nice, finally a mouse with a similar firmware/protocol! :)

I've only looked at button remapping in my mouse very superficially, but iirc it looked like what you've got here, with simple button-to-button remapping being done through the 0x12 descriptor and USB HID scancodes. Key sequences were stored as variable-length arrays with a separate HID feature report and referenced from 0x12.

src/driver-sinowealth.c
```diff
@@ -137,19 +138,39 @@ _Static_assert(sizeof(struct sinowealth_config_report) == SINOWEALTH_CONFIG_SIZE
...
+sinowealth_raw_to_dpi(struct ratbag_device *device, int raw)
 {
-	return (raw + 1) * 100;
+	struct sinowealth_data *drv_data = device->drv_data;
+	int dpi;
+
+	if (drv_data->is_sensor_3360)
+		dpi = (raw + 1) * 100;
+	else
+		dpi = raw * 100;
```
enkore, line 153: Is this correct? The DM5 is supposed to have a PWM3389, which has a 50 CPI step size.

src/driver-sinowealth.c
```diff
@@ -137,19 +138,39 @@ _Static_assert(sizeof(struct sinowealth_config_report) == SINOWEALTH_CONFIG_SIZE
 struct sinowealth_data {
 	/* this is kinda unnecessary at this time, but all the other drivers do it too ;) */
 	struct sinowealth_config_report config;
+	bool is_sensor_3360;
```
enkore, line 141: A comment here about the meaning of `!is_sensor_3360` might be good

src/driver-sinowealth.c
```diff
@@ -314,43 +335,164 @@ sinowealth_read_profile(struct ratbag_profile *profile)
...
-	dpis[0] = 0; /* 0 DPI = disabled */
-	for (int i = 1; i < num_dpis; i++) {
-		dpis[i] = SINOWEALTH_DPI_MIN + (i - 1) * SINOWEALTH_DPI_STEP;
+	if (led_count < 0)
+	{
+		log_error(device->ratbag, "Led count must be defined in .device file\n");
+		led_count = 0;
 	}
 
+	drv_data->is_sensor_3360 = dpirange->max == 12000;
```
enkore, line 469: Add this to the glorious device file instead?

src/driver-sinowealth.c
phomes: from my testing it seems the 100 CPI step is correct

src/driver-sinowealth.c
phomes: Like a "Sensor=3360" entry or so? We could do that. The downside is the increase in libratbag-data.[ch] for something that we so far can guess from the dpi Or we could set the dpi from the new sensor field instead...

src/driver-sinowealth.c
phomes: Maybe if we make it an enum sensor instead it would be more obvious?

phomes: I pushed a commit that changes this to an enum. Does that make it clearer?

phomes:
@enkore I pushed a commit to add button write support. In my eyes it should be correct but the mouse does not perform the change. Can you see anything wrong?

src/driver-sinowealth.c
```diff
@@ -314,43 +361,233 @@ sinowealth_read_profile(struct ratbag_profile *profile)
...
+			}
+
+			button_data.type = SINOWEALTH_BUTTON_TYPE_KEY;
+			button_data.data[0] = modifiers;
+			button_data.data[1] = key;
+		}
+	}
+
+	rc = ratbag_hidraw_set_feature_report(device, SINOWEALTH_REPORT_ID_CONFIG,
+					      (uint8_t*) buf, SINOWEALTH_CONFIG_SIZE);
```
enkore, line 521: IIRC the software always writes the macroblob (where it puts key-sequence macros) before this, perhaps the firmware requires that write order?

enkore:
It can write button remaps now, because remapping the DPI switching works, but macros/key presses don't. Hard-coding the key code in the write buttons function to "a" (0x4) doesn't work.

I use a command line like this to try:

```
./ratbagctl chanting-squirrel profile 0 button 4 action set macro KEY_A
```

src/driver-sinowealth.c
enkore: Very neat! :)

enkore:
Oh well, it actually DOES work. It's just that "a" becomes left alt somehow. So it's just an encoding issue on writing now.

```
# (after ratbad restart)
$ ./ratbagctl chanting-squirrel profile 0 button 4 action get
Button: 4 is mapped to macro '↓LEFTALT ↕RESERVED ↑LEFTALT'

$ xev
KeyPress event, serial 40, synthetic NO, window 0x9200001,
    root 0x1e2, subw 0x0, time 2519913, (72,50), root:(72,79),
    state 0x0, keycode 64 (keysym 0xffe9, Alt_L), same_screen YES,
    XLookupString gives 0 bytes:
    XmbLookupString gives 0 bytes:
    XFilterEvent returns: False
```

enkore:
It works now :) It was just key/modifier swapped and the keycode needs to be converted back to its HID usage code.

enkore:
Re-tested & confirmed: The model D's main LED is indeed RBG, as G/B are now swapped :D

phomes:
Perfect. You fixed it :) :+1:

phomes:
> Re-tested & confirmed: The model D's main LED is indeed RBG, as G/B are now swapped :D
> 
> I'm not sure how this is best handled. Add this to the .device file and pass the device into the color_to_raw functions so they can figure out what's needed for the device at hand?

That might be the only way if there is no way to detect the difference by inspecting the device. Perhaps we could check the firmware version and flip the behavior on that? I see that you have already added a definition for querying the firmware version but not yet made use of it.

phomes:
I pushed a commit to add a "LedType" into the .device file. Can you test if that fixes your device?

data/devices/glorious-model-d.device
```diff
@@ -3,3 +3,7 @@ Name=Glorious Model D
 DeviceMatch=usb:258a:0033
 Driver=sinowealth
 
+[Driver/sinowealth]
+Buttons=6
+Leds=2
+DpiRange=100:12000@100
```
enkore, line 9: Needs LedType=RBG as well

enkore:
Yep, works fine!

staticssleever668:
This should also properly support having no LEDs.

I didn't test much, but basic button remapping works on G-Wolves Hati HT.

src/driver-sinowealth.c
```diff
@@ -314,43 +367,243 @@ sinowealth_read_profile(struct ratbag_profile *profile)
...
+sinowealth_write_buttons(struct ratbag_profile *profile)
+{
+	struct ratbag_device *device = profile->device;
+	struct sinowealth_data *drv_data = device->drv_data;
+	struct sinowealth_button_report *buf = &drv_data->buttons;
+	struct ratbag_button *button;
+	int rc;
+
+	buf->report_id = 0x04;
+	buf->command_id = 0x12;
```
staticssleever668, line 496: Shouldn't these two be `SINOWEALTH_REPORT_ID_CONFIG` and `SINOWEALTH_CMD_GET_BUTTONS`?

src/driver-sinowealth.c
```diff
@@ -314,43 +367,243 @@ sinowealth_read_profile(struct ratbag_profile *profile)
...
+	else
+		drv_data->led_type = LED_RGB;
+
+	ratbag_device_init_profiles(device, 1, SINOWEALTH_NUM_DPIS, button_count, led_count);
+
+	profile = ratbag_device_get_profile(device, 0);
+
 	ratbag_profile_for_each_resolution(profile, resolution) {
-		ratbag_resolution_set_dpi_list(resolution, dpis, num_dpis);
+		ratbag_resolution_set_dpi_list_from_range(resolution,
```
staticssleever668, line 587: We should also somehow allow 0 DPI to disable the slot.

src/driver-sinowealth.c
```diff
@@ -436,14 +695,17 @@ sinowealth_commit(struct ratbag_device *device)
...
 		break;
 	case RATBAG_LED_CYCLE:
 		config->rgb_effect = RGB_GLORIOUS;
 		config->glorious_mode = sinowealth_led_to_rgb_mode(led);
 		break;
 	case RATBAG_LED_BREATHING:
-		config->rgb_effect = RGB_BREATHING1;
+		config->rgb_effect = RGB_BREATHING7;
 		config->breathing1_color = sinowealth_color_to_rbg(led->color);
 		config->breathing1_mode = sinowealth_led_to_rgb_mode(led);
```
staticssleever668, line 710: This should probably change options for RGB_BREATHING7.

perigoso:
Hey, do you have any time to rebase and test this? would be nice to finally merge

hrw:
I have "Genesys Xenon 770" which is also 258a:0027 sinowealth mouse. With 3 or 9 side buttons (depending on which addon I use).

hrw:
Rebased it on top of master branch:

```
12:24 (s) marcin@puchatek:build$ ratbagctl singing-gundi info
singing-gundi - SINOWEALTH Wired Gaming Mouse
             Model: usb:258a:0027:0
 Number of Buttons: 17
    Number of Leds: 2
Number of Profiles: 1
Profile 0: (active)
  Name: n/a
  Report Rate: 1000Hz
  Resolutions:
    0: 700x700dpi
    1: 1100x1100dpi
    2: 2300x2300dpi
    3: 3500x3500dpi (active) (default)
    4: 4700x4700dpi
    5: 6100x6100dpi
    6: <disabled>
    7: <disabled>
  Button: 0 is mapped to 'button 1'
  Button: 1 is mapped to 'button 2'
  Button: 2 is mapped to 'button 3'
  Button: 3 is mapped to UNKNOWN
  Button: 4 is mapped to UNKNOWN
  Button: 5 is mapped to 'button 5'
  Button: 6 is mapped to 'button 4'
  Button: 7 is mapped to 'button 5'
  Button: 8 is mapped to macro '↓LEFTCTRL ↕RESERVED ↑LEFTCTRL'
  Button: 9 is mapped to macro '↕PAGEUP'
  Button: 10 is mapped to macro '↕4'
  Button: 11 is mapped to macro '↓LEFTSHIFT ↕RESERVED ↑LEFTSHIFT'
  Button: 12 is mapped to macro '↕PAGEDOWN'
  Button: 13 is mapped to macro '↓LEFTALT ↕A ↑LEFTALT'
  Button: 14 is mapped to UNKNOWN
  Button: 15 is mapped to UNKNOWN
  Button: 16 is mapped to UNKNOWN
  LED: 0, depth: rgb, mode: cycle, duration: 1000, brightness: 192
  LED: 1, depth: rgb, mode: off
```

I need to check how buttons are counted.

hrw:
rebased tree: https://github.com/hrw/fork-libratbag/tree/upstream-dm5-rebased-for-genesys-nexus-770

hrw:
I wonder how to detect is 258a:0027 a DM5 or GN770 mouse.

staticssleever668:
@hrw, hi, I think you should know that G-Wolves Hati HT-M also has the same PID. :)
It has no LEDs (which is detectable in configuration data, but that zone is writable) and only has 6 buttons.
If you wish to improve the Sinowealth driver further, you should check out [gloryctl](https://github.com/sammko/gloryctl), which is a standalone Glorious Model O configuration program. I believe it also has the same license as libratbag.

staticssleever668:
Since there are at least 3 mice with the same PID, I now think device file in this PR should be renamed to something like "Sinowealth Generic Mouse". This also further proofs we should detect button and LED count at runtime.

src/driver-sinowealth.c
```diff
@@ -314,43 +367,243 @@ sinowealth_read_profile(struct ratbag_profile *profile)
...
+	const char *led_type;
+
+	dpirange = ratbag_device_data_sinowealth_get_dpi_range(device->data);
+	button_count = ratbag_device_data_sinowealth_get_button_count(device->data);
+	led_count = ratbag_device_data_sinowealth_get_led_count(device->data);
+	led_type = ratbag_device_data_sinowealth_get_led_type(device->data);
+
+	if (!dpirange)
+	{
+		log_error(device->ratbag, "DpiRange must be defined in .device file\n");
```
staticssleever668, line 554: We need some kind of fallback value in this case as we use the value several times later on.

PR #1067, opened 2020-09-28
meson: bump required meson version

phomes:
Bump the required meson version to match the features that we make use of.

We now also make use of the 'copy:' keyword in configure_file that is available in version 0.47.


FFY00:
Can you rebase on master so that we have feadore 32?

phomes:
> Can you rebase on master so that we have fedora 32?

done :+1:

FFY00 (approved):

PR #1068, opened 2020-09-28
Firmware version

phomes:
This adds fields to the device struct to store the firmware version on the device. It is also exposed over dbus to the clients.

The firmware of a device can define the protocol or memory layout that should be used to talk to the device. Several devices can query the firmware version via separate commands and we can use this information in the drivers to change behavior of the driver based on the device firmware.

The firmware version is also exposed over dbus so that it is easily available for users via e.g. ratbagctl. It can also be used in piper to warn users that a known bug in a firmware version can cause problems when trying to configure with libratbag.

### DIFF
--- a/doc/dbus.rst
+++ b/doc/dbus.rst
@@ -151,6 +151,20 @@ known to ratbagd.
         Vendor or product IDs of 0 are valid IDs (e.g. used used by test
         devices).
 
+.. attribute:: FirmwareMajor
+
+        :type: i
+        :flags: read-only, constant
+
+        The device's firmware major version.
+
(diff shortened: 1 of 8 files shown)


FFY00 (commented):
Hum, I think this is probably better as a string. A single major and minor number is a bit limiting.

enkore:
iirc for the sinowealth mice it looked like "V1.0.4" or something like that. I think a string is better suited here, too.

perigoso:
I agree, for instance openinput devices with non-release firmware would have a version related to the git hash like `r281.beb4a7a` hence being unmanageable with this strategy

Issue #1008, opened 2020-06-12
Logitech G915 TKL is not recognized


zasran:
**Is your feature request related to a problem? Please describe.**

Logitech G915 TKL is not recognized ratbagctl (from ratbagd 0.13-1 which is part of Ubuntu 20.04). It's the new (released June 2020) smaller version of full size G915. It seems like G915 is recognized, judging by presence of /usr/share/libratbag/logitech-g915.device (also read that it was added to libratbag).

**Describe the solution you'd like**

Would like for G915 TKL to be recognized (all connections: lightspeed, bluetooth, wired) and to be able to set LEDs etc.

**Describe alternatives you've considered**

Tried lightspeed and wired connections, ran ratbagctl with and without sudo. None of those help.

**Additional context**

Output of ratbagctl (I do have MX Master 2S connected via bluetooth):

```
$ ratbagctl list
hooting-chinchilla:  MX Master 2S           
```         

When I connect the keyboard via USB cable I get following /var/log/syslog messages:
```
Jun 12 01:33:49 jozinko kernel: [12280.948661] usb 1-2: new full-speed USB device number 14 using xhci_hcd
Jun 12 01:33:49 jozinko kernel: [12281.100717] usb 1-2: New USB device found, idVendor=046d, idProduct=c343, bcdDevice=14.00
Jun 12 01:33:49 jozinko kernel: [12281.100725] usb 1-2: New USB device strings: Mfr=1, Product=2, SerialNumber=3
Jun 12 01:33:49 jozinko kernel: [12281.100729] usb 1-2: Product: G915 TKL LIGHTSPEED Wireless RGB Mechanical Gaming Keyboard
Jun 12 01:33:49 jozinko kernel: [12281.100732] usb 1-2: Manufacturer: Logitech
Jun 12 01:33:49 jozinko kernel: [12281.100735] usb 1-2: SerialNumber: BA346238
Jun 12 01:33:49 jozinko kernel: [12281.107039] input: Logitech G915 TKL LIGHTSPEED Wireless RGB Mechanical Gaming Keyboard as /devices/pci0000:00/0000:00:14.0/usb1/1-2/1-2:1.0/0003:046D:C343.000E/input/input41
[...]

PR #1071, opened 2020-10-01
Add further device identificator for Logitech G915

opal06:
This is the output of lsusb:

`Bus 001 Device 003: ID 046d:c541 Logitech, Inc. USB Receiver`

It seems like Logietch might use different USB receivers on newer samples of this product.

### DIFF
--- a/data/devices/logitech-g915.device
+++ b/data/devices/logitech-g915.device
@@ -1,6 +1,6 @@
 [Device]
 Name=Logitech G915
-DeviceMatch=usb:046d:c33e
+DeviceMatch=usb:046d:c33e;usb:046d:c541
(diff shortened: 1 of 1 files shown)


FFY00:
See https://github.com/libratbag/libratbag/issues/1008#issuecomment-663874369

TLDR: Can't merge this now because we have no way to distinguish between different devices using the same receiver ID.

opal06:
I see. So the issue is that if multiple g915s are connected to the same lightspeed dongle it would not be able to tell which device is which? Or that multiple devices use this dongle id and so devices could be identified as the wrong device?
Anyway, take your time and thanks for your response, atm it works for me by modifying the device file, so I'm fine for now.

PR #1077, opened 2020-10-15
data: Add configuration for SteelSeries Kinzu V2 Pro Edition

kepstin:
Appears to be basically the same as the V2 - the differences were
mostly in hardware (upgraded switches & image sensor). I've verified
that the list of DPI settings is correct and matches the Windows
software. Adjusting DPI config in both profiles works correctly.


FFY00 (approved):
:tada:

PR #1112, opened 2021-01-02
Add support for Glorious Model O Wireless

vihu:
- Update meson.build
- Add data/devices/glorious-model-o-wireless.device

Any pointers on how to test this would be appreciated. Also I have casually assumed that the drivers have not changed between model O vs model O wireless.


FFY00:
Just go over the features that are exposed (DPI, leds, button mappings) and see if they work. If anything changed they probably won't.

vihu:
Yeah, this doesn't quite work as I had intended. Closing.

PR #1118, opened 2021-01-14
Add Logitech G Pro X Superlight device

brianegan:
[Logitech G Pro X Superlight](https://www.logitechg.com/en-us/products/gaming-mice/pro-x-superlight-wireless-mouse.910-005878.html) is a new Logitech Gaming mouse released around November 2020. Just got one, needed libratbag support, so added this file.

Verified as "Working on my machine" and thought it would help others, not sure if there's any other things you need for a merge.

### DIFF
--- a/data/devices/logitech-g-pro-x-superlight.device
+++ b/data/devices/logitech-g-pro-x-superlight.device
@@ -1,0 +1,7 @@
+[Device]
+Name=Logitech G Pro X Superlight
+DeviceMatch=usb:046d:c547;usb:046d:c094
+Driver=hidpp20
+
+[Driver/hidpp20]
+DeviceIndex=1
--- a/meson.build
+++ b/meson.build
@@ -296,6 +296,7 @@ data_files = files(
 	'data/devices/logitech-T650.device',
 	'data/devices/logitech-Wireless-Touchpad.device',
 	'data/devices/logitech-g-powerplay.device',
+	'data/devices/logitech-g-pro-x-superlight.device',
 	'data/devices/logitech-g-pro-wireless.device',
 	'data/devices/logitech-g-pro.device',
 	'data/devices/logitech-g102-g203.device',


FFY00:
Hi, thanks! You also need to add the .device file to `meson.build` so that it gets installed to the system.

brianegan:
Thanks, all checks passing now :) Please lemme know if I can make any further changes.

data/devices/logitech-g-pro-x-superlight.device
```diff
@@ -0,0 +1,7 @@
+[Device]
+Name=Logitech G Pro X Superlight
+DeviceMatch=usb:046d:c547;usb:046d:c094
```
FFY00, line 3: `c547` is the ID of the receiver, the kernel driver export a hidraw node just for the device itself, that what we want. Can you use that instead? You can find it by either looking at the logs or using https://github.com/FFY00/lshid.

FFY00 (changes requested):

data/devices/logitech-g-pro-x-superlight.device
FFY00: Actually nevermind, this receiver is not yet supported in the kernel driver. This PR is stalled.

brianegan: Ah, interesting -- I have no experience with hardware, but there must be some kind of support for this receiver since this configuration file allows me to detect and configure the mouse using ratbagctl or my fork of piper: https://github.com/libratbag/piper/pull/616.

Anyhow, I'll close these PRs out for now since I don't have any more time to dedicate to them, but please feel free to use the code and SVGs for piper in the future if it's helpful!

FFY00: No worries, the problem is that different device might use the same receiver, so same ID. While it is fine for you to have use the receiver ID locally, as you do not have any devices that collide, we can't do that in the upstream.

PR #1122, opened 2021-01-22
data: add G Pro X Wireless Superlight

FFY00:
Support for this receiver, and for the device using the powerplay
receiver, still needs to be added to the kernel driver. For now, you can
used the wired connection to configure the device.

You should also not try to use with the powerplay receiver as the kernel
driver does not support the connection and won't forward any input
events to userspace.

### DIFF
--- a/data/devices/logitech-g-pro-x-wireless-superlight.device
+++ b/data/devices/logitech-g-pro-x-wireless-superlight.device
@@ -1,0 +1,5 @@
+[Device]
+Name=Logitech G Pro X Wireless Superlight
+DeviceMatch=usb:046d:4093;usb:046d:c094
+LedTypes=logo;battery;dpi;
+Driver=hidpp20
--- a/meson.build
+++ b/meson.build
@@ -297,6 +297,7 @@ data_files = files(
 	'data/devices/logitech-Wireless-Touchpad.device',
 	'data/devices/logitech-g-powerplay.device',
 	'data/devices/logitech-g-pro-wireless.device',
+	'data/devices/logitech-g-pro-x-wireless-superlight.device',
 	'data/devices/logitech-g-pro.device',
 	'data/devices/logitech-g102-g203.device',
 	'data/devices/logitech-g300.device',


FFY00:
@bentiss, you may want to be aware of this behavior. Essentially the kernel driver trips when trying to determine the connection type and won't forward any events to userspace because it does not know how to handle the connection.

```
[915977.398471] logitech-djreceiver 0003:046D:C53A.0107: unusable device of type UNKNOWN (0x0f) connected on slot 1
```

This is fine on the receiver that came with the device, as we will add support for that receiver together with the new connection type. But for receivers that support multiple types, which right now it's only the powerplay mat, that won't happen. In which case the user could just use the receiver that came with the device, or blocklist the kernel driver.

I am not really sure how to handle this, I will have to have look at the driver.

deadmeu:
I own this mouse and would very much like to start configuring it with piper asap :slightly_smiling_face:

Vash63:
Are those logitech-dj patches still trying to be upstreamed? Are they still applicable to 5.13?

FFY00:
The one adding the new receiver hasn't been upstreamed yet, the one fixing device when connected to a Powerplay mat is. Currently, the device only works with libratbag when connected via USB directly.

Vash63:
Ok thanks, I can confirm it works via USB (would like to see the battery level also like I could on my G503). I was just checking if there was news since the last notes I could find on LKML were from early march for the 5.11 cycle.

SebbeM:
Should this be working yet? I'm not sure if I'm doing something wrong or I just misunderstood what has been said in the discussion. I am able to use the mouse with the provided Lightspeed receiver and directly over USB but I get no connection to the Powerplay receiver. Shouldn't it be working through the Powerplay receiver as well now?

FFY00:
There was kernel bug regarding this, fixed in https://github.com/torvalds/linux/commit/fab3a95654eea01d6b0204995be8b7492a00d001. Which kernel are you running?

SebbeM:
Should have included that right away, of course. I'm running Linux 5.15.2-2.

SebbeM:
I have done some basic testing now but can not figure out what is wrong. The device shows up in Solaar with charge, DPI and such but the mouse does send any input. I can also unpair the mouse from Solaar, which causes the Powerplay pad to enter pairing mode and when I restart the mouse, it gets paired again. Piper still only shows the 404 image but I can change the LED color from the LED tab.

SebbeM:
Yes, I thought that I should return and update on the issue but I am a bad person. When I updated the kernel to version 5.17.9-1, it all started working properly. I can't be completely sure that nothing else updated at about the same time but I think it was the kernel update that did it. I'm running Manjaro, if that matters.

lpadula:
> Yes, I thought that I should return and update on the issue but I am a bad person. When I updated the kernel to version 5.17.9-1, it all started working properly. I can't be completely sure that nothing else updated at about the same time but I think it was the kernel update that did it. I'm running Manjaro, if that matters.

Im running Manjaro with 5.17 kernel, but Piper wont recognize my Logitech G Pro X Wireless Superlight. What Piper package version is working for you?

SebbeM:
I'm running version 0.6-2, installed from official repositories, packaged by Felix Yan.

gezawatt:
I have this mouse with this receiver and I can't wait for this commit to get accepted

PR #1128, opened 2021-02-05
Add openinput driver

FFY00:
WIP https://github.com/openinput-fw/openinput/tree/master

cc @perigoso

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -344,6 +344,7 @@ data_files = files(
 	'data/devices/logitech-MX-Anywhere2S.device',
 	'data/devices/logitech-MX-Anywhere-3.device',
 	'data/devices/logitech-MX-Vertical.device',
+	'data/devices/openinput.device',
 	'data/devices/roccat-kone-pure.device',
 	'data/devices/roccat-kone-xtd.device',
 	'data/devices/steelseries-kinzu-v2.device',
--- a/meson.build
+++ b/meson.build
@@ -174,6 +174,7 @@ src_libratbag = [
 	'src/driver-roccat.c',
 	'src/driver-roccat-kone-pure.c',
 	'src/driver-gskill.c',
+	'src/driver-openinput.c',
 	'src/driver-steelseries.c',
 	'src/driver-sinowealth.c',
 	'src/driver-sinowealth-nubwo.c',
--- a/src/driver-openinput.c
+++ b/src/driver-openinput.c
@@ -168,8 +168,9 @@ openinput_info_version(struct ratbag_device *device)
 	return 0;
 }
 
-#define OI_FUNCTION_FW_INFO_VENDOR	0x00
-#define OI_FUNCTION_FW_INFO_VERSION	0x01
+#define OI_FUNCTION_FW_INFO_VENDOR		0x00
+#define OI_FUNCTION_FW_INFO_VERSION		0x01
+#define OI_FUNCTION_FW_INFO_DEVICE_NAME		0x02
 
 static int
 openinput_info_fw_info(struct ratbag_device *device,
@@ -234,6 +235,10 @@ openinput_probe(struct ratbag_device *device)
 	ret = openinput_info_fw_info(device, OI_FUNCTION_FW_INFO_VERSION, str, sizeof(str));
 	if (!ret)
 		log_info(device->ratbag, "openinput: firmware version: %s\n", str);
+
+	ret = openinput_info_fw_info(device, OI_FUNCTION_FW_INFO_DEVICE_NAME, str, sizeof(str));
+	if (!ret)
+		log_info(device->ratbag, "openinput: device: %s\n", str);
 
 	ratbag_device_init_profiles(device,
 				    drv_data->num_profiles,
--- a/src/driver-openinput.c
+++ b/src/driver-openinput.c
@@ -1,0 +1,79 @@
+/*
+ * Copyright © 2021 Filipe Laíns
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a
+ * copy of this software and associated documentation files (the "Software"),
+ * to deal in the Software without restriction, including without limitation
+ * the rights to use, copy, modify, merge, publish, distribute, sublicense,
+ * and/or sell copies of the Software, and to permit persons to whom the
+ * Software is furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice (including the next
+ * paragraph) shall be included in all copies or substantial portions of the
+ * Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.  IN NO EVENT SHALL
+ * THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+ * FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+ * DEALINGS IN THE SOFTWARE.
+ */
+
+#include "config.h"
+
+#include "libratbag-private.h"
+
+static unsigned int report_rates[] = { 125, 250, 500, 750, 1000 };
+
+struct openinput_drv_data {
+	unsigned int num_profiles;
+	unsigned int num_resolutions;
+	unsigned int num_buttons;
+	unsigned int num_leds;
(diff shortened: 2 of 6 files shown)


FFY00:
Arch is failing due to https://bugs.archlinux.org/task/69563, I believe this should be fixed when the docker image is updated.

FFY00:
I will add support for custom functions in future PRs.

meson.build
```diff
@@ -343,6 +344,7 @@ data_files = files(
 	'data/devices/logitech-MX-Anywhere2S.device',
 	'data/devices/logitech-MX-Anywhere-3.device',
 	'data/devices/logitech-MX-Vertical.device',
+	'data/devices/openinput.device',
```
whot, line 347: unrelated to this commit: we should use `install_subdir()` here which avoids having to name all `.device` files.

src/driver-openinput.c
```diff
@@ -0,0 +1,483 @@
...
+	uint64_t supported;
+};
+
+struct oi_report_t {
+	uint8_t id;
+	uint8_t function_page;
+	uint8_t function;
+	uint8_t data[29];
+} __attribute__((__packed__));
+
```
whot, line 76: /me giggles at the `oi` prefix.

it's the Australian version of "Hey, you!"

src/driver-openinput.c
```diff
@@ -0,0 +1,483 @@
...
+
+/* gcc is being picky and asking for the prototypes to bve defined (?) */
```
whot, line 78: you forgot the `static`, that's why :)

src/driver-openinput.c
```diff
@@ -0,0 +1,483 @@
...
+const char* openinput_function_get_name(uint8_t page, uint8_t function);
+
+#define CASE_RETURN_STRING(a) case a: return #a; break
+const char* openinput_function_page_get_name(uint8_t page)
+{
+	static char numeric[16];
+	char *str;
+
+	switch(page)
+	{
```
whot, line 89: nitpick on style: rest of the code used to/uses `{` on the same line

src/driver-openinput.c
```diff
@@ -0,0 +1,483 @@
...
+
+	if (!str)
+	{
+		sprintf_safe(numeric, "0x%02x 0x%02x", page, function);
+		str = numeric;
+	}
+
+	return str;
+}
+#undef CASE_RETURN_STRING
```
whot, line 128: might be worth moving that macro to `shared-macro.h` but that's mostly unrelated to this patchset.

src/driver-openinput.c
```diff
@@ -0,0 +1,483 @@
...
+	default:
+		sprintf_safe(numeric, "0x%02x", page);
+		str = numeric;
+		break;
+	}
+
+	return str;
+}
+
+const char* openinput_function_get_name(uint8_t page, uint8_t function)
```
whot, line 103: nitpick on style: at least the first file I looked at (`driver-hidpp10.c`) uses:
```
const char*
function name(...)
```

src/driver-openinput.c
```diff
@@ -0,0 +1,483 @@
...
+		case OI_ERROR_CUSTOM:
+			memcpy(help_str, report->data, sizeof(help_str)); /* copy error string */
+			help_str[sizeof(help_str)-1] = '\0'; /* terminate string */
+			asprintf(&str, "Custom error (%s)", help_str);
+			break;
+		case OI_ERROR_UNSUPPORTED_FUNCTION:
+			asprintf(&str, "Unsupported function (0x%02x, 0x%02x)", report->data[0], report->data[1]);
+			break;
+		default:
+			asprintf(&str, "Unkown error (%u)", report->function);
```
whot, line 148: If `asprintf` fails, `str` is undefined, use our own `xasprintf` instead

src/driver-openinput.c
```diff
@@ -0,0 +1,483 @@
...
+		str = numeric;
+	}
+
+	return str;
+}
+#undef CASE_RETURN_STRING
+
+static const char* openinput_get_error_string(struct oi_report_t *report)
+{
+	char *str = "", help_str[OI_REPORT_LONG_SIZE - OI_REPORT_DATA_INDEX + 1];
```
whot, line 132: two things here:
- make this two separate declarations so it's clearer. 
  - You could also just do `help_str[...] = {0}` and later `memcpy(sizeof-1)` to force null-termination but that's up to you
- `str` doesnt' need initialization since you overwrite it in all cases

src/driver-openinput.c
```diff
@@ -0,0 +1,483 @@
...
+		case OI_ERROR_INVALID_VALUE:
+			asprintf(&str, "Invalid value (in position %u)", report->data[2]);
+			break;
+		case OI_ERROR_CUSTOM:
+			memcpy(help_str, report->data, sizeof(help_str)); /* copy error string */
+			help_str[sizeof(help_str)-1] = '\0'; /* terminate string */
+			asprintf(&str, "Custom error (%s)", help_str);
+			break;
+		case OI_ERROR_UNSUPPORTED_FUNCTION:
+			asprintf(&str, "Unsupported function (0x%02x, 0x%02x)", report->data[0], report->data[1]);
```
whot, line 145: printf directive `%#x` prints the `0x` prefix, so `%02#x` in this case, iirc.

src/driver-openinput.c
```diff
@@ -0,0 +1,483 @@
...
+		if (total != (read + count + left))
+		{
+			log_error(ratbag, "openinput: invalid number of functions left to read (%u)\n", left);
+			return -EINVAL;
+		}
+		log_debug(ratbag, "openinput: read %u functions, %u left\n", count, left);
+
+		memcpy(functions + read, buffer, count);
+		read += count;
+	}
```
whot, line 347: very much a nitpick but it *seems* like this could be a `do { } while()` instead to not have two separate but identical invocations of `openinput_info_supported_functions`. haven't fully engaged my brain to make sure that's true though.

FFY00: I did that originally but it was a bit of a mess doing to the validation logic, note the declaration of `functions` based on a value return by the command, so I opted to duplicate the command making it a bit easier to read. This does seem like a good optimization at first sight but I found that the code duplication is much easier to read than the increased logic.

src/driver-openinput.c
FFY00: I can do that in a future PR.

src/driver-openinput.c
FFY00: Ah 😅

src/driver-openinput.c
FFY00: Here too 😛

meson.build
FFY00: Yes, I am not sure when that got added, meson didn't used to have it IIRC.

whot: ages ago. I remember looking it up for libwacom (same thing there) and eventually stopped bothering finding the exact version number. I think 0.22 or something like that, quick git log shows patches from 2014. Not sure if it was just a matter of me copy/pasting or not finding it when the inital meson support was added but eh.

src/driver-openinput.c
FFY00: When using this in `xasprintf`, it gives me:
```
../src/driver-openinput.c:139:45: warning: unknown conversion type character ‘#’ in format [-Wformat=]
  139 |   xasprintf(&str, "Unsupported function (%02#x, %02#x)", report->data[0], report->data[1]);
      |                                             ^
../src/driver-openinput.c:139:52: warning: unknown conversion type character ‘#’ in format [-Wformat=]
  139 |   xasprintf(&str, "Unsupported function (%02#x, %02#x)", report->data[0], report->data[1]);
      |                                                    ^
../src/driver-openinput.c:139:19: warning: too many arguments for format [-Wformat-extra-args]
  139 |   xasprintf(&str, "Unsupported function (%02#x, %02#x)", report->data[0], report->data[1]);
      |                   ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

The `sprintf_safe` usages seem to be fine.

FFY00: Well, they are not fine, they are printing the literal. Seems like `%#02x` works there. But I noticed that using `#` will not add the prefix when the value is 0 and will not correctly pad with 0 like it previously did.

```
ratbag debug: openinput: found function 00 00
ratbag debug: openinput: found function 00 0x1
ratbag debug: openinput: found function 00 0x2
ratbag debug: openinput: found function 00 0x3
```

Going back to the old format:

```
ratbag debug: openinput: found function 0x00 0x00
ratbag debug: openinput: found function 0x00 0x01
ratbag debug: openinput: found function 0x00 0x02
ratbag debug: openinput: found function 0x00 0x03
```

So I think I'll stick with that :sweat_smile:

whot: > But I noticed that using # will not add the prefix when the value is 0 

urgh, true. forgot about that. fun fact, `%#X` will uppercase hex but it will also uppercase the `X` to `0X12AB` which is probably the most useless feature ever...

> and will not correctly pad with 0 like it previously did.

the `0x` counts towards the length so we would need `%#04x` here to get `0x01`. But yeah, the zero is the killer here.

PR #1138, opened 2021-02-28
ci: temporarily disable arch

FFY00:
See my description of the problem in [1].
Essentially, docker is mishandling the failuire when a syscall is not
available, returning EPERM instead of ENOSYS, and making code misbehave.
The current issue with arch is glibc, which explodes on EPERM, causing
pacman to misbehave.

Once the fix hits CircleCI, we will re-enable.

[1] https://github.com/fwupd/fwupd/issues/2894

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -258,7 +258,7 @@ workflows:
   compile_and_test:
     jobs:
       - ubuntu_20_04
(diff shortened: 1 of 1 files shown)


whot (approved):
does what it says on the box :)

PR #1146, opened 2021-04-03
data: Add (old) Logitech Performance Mouse MX

strangemax:
Add support for this old business Logitech mouse.

Hint: I'm new to libratbag, so sorry if there is something missing.


FFY00 (approved):
Sorry for the delay. Thanks :blush:

FFY00:
Actually, sorry. I misread the USB ID. This ID belongs to the receiver, we need to add support to the kernel driver first.

PR #1151, opened 2021-04-16
openinput: ignore non protocol related reports

perigoso:
Ignore reports unrelated to the protocol communication.

Signed-off-by: Rafael Silva <perigoso@riseup.net>

### DIFF
--- a/src/driver-openinput.c
+++ b/src/driver-openinput.c
@@ -175,11 +175,24 @@ openinput_send_report(struct ratbag_device *device, struct oi_report_t *report)
 		return ret;
 	}
 
-	ret = ratbag_hidraw_read_input_report(device, buffer, OI_REPORT_MAX_SIZE);
-	if (ret < 0) {
-		log_error(device->ratbag, "openinput: failed to read data from device (%s)\n",
-			  strerror(-ret));
-		return ret;
+	/* ignore up to 10 non protocol related reports */
+	uint8_t read_attempts = 10;
+
+	do
+	{
+		ret = ratbag_hidraw_read_input_report(device, buffer, OI_REPORT_MAX_SIZE);
+		if (ret < 0) {
+			log_error(device->ratbag, "openinput: failed to read data from device (%s)\n",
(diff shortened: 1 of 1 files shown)


src/driver-openinput.c
```diff
@@ -175,11 +175,24 @@ openinput_send_report(struct ratbag_device *device, struct oi_report_t *report)
 		return ret;
 	}
 
-	ret = ratbag_hidraw_read_input_report(device, buffer, OI_REPORT_MAX_SIZE);
-	if (ret < 0) {
-		log_error(device->ratbag, "openinput: failed to read data from device (%s)\n",
-			  strerror(-ret));
-		return ret;
+	/* ignore up to 10 non protocol related reports */
+	uint8_t read_attempts = 10;
```
FFY00, line 179: I think we should do a timeout instead. 10 read attempts may be easily saturated by high report rate devices, and this routine may block much more easily if there are no reports coming in.

FFY00 (changes requested):

src/driver-openinput.c
FFY00: To implement a timeout properly, we need to use poll. Example: https://github.com/libratbag/libratbag/blob/230647c7f3ef28938bb66745bf0c81082631af95/src/hidpp-generic.c#L282

We should create a new version of `ratbag_hidraw_read_input_report` that has a timeout, `ratbag_hidraw_read_input_report_timeout`.

FFY00: Hum, maybe we should add a variant of `ratbag_hidraw_read_input_report` that takes a function pointer for a callback that verifies if the data is valid or not. I'd call it `ratbag_hidraw_read_input_report_filter`.

perigoso: right, I wanted to implement a timeout initially, but didn't see any obvious way to do it.

> and this routine may block much more easily if there are no reports coming in.
in that case it would only block for a single timeout, 1 second

PR #1152, opened 2021-04-18
data: add SEL variant to Roccat Kone Pure

Sethox:
The last part added is for a variant called "Roccat Kone Pure SEL".


FFY00 (approved):
Thanks!

PR #1156, opened 2021-04-27
libratbag-hidraw: add filter to hidraw read

FFY00:
Currently, we only have functions that will read and timeout after 1s if
no packets were received. The issue with this is that a lot of times we
may catch unrelated packets, such as mouse or keyboard changes.
To solve this, this patch adds a new set of functions suffixed with
_filter, that will keep reading packets until a valid packet if found.
Users define a filter function and pass it as an argument.

### DIFF
--- a/src/libratbag-hidraw.c
+++ b/src/libratbag-hidraw.c
@@ -1642,16 +1642,19 @@ ratbag_hidraw_output_report(struct ratbag_device *device, uint8_t *buf, size_t l
 }
 
 int
-ratbag_hidraw_read_input_report(struct ratbag_device *device, uint8_t *buf, size_t len)
+ratbag_hidraw_read_input_report(struct ratbag_device *device, uint8_t *buf, size_t len,
+				 ratbagd_hidraw_filter_t filter)
 {
-	return ratbag_hidraw_read_input_report_index(device, buf, len, 0);
+	return ratbag_hidraw_read_input_report_index(device, buf, len, 0, filter);
 }
 
 int
-ratbag_hidraw_read_input_report_index(struct ratbag_device *device, uint8_t *buf, size_t len, int hidrawno)
+ratbag_hidraw_read_input_report_index(struct ratbag_device *device, uint8_t *buf, size_t len, int hidrawno,
+				 ratbagd_hidraw_filter_t filter)
 {
-	int rc;
+	int rc, nfds;
 	struct pollfd fds;
+	int ts, ts_end;
 
 	assert(hidrawno >= 0 && hidrawno < MAX_HIDRAW);
 
@@ -1661,17 +1664,32 @@ ratbag_hidraw_read_input_report_index(struct ratbag_device *device, uint8_t *buf
 	fds.fd = device->hidraw[hidrawno].fd;
 	fds.events = POLLIN;
 
-	rc = poll(&fds, 1, 1000);
-	if (rc == -1)
-		return -errno;
+	/* convert to ms */
+	ts = now(CLOCK_MONOTONIC_RAW) / 1000 / 1000;
+	ts_end = ts + 1000; /* end in now + 1000ms */
 
-	if (rc == 0)
-		return -ETIMEDOUT;
+	while (ts < ts_end) {
+		nfds = poll(&fds, 1, ts_end - ts); /* poll for the remainder of timeout */
 
-	rc = read(device->hidraw[hidrawno].fd, buf, len);
+		if (nfds < 0)
+			return -errno;
 
-	if (rc > 0)
-		log_buf_raw(device->ratbag, "input report:  ", buf, rc);
+		if (nfds > 0) {
+			rc = read(device->hidraw[hidrawno].fd, buf, len);
 
-	return rc >= 0 ? rc : -errno;
+			if (rc < 0)
+  				return -errno;
+
+			if (rc > 0) {
+				if(!filter || (filter && filter(buf, rc))) {
+					log_buf_raw(device->ratbag, "input report:  ", buf, rc);
+					return rc;
+				}
+			}
+		}
+
+		ts = now(CLOCK_MONOTONIC_RAW) / 1000 / 1000;
+	}
+
+	return -ETIMEDOUT;
 }
--- a/src/libratbag-hidraw.h
+++ b/src/libratbag-hidraw.h
@@ -47,6 +47,8 @@ struct ratbag_hidraw {
 	char *sysname;
 };
 
+typedef bool (*ratbagd_hidraw_filter_t)(uint8_t *buf, size_t len);
+
 /**
  * Open the hidraw device associated with the device.
  *
@@ -162,27 +164,32 @@ int ratbag_hidraw_output_report(struct ratbag_device *device, uint8_t *buf, size
 
 /**
  * Read an input report from the device
+ * Optional filter function can be provided, when the filter returns false the packet is ignored
  *
  * @param device the ratbag device
  * @param[out] buf resulting raw data
  * @param len length of buf
+ * @param filter filter function
  *
  * @return count of data transfered, or a negative errno on error
  */
-int ratbag_hidraw_read_input_report(struct ratbag_device *device, uint8_t *buf, size_t len);
+int ratbag_hidraw_read_input_report(struct ratbag_device *device, uint8_t *buf, size_t len,
+				 ratbagd_hidraw_filter_t filter);
 
 /**
  * Read an input report from the device from a specific hidraw index
+ * Optional filter function can be provided, when the filter returns false the packet is ignored
  *
  * @param device the ratbag device
  * @param[out] buf resulting raw data
  * @param len length of buf
- * @param interal index of hidraw array
+ * @param hidrawno index of hidraw array
+ * @param filter filter function
  *
  * @return count of data transfered, or a negative errno on error
  */
-int ratbag_hidraw_read_input_report_index(struct ratbag_device *device, uint8_t *buf, size_t len, int hidrawno);
-
+int ratbag_hidraw_read_input_report_index(struct ratbag_device *device, uint8_t *buf, size_t len, int hidrawno,
+				 ratbagd_hidraw_filter_t filter);
 
 /**
  * Tells if a given device has the specified report ID.
--- a/src/driver-openinput.c
+++ b/src/driver-openinput.c
@@ -159,6 +159,22 @@ openinput_get_report_size(unsigned int report)
 	}
 }
 
+static bool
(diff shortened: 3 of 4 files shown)


FFY00:
Replaces #1151 

@perigoso, please give it a try

FFY00:
I just updated the to only log the read buffers on valid packets, which prevents ratbagd from being a keylogger in most situations if these functions are used.

perigoso:
Tested on the Rival 310 where I was abled to reproduce the issue reliably, everything working as expected.

```
ratbag info: openinput: protocol version 0.0.1
ratbag info: openinput: firmware vendor: openinput-git
ratbag info: openinput: firmware version: r206.419530a.dirty
ratbag info: openinput: device: openinput Device
```

perigoso (approved):

src/libratbag-hidraw.c
```diff
@@ -1660,3 +1660,49 @@ ratbag_hidraw_read_input_report_index(struct ratbag_device *device, uint8_t *buf
 
 	return rc >= 0 ? rc : -errno;
 }
+
+int
+ratbag_hidraw_read_input_report_filter(struct ratbag_device *device, uint8_t *buf, size_t len, int *(*filter)(uint8_t *buf, size_t len))
```
whot, line 1665: style nitpick: having the function pointer on the same line wraps it in github (and probably in most narrower terminals) so it makes it slightly confusing. Also, any reason you don't use a `typedef` for the function signature?

src/libratbag-hidraw.c
```diff
@@ -1660,3 +1660,49 @@ ratbag_hidraw_read_input_report_index(struct ratbag_device *device, uint8_t *buf
...
+{
+	return ratbag_hidraw_read_input_report_index_filter(device, buf, len, 0, filter);
+}
+
+int
+ratbag_hidraw_read_input_report_index_filter(
```
whot, line 1671: same goes for the function above, I'm wondering if `filtered` is clearer for a name, i.e.
`ratbag_hidraw_read_input_report_index_filtered`

src/libratbag-hidraw.c
```diff
@@ -1660,3 +1660,49 @@ ratbag_hidraw_read_input_report_index(struct ratbag_device *device, uint8_t *buf
...
+		rc = poll(&fds, 1, 1);  /* poll for 1msec */
+
+		if (rc == -1)
+			return -errno;
+
+		if (rc > 0) {
+			rc = read(device->hidraw[hidrawno].fd, buf, len);
+
+			if (rc >= 0) {
+				if (filter(buf, rc)) {
```
whot, line 1696: `filter()` should be a return type `bool`?

src/libratbag-hidraw.c
```diff
@@ -1660,3 +1660,49 @@ ratbag_hidraw_read_input_report_index(struct ratbag_device *device, uint8_t *buf
...
+
+	if (len < 1 || !buf || device->hidraw[hidrawno].fd < 0)
+		return -EINVAL;
+
+	fds.fd = device->hidraw[hidrawno].fd;
+	fds.events = POLLIN;
+
+	/* wait for valid messages for 1000 iterations (~1s) */
+	for (unsigned int i = 0; i < 1000; i++) {
+		rc = poll(&fds, 1, 1);  /* poll for 1msec */
```
whot, line 1687: I don't quite understand this approach, you timeout after 1ms but loop 1000 times - why not just time out after 1000ms? 
If the message comes faster it doesn't matter anyway, if nothing happens you skipped a bit of busy work.

Worst case you could adjust the timeout based on whether you already received some messages and keep looping, so that the first loop has 1000ms but the second loop only has e.g. 10ms for follow-up messages.

src/libratbag-hidraw.h
```diff
@@ -183,6 +183,33 @@ int ratbag_hidraw_read_input_report(struct ratbag_device *device, uint8_t *buf,
  */
 int ratbag_hidraw_read_input_report_index(struct ratbag_device *device, uint8_t *buf, size_t len, int hidrawno);
 
+/**
+ * Read an input report from the device util a specified filter function is successfull
+ *
+ * @param device the ratbag device
+ * @param[out] buf resulting raw data
+ * @param len length of buf
+ * @param filter filter function
```
whot, line 192: that one needs documentation because just looking at this I wouldn't know what to return for actual filtered data.

src/libratbag-hidraw.h
```diff
@@ -183,6 +183,33 @@ int ratbag_hidraw_read_input_report(struct ratbag_device *device, uint8_t *buf,
...
+ *
+ * @return count of data transfered, or a negative errno on error
+ */
+int ratbag_hidraw_read_input_report_filter(
+	struct ratbag_device *device, uint8_t *buf, size_t len, int *(*filter)(uint8_t *buf, size_t len));
```
whot, line 197: /me stares hard at `int *(*filter)` and wonders if this returns an `int*`?

one day I'll manage to read C function pointer declarations without having to think about it.

src/libratbag-hidraw.c
```diff
@@ -1660,3 +1660,49 @@ ratbag_hidraw_read_input_report_index(struct ratbag_device *device, uint8_t *buf
...
+
+			if (rc >= 0) {
+				if (filter(buf, rc)) {
+					log_buf_raw(device->ratbag, "input report:  ", buf, rc);
+					return rc;
+				}
+				continue;
+			}
+
+			return -errno;
```
whot, line 1703: I was wondering where the `errno` comes from but then I noticed that the `if (rc > 0)` has no path that gets here. Personally, I'd write this all as:
```c
if (rc < 0) 
  return -errno:

if (filter()) {
    return rc
}
```

Also, I *really* recommend not re-using `rc` for the `poll` and the `read` result, it's going to bite you one day. local variables are cheap, just use `nfds` or so for the `poll()` call.

src/libratbag-hidraw.c
FFY00: Because it provides resolution. Let's say I have a 1000ms loop and then 10 10ms ones. I start polling, and get a message after 30ms, but it does not pass the filter, we move over to the next loop, which is 10ms, and we only have 10 of them, which means the function will not wait 1s for valid messages, as I expected, here it will only wait 130ms, because we exhausted the loops.
The solution is many loops, small timeouts, so that finishing a loop before the expected timeout doesn't have much effect.

perigoso: If we could get the elapsed time from the poll we could dynamically set the next poll timeout to the "remainder" of the total timeout, but I don't think that can be easily done

whot: `clock_gettime()` should do the trick? So something like:
```c
timestamp = clock_gettime(); // note: convert to ms here too
endtime = timestamp + 1000ms;
while (timestamp < endtime) {
    poll and rest of stuff
    timestamp = clock_gettime();
}
```

perigoso: yeah, that should work, I was thinking without including time because its not used for anything else

amfern:
@FFY00 I am fine with merging this PR as it, unless you want to address some of @whot comments

FFY00:
I do.

perigoso:
@FFY00  can't request a review from you, so heres a mention

perigoso:
I wasn't able to reproduce the initial issue (the devices firmware currently only sends protocol packets, no regular mouse or keyboard), but it is working in the "non edge case"/regular state

src/libratbag-hidraw.h
```diff
@@ -183,6 +185,36 @@ int ratbag_hidraw_read_input_report(struct ratbag_device *device, uint8_t *buf,
  */
 int ratbag_hidraw_read_input_report_index(struct ratbag_device *device, uint8_t *buf, size_t len, int hidrawno);
 
+/**
+ * Read an input report from the device util a specified filter function is successfull
```
whot, line 189:
```suggestion
 * Read an input report from the device until a specified filter function is successful
```

src/libratbag-hidraw.h
```diff
@@ -183,6 +185,36 @@ int ratbag_hidraw_read_input_report(struct ratbag_device *device, uint8_t *buf,
...
+ * @param filter filter function
+ *
+ * @return count of data transfered, or a negative errno on error
+ */
+int
+ratbag_hidraw_read_input_report_filtered(struct ratbag_device *device, uint8_t *buf, size_t len,
+				 ratbagd_hidraw_filter_t filter);
+
+/**
+ * Read an input report from the device from a specific hidraw index util a specified filter function is successfull
```
whot, line 204:
```suggestion
 * Read an input report from the device from a specific hidraw index until a specified filter function is successful
```

src/libratbag-hidraw.c
```diff
@@ -1675,3 +1675,55 @@ ratbag_hidraw_read_input_report_index(struct ratbag_device *device, uint8_t *buf
...
+int
+ratbag_hidraw_read_input_report_filtered(struct ratbag_device *device, uint8_t *buf, size_t len,
+				 ratbagd_hidraw_filter_t filter)
+{
+	return ratbag_hidraw_read_input_report_index_filtered(device, buf, len, 0, filter);
+}
+
+int
+ratbag_hidraw_read_input_report_index_filtered(struct ratbag_device *device, uint8_t *buf, size_t len,
+				 int hidrawno, ratbagd_hidraw_filter_t filter)
```
whot, line 1688: `ratbag_hidraw_read_input_report_index` should be able to call this function with a `NULL` filter so we can re-use that code.

Or, even better since all of libratbag is effectively internal API, you could just change all callers of `ratbag_hidraw_read_input_report_index()` to pass NULL as filter so we only have one function.

perigoso: hm, yeah reusing this makes sense, simplifies things

perigoso:
Can someone with devices that use the steelseries driver test this?

staticssleever668:
I'm not sure whether you want mice with custom or official firmware tested as don't see any SteelSeries-specific changes, but I tried this branch with my SteelSeries Rival 100 in case it matters. It works fine, I can change LED color and resolutions. It should be noted that my mouse doesn't support reading configuration data from it and doesn't support changing button behaviors.

perigoso:
`driver-steelseries.c ` uses hidraw functions that were changed here

staticssleever668:
I put some break points to check if changed pieces of code are used on my mouse, only firmware version is used (which gives the same result), read configuration is not (who would have thought).

perigoso:
well thats good enough for me, I didnt think there would be any problems

whot:
merged, thanks

Issue #1126, opened 2021-02-01
Polling rates are not set correctly


deathtrip:
**Information**

- `ratbagd` version (`ratbagd --version`): 0.15
- Distribution: Arch Linux
- Kernel version (ex. `uname -srmo`): `Linux 5.10.11-hardened1-1-hardened x86_64 GNU/Linux`

Device info:
```
hooting-chinchilla - Logitech G Pro
Model: usb:046d:4079:0
```

**Describe the bug**
It looks like mouse polling rates are set only in steps of 100, even if given another value. 
When i try to set 125 i get 100 and 200 instead of 250. All verified with [evhz](https://git.sr.ht/~iank/evhz).
It was also mentioned [here](https://github.com/libratbag/libratbag/issues/692#issuecomment-496304983), however it's not the reporting of ratbagctl that's wrong, but the actual polling rate that's being set.

Another problem it that currently ratbagctl can set polling rates not officially supported by manufacturers, and i imagine that can lead to problems on some hardware.

**To Reproduce**
1. `ratbagctl <device> rate get-all`
2. Set polling rate below 500, eg. 250 or 125.
3. `ratbagctl <device> rate get` You should see 100 instead of 125 or 200 instead of 250.
4. Verify with evhz

PR #1157, opened 2021-04-27
Add basic support for SteelSeries Rival 100

staticssleever668:
First commit fixes report rate changing for version 1 devices.
Second commit adds support for Rival 100. As stated in the commit message, neither button nor LED configuration changing doesn't work. DPI changing works almost flawlessly (can't switch between first and second resolution), but polling rate changing doesn't work fully for now because of #1126.

Edit: forgot to mention: it doesn't support reading any data, including report rate and resolutions.

### DIFF
--- a/data/devices/steelseries-rival-100.device
+++ b/data/devices/steelseries-rival-100.device
@@ -1,0 +1,15 @@
+[Device]
+# 100; 100 (Dell China); 100 Dota 2 Edition (retail); 100 Dota 2 Edition (Lenovo); 105.
+DeviceMatch=usb:1038:1702;usb:1038:170a;usb:1038:170b;usb:1038:170c;usb:1038:1814
+Driver=steelseries
+Name=SteelSeries Rival 100/105
+
+[Driver/steelseries]
+AlternateLed=1
+# Actually 6.
+Buttons=0
+DeviceVersion=1
+DpiList=250;500;1000;1250;1500;1750;2000;4000
+Leds=1
+# Not sure what this is needed for.
+MacroLength=0
--- a/src/driver-steelseries.c
+++ b/src/driver-steelseries.c
@@ -611,10 +611,10 @@ steelseries_write_led_v1(struct ratbag_led *led)
 	uint8_t buf[STEELSERIES_REPORT_SIZE_SHORT] = {0};
 	int ret;
 
+	const int alternate_led = ratbag_device_data_steelseries_get_alternate_led(device->data);
+
 	buf[0] = STEELSERIES_ID_LED_EFFECT_SHORT;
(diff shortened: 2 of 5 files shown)


staticssleever668:
What a mess I made here... I didn't think at first that it would make no sense to force push if it's not before merging...
So, I think this is as far as I can get. I still didn't implement buttons.
To support LEDs I had to add an option to device configuration file, and I'm not sure it has a proper name please check it. Other than that, I think PR is ready for a proper review.

src/libratbag-data.c
```diff
@@ -82,6 +82,7 @@ struct data_steelseries {
 	int macro_length;
 	int mono_led;
 	int short_button;
+	int alternate_led;
```
staticssleever668, line 85: Should I move it to the beginning? Wouldn't this break API?

src/driver-steelseries.c
```diff
@@ -625,8 +651,13 @@ steelseries_write_led_v1(struct ratbag_led *led)
 			buf[2] = (led->brightness / 86) + 2;
 		}
 	} else {
-		buf[0] = STEELSERIES_ID_LED_COLOR_SHORT;
-		buf[1] = led->index + 1;
+		if (alternate_led) {
+			buf[0] = 0x05;
```
staticssleever668, line 655: All the other commands are added as constants in the beginning, should it be done the same way here? If so, should I name it "STEELSERIES_ID_LED_COLOR_SHORT_ALTERNATE" or somehow different?

src/driver-steelseries.c
```diff
@@ -443,9 +443,29 @@ steelseries_write_report_rate(struct ratbag_profile *profile)
...
+				reported_rate = 0x03;
+				break;
+			case 500:
+				reported_rate = 0x02;
+				break;
+			case 1000:
+				reported_rate = 0x01;
+				break;
+			default:
+				log_info(device->ratbag, "invalid report rate (%d); falling back to 1000 hz.\n", profile->hz);
```
staticssleever668, line 461: Should log level be changed?

data/devices/steelseries-rival-100.device
```diff
@@ -0,0 +1,15 @@
...
+
+[Driver/steelseries]
+AlternateLed=1
+# Actually 6.
+Buttons=0
+DeviceVersion=1
+DpiList=250;500;1000;1250;1500;1750;2000;4000
+Leds=1
+# Not sure what this is needed for.
+MacroLength=0
```
staticssleever668, line 15: This seems to be used when button behavior is changed, but I did not implement anything related to buttons. Should I remove this?

staticssleever668:
I will remake this patch later today taking into consideration comments I made.

PR #1159, opened 2021-04-27
sinowealth: implement poll rate changing

staticssleever668:
@phomes and @enkore, sorry for the ping, but since you both own SinoWealth-based mice, I'd be really grateful if you could check this.

NOTE: at least on G-Wolves Hati when poll rate is set to a wrong value, the mouse stops moving. I guess official software may fix this, but I'm not too sure, so proceed with caution.

### DIFF
--- a/src/driver-sinowealth.c
+++ b/src/driver-sinowealth.c
@@ -443,6 +443,24 @@ sinowealth_commit(struct ratbag_device *device)
 	int rc;
 	uint8_t dpi_enabled = 0;
 
+	/* Update report rate. */
+	uint8_t reported_rate = 0;
+	if (profile->hz >= 1000) {
+		profile->hz = 1000;
+		reported_rate = 0x04;
+	} else if (profile->hz >= 375) {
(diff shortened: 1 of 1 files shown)


src/driver-sinowealth.c
```diff
@@ -404,6 +410,30 @@ sinowealth_commit(struct ratbag_device *device)
...
+			break;
+		default:
+			log_info(device->ratbag, "invalid report rate (%d); falling back to 1000 hz.\n", profile->hz);
+			reported_rate = 0x04;
+			break;
+	}
+
+	/* Update report rate */
+	config->config &= ~0x0f;
+	config->config |= reported_rate;
```
staticssleever668, line 435: Can this be made in a more optimal way?

staticssleever668: I checked it it assembly, and it's better when it's done separately on two lines. Still, 0x0f mask should be replaced with 0b111 (0x7).

FFY00 (approved):

PR #1160, opened 2021-04-28
steelseries: fix report rate changing for version 1 devices

staticssleever668:
Judging by rivalcfg's data for Sensei Raw, Rival 100, Rival (300) and Kinzu v2 (all are version 1 devices in our data), it should have been like that from the beginning.
Confirmed on Rival 100.

### DIFF
--- a/src/driver-steelseries.c
+++ b/src/driver-steelseries.c
@@ -443,9 +443,24 @@ steelseries_write_report_rate(struct ratbag_profile *profile)
 	uint8_t buf[STEELSERIES_REPORT_SIZE] = {0};
 
 	if (device_version == 1) {
+		uint8_t reported_rate = 0;
+		if (profile->hz >= 1000) {
+			profile->hz = 1000;
+			reported_rate = 0x01;
+		} else if (profile->hz >= 375) {
+			profile->hz = 500;
+			reported_rate = 0x02;
+		} else if (profile->hz <= 125) {
+			profile->hz = 125;
+			reported_rate = 0x04;
+		} else {
+			profile->hz = 250;
+			reported_rate = 0x03;
+		}
+
 		buf_len = STEELSERIES_REPORT_SIZE_SHORT;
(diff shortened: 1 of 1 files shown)


src/driver-steelseries.c
```diff
@@ -443,9 +443,29 @@ steelseries_write_report_rate(struct ratbag_profile *profile)
...
+				reported_rate = 0x03;
+				break;
+			case 500:
+				reported_rate = 0x02;
+				break;
+			case 1000:
+				reported_rate = 0x01;
+				break;
+			default:
+				log_info(device->ratbag, "invalid report rate (%d); falling back to 1000 hz.\n", profile->hz);
```
FFY00, line 461: We should fallback to the closest here.

FFY00 (changes requested):

src/driver-steelseries.c
staticssleever668: I'm not really sure how to implement that... What about this? It falls back not to the closest one, but to the higher one though.
```c
static unsigned int REPORT_RATES_V1[] = { 125, 250, 500, 1000 };
static uint8_t REPORT_RATES_V1_ENCODED[] = { 0x04, 0x03, 0x02, 0x01};

...

char reported_rate = '\0';
for (size_t i = 0; i < ARRAY_LENGTH(REPORT_RATES_V1); i++) {
	if (profile->hz <= REPORT_RATES_V1[i]) {
		reported_rate = REPORT_RATES_V1_ENCODED[i];
		profile->hz = REPORT_RATES_V1[i];
		break;
	};
}
`

staticssleever668: Done that and rebased to master.

FFY00 (approved):
Looks fine. I don't have any supported device, @phomes can you give it a try?

staticssleever668:
Above force-push tidies the code up a bit.

FFY00 (approved):
Thanks :blush:

PR #1161, opened 2021-04-28
ratbagd-profile: modify report rate change sanity check 

### LINKED ISSUE #692: Low poll/refresh-rates not avaliable

### LINKED ISSUE #1126: Polling rates are not set correctly

staticssleever668:
Fixes #692 and closes #1126. Also unblocks https://github.com/libratbag/piper/pull/494.

### DIFF
--- a/ratbagd/ratbagd-profile.c
+++ b/ratbagd/ratbagd-profile.c
@@ -507,8 +507,8 @@ ratbagd_profile_set_report_rate(sd_bus *bus,
 		return r;
 
 	/* basic sanity check */
-	if (rate > 5000)
-		rate = 5000;
+	if (rate > 8000)
+		rate = 8000;
 
 	r = ratbag_profile_set_report_rate(profile->lib_profile, rate);
 	if (r == 0) {
--- a/ratbagd/ratbagd-profile.c
+++ b/ratbagd/ratbagd-profile.c
@@ -509,8 +509,6 @@ ratbagd_profile_set_report_rate(sd_bus *bus,
 	/* basic sanity check */
 	if (rate > 5000)
 		rate = 5000;
-	else if (rate % 100)
-		rate = rate - (rate % 100);
 
 	r = ratbag_profile_set_report_rate(profile->lib_profile, rate);
 	if (r == 0) {


stephanlachnit (approved):

ratbagd/ratbagd-profile.c
```diff
@@ -509,8 +509,6 @@ ratbagd_profile_set_report_rate(sd_bus *bus,
 	/* basic sanity check */
 	if (rate > 5000)
 		rate = 5000;
-	else if (rate % 100)
-		rate = rate - (rate % 100);
```
whot, line 512: tbh, I don't recommend this change. There are very few devices that need a report rate of e.g. 476Hz and where this happens its likely more a bug than anything. The issue from a quick skim is that you need 125Hz and 250Hz to work - just add those explicitly to the list here and keep the alignment. I doubt even rates like 2100 Hz are useful - my guess is 125, 250, 500, 1000 and from then on multiples of 1000.

ratbagd/ratbagd-profile.c
```diff
@@ -507,8 +507,8 @@ ratbagd_profile_set_report_rate(sd_bus *bus,
 		return r;
 
 	/* basic sanity check */
-	if (rate > 5000)
-		rate = 5000;
+	if (rate > 8000)
+		rate = 8000;
```
whot, line 511: Which devices has a poll rate of 8000Hz? Even 5000Hz is over the top, iirc USB only supports 1000Hz anyway and the one device I've seen in the past that advertises 2000Hz does so by using multiple event nodes and then randomly sending the events through whichever one is next in the internal fw ring buffer.

whot (commented):
Technically correct and does the job, I don't think it's a good idea this way though. Please see my comments above.

ratbagd/ratbagd-profile.c
staticssleever668: I think there is only one such mouse in the market right now - Raver Viper 8K. I don't know whether it uses a technique like the one described though.

ratbagd/ratbagd-profile.c
staticssleever668: I've breezed through all the drivers, and most supported report rates are 125, 250, 500 and 1000 hz. There were also several (or just one, don't remember) mice with 333 and 750 hz added to the list. Logitech G600 driver also defines support for 142, 166 and 200 hz. Checking for all of them would be cumbersome. If we do such such a check after all, I think we should align to 125 because it scales to all the most popular rates, though it won't work with these more rare report rates.
Although IMHO such check should be done on driver level. If there are devices that support such random report rates, why limit them?

stephanlachnit: Tbh I don't see why limiting the report rates lower than the max makes any sense for normal use. In https://github.com/libratbag/piper/issues/491 the only reason is some issues with wine. So IMHO 125Hz (apparently HID default), 250Hz (for wine), 500Hz (max for older gaming mice), 1000Hz (normal mice) and multiple of 1000 (over-the-top mice) should be enough.

staticssleever668: @stephanlachnit, I didn't quite catch that. Do you mean we should only allow 125, 250, 500 and the multiples of 1000, or we should not do the check at all?

ratbagd/ratbagd-profile.c
FFY00: I second this particular change. You can do 8kHz in USB HS, some high-end devices are doing this. I and designing an open source/hardware mice with a friend that supports this.

FFY00 (approved):
I am gonna make the call and accept this, we can revert if we find this causing any issues.

PR #1162, opened 2021-04-28
sinowealth: print firmware version when probing

staticssleever668:
Firmware version getting command ID was already there, so why not put it to use.


torrocus (approved):

enkore (approved):
Matches implementation at https://github.com/enkore/gloriousctl/blob/latest/gloriousctl.c#L451

PR #1163, opened 2021-04-28
steelseries: improve v1 breathing LED duration configuration

staticssleever668:
SteelSeries mice using V1 driver don't support changing breathing LED duration precisely, so it was always locked on 5 seconds. Now we try to match desired duration with factory preset values: 3, 5 and 7 seconds.

### DIFF
--- a/src/driver-steelseries.c
+++ b/src/driver-steelseries.c
@@ -600,7 +600,17 @@ steelseries_write_led_v1(struct ratbag_led *led)
 		buf[2] = 0x01;
 		break;
 	case RATBAG_LED_BREATHING:
-		buf[2] = 0x03; /* 0x02: slow, 0x03: medium, 0x04: fast */
+		/* 0x2/3/4 - speed (by eye it's 3, 5 and 7 seconds) */
+		if (led->ms <= 3000) {
+			led->ms = 3000;
+			buf[2] = 0x04;
+		} else if (led->ms <= 5000) {
+			led->ms = 5000;
+			buf[2] = 0x03;
+		} else {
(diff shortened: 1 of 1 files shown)


FFY00 (commented):
Hum... I agree that this is annoying, and it would be better to have *something* working rather than nothing at all. Given our API doesn't support this, I think doing something like this is reasonable, but we just overwrite `led->ms` with the value we are applying.

FFY00:
@whot any thoughts?

staticssleever668:
Actually, we weren't overwriting `led->ms` but just the value sent to mouse. This commit changes it to do so as well. NOTE: It works with ratbagctl but in Piper the value isn't updated.

whot:
Main problem (generally a problem for libratbag): the various devices have way too many different options. So when it comes to the public API, you pick what makes sense for *most* (or sometimes just: *many*) devices and go with that. It means you can't do certain things on certain devices, but the alternative is to have device-specific configuration tools for each one of them.

So if there's a device that only handles discrete steps for something we expect a more continuous range - snap to the supported steps in the driver and communicate this back to the client (i.e. Piper) so it can update the UI. But unless there's a significant number of devices that only have a few steps, I wouldn't bother having this exported through the API - it makes every client more complicated for very little benefit.

staticssleever668:
Even if such API is implemented, the driver would still need changes in this exact way (additionally it would only need to specify that it supports only these timings), so I think it's okay to merge as-is.

FFY00:
This seems good, but I would like someone else to test it.

FFY00 (approved):

PR #1165, opened 2021-04-28
Add support for G-Wolves Hati HT-M

staticssleever668:
I feel like this is some extreme spaghetti code... I'd appreciate any feedback.


src/driver-sinowealth.c
```diff
@@ -358,8 +364,22 @@ sinowealth_init_profile(struct ratbag_device *device)
...
+	rc = ratbag_hidraw_has_report(device, SINOWEALTH_REPORT_ID_CONFIG_LONG);
+	if (rc) {
+		struct sinowealth_data *drv_data = device->drv_data;
+		drv_data->is_long = true;
+
+		return rc;
+	}
+
+	return 0;
 }
```
staticssleever668, line 383: I'm not sure about this, but I think official software may do such compatibility check in a different way somewhere at the time of launching.
Here's what it does:
1) It checks firmware version (command "05 01"). My G-Wolves Hati returns "V161".
2) Gets the result of this command "05 02" (which returns "05 02 01" for me), true purpose of which I couldn't reverse engineer.
3) Gets the status of lift-off distance and angle snapping ("05 1b"). First bit of the third byte of returned value is always 1, which is suspicious.
4) Gets current configuration, finally using a different report id (6 instead of 4) and getting more data (167 bytes instead of 131).

src/driver-sinowealth.c
```diff
@@ -129,13 +131,14 @@ struct sinowealth_config_report {
...
-	uint8_t padding[SINOWEALTH_CONFIG_SIZE - SINOWEALTH_CONFIG_SIZE_USED];
+	uint8_t unknown5[36];
+	uint8_t padding[SINOWEALTH_CONFIG_SIZE - SINOWEALTH_CONFIG_SIZE_USED_LONG];
 } __attribute__((packed));
 
 _Static_assert(sizeof(struct sinowealth_config_report) == SINOWEALTH_CONFIG_SIZE, "Invalid size");
 
 struct sinowealth_data {
-	/* this is kinda unnecessary at this time, but all the other drivers do it too ;) */
+	bool is_long;
```
staticssleever668, line 141: It doesn't look pretty with current implementation. Maybe it would make more sense if instead of `is_long` we had `config_size` and `report_id_config` here?

src/driver-sinowealth.c
```diff
@@ -129,13 +131,14 @@ struct sinowealth_config_report {
 	 * 0x2 - 3 mm
 	 */
 	uint8_t lift_off_distance;
```
staticssleever668, line 133: These bytes are always 0xFF for me. I'm not sure if it works for Glorious mice, but for G-Wolves Hati it works like in the snippet in it's add request thread.

staticssleever668:
I just noticed it has the same PID as Dream Machines DM5 in #1065. Fortunately, I wrote the report ID compatibility check in a way that won't need an entry in the data file, but that's something to keep in mind.

src/driver-sinowealth.c
```diff
@@ -78,6 +78,12 @@ struct sinowealth_config_report {
 	 * CONFIG_SIZE_USED-8 - write
 	 */
 	uint8_t config_write;
+	/*
+	 * ss668:
+	 * For me it's these values: {0, 0, 0, 0, 0x64, 0x6}.
+	 * Wild guess: last two could be DPI step and button count. My mouse has
+	 * no LEDs, so previous values could indicate LED count.
```
staticssleever668, line 85: Please, check this.

src/driver-sinowealth.c
```diff
@@ -65,17 +67,29 @@ enum rgb_effect {
...
+	 * CONFIG_SIZE_USED-8 - write.
 	 */
 	uint8_t config_write;
+	/*
+	 * ss668:
+	 * For me it's these values: {0, 0, 0, 0, 0x64, 0x6}.
+	 * Wild guess: last two could be DPI step and button count. ~~My mouse has
+	 * no LEDs, so previous values could indicate LED count~~.
+	 * Update: These values are the same on Glorious Model O. Now need to check
+	 * on Dream Machines DM5.
```
staticssleever668, line 91: Please, check this on Dream Machines DM5.

perigoso:
did you mean to close this @staticssleever668 ?

staticssleever668:
Hi, yes, I did. I added an explanation as to why in pull request's description.

Issue #990, opened 2020-05-21
0.14: gcc compile time warnings


kloczek:
```
../src/hidpp-generic.c: In function 'hidpp_log_buffer':
../src/hidpp-generic.c:377:2: warning: ignoring return value of 'asprintf' declared with attribute 'warn_unused_result' [-Wunused-result]
  377 |  asprintf(&output_buf, "%s %s", header ? header : "", bytes);
      |  ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
../src/hidpp10.c: In function 'hidpp10_profile_parse_names':
../src/hidpp10.c:1418:23: warning: taking address of packed member of 'struct <anonymous>' may result in an unaligned pointer value [-Waddress-of-packed-member]
 1418 |        metadata->lgs02.name,
      |        ~~~~~~~~~~~~~~~^~~~~
../src/hidpp10.c:1423:36: warning: taking address of packed member of 'struct <anonymous>' may result in an unaligned pointer value [-Waddress-of-packed-member]
 1423 |         metadata->lgs02.macro_names[i],
      |         ~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~
../src/hidpp10.c: In function 'hidpp10_profile_set_names':
../src/hidpp10.c:1445:43: warning: taking address of packed member of 'struct <anonymous>' may result in an unaligned pointer value [-Waddress-of-packed-member]
 1445 |  hidpp10_uchar8_to_uchar16(metadata->lgs02.name,
      |                            ~~~~~~~~~~~~~~~^~~~~
../src/hidpp10.c:1449:56: warning: taking address of packed member of 'struct <anonymous>' may result in an unaligned pointer value [-Waddress-of-packed-member]
 1449 |   hidpp10_uchar8_to_uchar16(metadata->lgs02.macro_names[i],
      |                             ~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~
hidpp@sha/hidpp.c: In function '_wrap_hidpp20_led_sw_ctrl_led_state_breathing_get':
hidpp@sha/hidpp.c:6824:26: warning: taking address of packed member of 'struct hidpp20_led_sw_ctrl_led_state' may result in an unaligned pointer value [-Waddress-of-packed-member]
 6824 |   result = (_breathing *)& ((arg1)->breathing);
      |                          ^~~~~~~~~~~~~~~~~~~~~
[...]

PR #1174, opened 2021-05-08
Fix some compile time warnings

staticssleever668:
There are still some warnings, however. Here they are: Clang and GCC's `unused-function` in `sinowealth` driver - will be fixed in #1065; Clang's `incompatible-pointer-types-discards-qualifiers`/GCC's `discarded-qualifiers` in `hidpp10` driver - it this would need an API change, as I understand; GCC's `address-of-packed-member` - don't know how to fix it properly. More details in #990.

### DIFF
--- a/src/driver-hidpp10.c
+++ b/src/driver-hidpp10.c
@@ -548,7 +548,7 @@ hidpp10_log(void *userdata, enum hidpp_log_priority priority, const char *format
 {
 	struct ratbag_device *device = userdata;
 
-	log_msg_va(device->ratbag, priority, format, args);
+	log_msg_va(device->ratbag, (enum ratbag_log_priority)priority, format, args);
 }
 
 static int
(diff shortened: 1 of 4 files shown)


src/driver-hidpp10.c
```diff
@@ -448,7 +448,7 @@ hidpp10drv_read_profile(struct ratbag_profile *profile)
 		free(profile->name);
 		profile->name = NULL;
 	}
-	if (p.name && p.name[0] != '\0')
+	if (*p.name && p.name[0] != '\0')
```
FFY00, line 451: Shouldn't this be `p.name != NULL`?

staticssleever668: Null check is always implied.

FFY00: That seems a weird to me, but if `p.name` is indeed never null, then we can just remove the `p.name` check. Replacing `p.name` with `*p.name` will just instead check the first character of the name.

staticssleever668: Hi, you are right. I removed the check.

staticssleever668:
Rebased on master and given commits better names for more clarity.

PR #1209, opened 2021-10-09
data: support steelseries rival variations

amfern:
Based on the work done in [rivalcfg](https://github.com/flozz/rivalcfg/blob/master/rivalcfg/devices/rival300.py), steel series rival 300 have many variations of product_ids. This PR adds them all to `steelseries-rival.device`


perigoso:
I think it makes sense to have them together, it's really all instances of the same device, afaik it's just visual

PR #1210, opened 2021-10-10
steelseries: set report_id of 0x00 

### LINKED ISSUE #804: Rival 300 isn't supported
It's really odd that you don't have rival 300 in supported devices list since there is another non gui program called rivalcfg that does support. I'm just suggesting that you should add support for this mouse as well because I would like to see it working with Piper so that I or future users of this mice won't have to use non-gui alternatives.

amfern:
Set 0x00 report_id for every steelseries device, similar to `rivalcfg project`

fixes #804

### DIFF
--- a/src/driver-steelseries.c
+++ b/src/driver-steelseries.c
@@ -42,6 +42,7 @@
 #define STEELSERIES_INPUT_HIDRAW	1
 
 /* not sure these two are used for */
+#define STEELSERIES_REPORT_ID			0x00 // steelseries doesn't use numbered reports
 #define STEELSERIES_REPORT_ID_1			0x01
 #define STEELSERIES_REPORT_ID_2			0x02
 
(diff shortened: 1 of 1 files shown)


amfern:
I decided to stick to a similar implication of hidhpp2, openinput and libusb/hidapi and pass the buffer instead of altering ratbag_hidraw_output_report, function.

amfern:
Also, removed the old firmware version check as it complicates the code and I don't see rivalcfg addressing older firmware either

klausenbusk:
I tested https://github.com/libratbag/libratbag/pull/1210/commits/85fee87b0fea42d1e9df8d708150f94f0009f90e with my SteelSeries Rival 300 (1038:1710).

Tested:
* Setting DPI (`ratbagctl singing-gundi dpi set 6500`, `ratbagctl singing-gundi resolution 0 dpi set 6500` and `ratbagctl singing-gundi resolution 0 dpi set 6500`)
* Setting LED mode `on` and `breathing`. `off` is also supported according to `ratbagctl singing-gundi led 0 capabilities` but setting it doesn't disable the LED
  * `ratbagctl singing-gundi led 0 set color 000000` does disable the LED tho, so maybe that should be wired to the `off` mode?
* Setting LED color (`ratbagctl singing-gundi led 0 set color ff0000` and `ratbagctl singing-gundi led 1 set color ff0000`)
* Setting LED duration (`ratbagctl singing-gundi led 0 set duration 10000` and `ratbagctl singing-gundi led 1 set duration 10000`)
* Setting LED brightness doesn't work (`ratbagctl singing-gundi led 0 set brightness 100` and `ratbagctl singing-gundi led 1 set brightness 100`)
* Swapping the left and right mouse button

perigoso:
Everything behaved the same as before, so this patch is fine.

I, like @klausenbusk, and like i experienced before, had issues setting the led modes to off, i ruled out hardware, and setting the color works fine the whole range, but when i set the leds to off sets blue to max and everything else off, not sure what the cause here is, but maybe mapping off to color 000000 is a good option.

amfern:
Thank you very much for testing this patch, I think at this point we will open a different issue for setting LED off and LED brightness.

amfern:
Hi, @FFY00.
Can you have another look at this PR?

amfern:
I've opened a follow-up issue for the LED off bug #1223 , and merging this PR as it seem to work for other people

Issue #1098, opened 2020-11-20
Add support for G915 TKL **WHITE**


kevinkga:
**Is your feature request related to a problem? Please describe.**
The G915 TKL is currently supported by libratbag. However, the more recent white version uses a different device identifier which is not recognised by libratbag.

**Describe the solution you'd like**
Please add the following device identifier to the g915-tkl .device file
[Device]
Name=Logitech G915 TKL
DeviceMatch=usb:046d:c343;**usb:046d:c545**;bluetooth:046d:b35f
LedTypes=logo;switches
Driver=hidpp20

[Driver/hidpp20]
DeviceIndex=1

**Describe alternatives you've considered**
none

**Additional context**
n/a

ScottGarman:
FWIW I just bought a new black version of the G915 TKL and its receiver also has an id of c545.

I will say though after making this modification and restarting ratbagd, ratbagctl is listing it as a "Logitech USB Receiver" for some reason. Piper seems to be aware that it's a keyboard and shows configuration options for Buttons and LEDs, but trying to change my LED settings isn't working. I realize piper is a separate project but wanted to report this in case it indicates anything useful.

My platform is Ubuntu 20.04 running the PPA package ratbagd 0.15-1+git1647-b3275e1~ubuntu20.04.1.

joggelxx:
Same for me. Just got a new, black G915 TKL also with id c545. After adding the respective device file ratbagd now detects it as "Logitech USB Receiver". It's listed with the same name in Piper and setting the LEDs works fine for me, but when trying to change the custom keys it looks like it's considered as mouse because I get options to configure left/right handed and so on. It's true that this might also be an issue with Piper, but at least the generic name and the additional device id should be handled in ratbag.

I'm running Manjaro with libratbag 0.15-1 and Piper 0.5.1-2.

amfern:
I think this issue is related to #1178, can someone from this thread with try the instructions in the comment https://github.com/libratbag/libratbag/issues/1178#issuecomment-1046331230

PR #1219, opened 2021-10-26
data: add additional USB id for Logitech G915 TKL

ScottGarman:
As discussed in Issue #1098, Logitech appears to have released both
black and white editions of this keyboard with another USB id. This adds
support for those units.


amfern:
As disused later in the #1098, the id `c545` is a receiver and this cannot be merged, I am sorry for the inconvenience

PR #1220, opened 2021-11-03
Updated to support different model

Saksey1:
currently not working for my Pro X Superlight,  So I just added the ID for it


perigoso:
Sorry, but the ID you added is the ID of the receiver, we can't merge this because a different device might use the same receiver, so same ID, causing a collision.
This needs to be supported at the kernel level, I believe there's a PR for that but is stalled, would need to look into it.

That said, you can use this `.device` locally to use your device with libratbad.

See #1118

PR #1225, opened 2021-11-06
tests: test for additonal lightspeed receviers pids

amfern:
this PR adds test for receiver used by logitech g915 keyboard https://github.com/libratbag/libratbag/issues/1098#issuecomment-952428307


perigoso (approved):

FFY00 (approved):
This is fine, though we could actually just fail on IDs starting with `c5`.

PR #1234, opened 2021-11-25
Add Logitech G502 Lightspeed mouse

teemuki3:


perigoso:
Hey, sorry, the PID 0xc539 is a receiver, and thus we cannot include it.
The syntax is also wrong:
`usb:usb=046d:407f;046d:c539` should be `usb:046d:407f;usb:046d:c539`

Also also, the device you are trying to add seems to be supported already:
https://github.com/libratbag/libratbag/blob/3f99446a1be984b78eac60e47bfd66e1c674e30c/data/devices/logitech-g502-hero-wireless.device

Does it not show up in ratbagd or piper?

teemuki3:
Indeed, it seems that g502 hero wireless is the same device as g502 lightspeed. My mouse wasn't properly recognized when I tried the first time. After adding the new device file, it started working, but perhaps it was just that this caused ratbag to restart and then recognize my mouse properly.

Anyway, everything works now as it should.

PR #1238, opened 2021-12-03
Add a GitHub Workflow as CI run

whot:
CircleCI is limiting how many runs we can do, which results in pipelines
randomly failing when we've used up the credits for this month.

Let's add a github workflow instead. Unfortunately github workflows are
a bit of a toy api, no support for YAML anchors or extending other jobs,
so there's a lot of duplication needed to get a job to do something.

The actions added in this patch help a bit to make the jobs easier to
understand (copied from libwacom).

This only runs on Ubuntu right now.

### DIFF
--- a/.github/actions/meson/action.yml
+++ b/.github/actions/meson/action.yml
@@ -1,0 +1,41 @@
+# A Github Action to build a repository with meson
+name: "Setup and run meson"
+description: "Setup and run meson"
+inputs:
+  builddir:
+    description: "The build directory"
+    default: "builddir"
+  srcdir:
+    description: "The source directory"
+    default: "."
+  meson_args:
+    description: "Arguments passed to meson setup"
+    default: ""
+  meson_skip_test:
+    description: "Set to a nonempty string to skip meson test"
+    default: ""
+  meson_test_args:
+    description: "Arguments passed to meson test"
+    default: ""
+  meson_precmd:
+    description: "Set to 'sudo' to run with sudo"
+    default: ""
+  ninja_args:
+    description: "Arguments passed to ninja"
+    default: ""
+  ninja_precmd:
+    description: "Set to 'sudo' to run with sudo"
+    default: ""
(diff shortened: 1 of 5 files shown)


whot:
Sigh, github doesn't run the workflow on a MR unless there's already a workflow file, so this will simply not run. The result for this commit is in my personal fork though: https://github.com/whot/libratbag/actions/runs/1534040822

perigoso:
LGTM

perigoso (approved):

.github/actions/meson/action.yml
```diff
@@ -0,0 +1,41 @@
...
+    description: "Arguments passed to ninja"
+    default: ""
+  ninja_precmd:
+    description: "Set to 'sudo' to run with sudo"
+    default: ""
+runs:
+  using: "composite"
+  steps:
+    - run: |
+        pushd ${{inputs.srcdir}}
```
FFY00, line 33: This is not needed, you can do `meson setup srcdir builddir` instead.

Eg. https://github.com/FFY00/mesonpy/blob/80563719cfafa11b059bb36fb56d2287f8d6e4e3/mesonpy.py#L436-L440

FFY00 (approved):

.github/actions/meson/action.yml
whot: good point, thanks. Fixed in 9bc2c60 but I'll squash that in before merging.

side-note: `meson setup builddir srcdir`, the two are swapped.

whot:
Minor note, I've duplicated the actions to this repo. We could use libwacom's actions directly with
```
- uses: linuxwacom/libwacom/.github/actions/pkginstall@master
```
I'm a bit unsure whether we should since I don't want to have to maintain those actions with a public API. But I guess we could just use `@sha` instead of `@master` so we don't have to worry about breakage? Not sure yet, any opinions?

PR #1242, opened 2021-12-10
Use meson's install_subdir instead of naming data files individually

whot:
I can't remember why we didn't do this in the first place but not requiring every `.device` file to be listed in `meson.build` seems like a win.


perigoso:
makes sense to me

perigoso (approved):

perigoso (changes requested):
maybe bump the meson required version? seemingly we are using features that require 0.50, but we only state 0.40 as required, this PR also requires >= 0.42

```
 * 0.42.0: {'exclude_files arg in install_subdir'}
 * 0.46.0: {'Python Module'}
 * 0.50.0: {'install arg in configure_file'}
```

amfern (approved):

whot:
> maybe bump the meson required version? 

Done in #1245 now, we can merge that one first and rebase this one so the history is pretty

perigoso (approved):

PR #1246, opened 2021-12-13
Drop CircleCI

whot:
Right now, CircleCI fails with ssh errors because for some reason it
wants to clone the repo over ssh, not https. This is probably fixable
but CircleCI now limits how many runs we can do per month anyway so it's
borderline useless if we do actual developement.

Having to re-learn the Circle CI syntax to fix the current issue is not
an efficient use of our time, so let's just drop this completely, git
will keep the history of what we used to do in the CI for us.

02f3487 added a github workflow which
is comparatively incomplete but could be extended to do the previous
test runs. That's a more future-proof path anyway.


perigoso (approved):
Sounds reasonable, the new workflow is just not as comprehensive correct? but it works

perigoso:
ah! we have a build status badge on the readme that links to circleci, we should remove that too...
I don't if there are badges for github workflows?

perigoso:
here:
it uses the wf name, so we might want something more explicit
linux

whot:
> ah! we have a build status badge on the readme that links to circleci, we should remove that too...

yeah, and the webhook and some other settings are still there but they can be removed after the fact.

I've removed the status from the README now but, call me grump and/or old, I never saw the point of those. We shouldn't ever merge patches that break the CI anyway and does anyone ever check the README for an actual status?

perigoso:
I like them, but i agree that they aren't important, especially in this case with a protected branch where only working code is expected to be merged

perigoso:
An example for an useful badge here, a bug counter, we have issue count up there on the tab but we have a lot of feature requests so you'd need to go there and manually set up the filter.

Anyway, LGTM

whot:
Yeah, tbh I'm not disagreeing if people want those badges, I just forget they exist since I barely ever look at the README rendering itself :) So feel free to add those.

Thanks for the review, much appreciated

whot:
I've merged this now and deleted the circleci webhook. This should at least make future pipelines succeed.

PR #1249, opened 2021-12-21
Add support for Logitech G303 Shroud Wireless

aw1cks:
Simply needed a device file. All looks to be working well with my device.

### DIFF
--- a/data/devices/logitech-g303-shroud-wireless.device
+++ b/data/devices/logitech-g303-shroud-wireless.device
@@ -1,0 +1,8 @@
+[Device]
+Name=Logitech G303 Shroud Wireless
+DeviceMatch=usb:046d:c547;usb:046d:c097
+LedTypes=battery;
+Driver=hidpp20
+
+[Driver/hidpp20]
+DeviceIndex=1


perigoso:
Sorry, is the `eviceMatch=usb:046d:c547;` required? that is the receiver ID, we can't merge that as it would cause conflicts with other devices with that same receiver

Would it be possible to add this as an extra entry to the g303 instead of a new `.device` file?

aw1cks:
Wouldn't omitting the receiver ID mean that the mouse would not work unless directly wired in? Unless I'm missing something here. When I remove the receiver ID, that's certainly the behavior that I see.
As to the second point, that does sound sensible, however I'm not sure how I could add the following _only_ for the wireless version seeing as they both use the same driver? Is there some extra option here that could help?
```
[Driver/hidpp20]
DeviceIndex=1
```

perigoso:
They kernel should be dealing with the receiver and exporting the correct device, @FFY00 should know more about this

data/devices/logitech-g303-shroud-wireless.device
```diff
@@ -0,0 +1,8 @@
+[Device]
+Name=Logitech G303 Shroud Wireless
+DeviceMatch=usb:046d:c547;usb:046d:c097
```
whot, line 3:
```suggestion
DeviceMatch=usb:046d:c097
```

whot (changes requested):
Let's drop the receiver ID for now, the kernel should magically handle this anyway, thanks.

data/devices/logitech-g303-shroud-wireless.device
whot: I was hoping I could just apply this and merge but ... apparently I can't. Please apply this locally, force-push and we can getr this merged. Thanks

aw1cks: Sorry I missed this! Just to clarify, in my testing that did NOT work.

whot: weird. we can't merge the c547 ID though because that one is shared across many devices. the CI *should* have picked this up, we have a test specifically for that, not sure why it didn't pick up on it.

Either way, the kernel is supposed to handle this and expose the device itself with the correct PID - apparently that doesn't work for these ones (yet? are you on a recent-enough kernel)?

staticssleever668:
Superseded by #1376.

Issue #925, opened 2020-02-18
Support for ASUS ROG Devices (e.g. Gladius 2, Gladius 2 Wireless,...)


SuperMeowBro:
It would be very cool, if support for ASUS ROG Devices like Gladius 2 or Gladius 2 Wireless could be added. There is an driver project on github (https://github.com/kyokenn/rogdrv), that could help to implement.

FFY00:
Yes, sure. @kyokenn what do you think? We can't do it for now as we don't have the hardware.

kyokenn:
I'm just a Python developer. I'm not that good at C/C++ and I'm not familiar with the API. So I would need some help with it, some debug tools for the library and some dummy driver where I could start.

stephanlachnit:
The driver with the most features is the [hidpp20](https://github.com/libratbag/libratbag/blob/master/src/driver-hidpp20.c) driver, but also quite complicated I guess. The easiest one is probably the [etekcity](https://github.com/libratbag/libratbag/blob/master/src/driver-etekcity.c) driver, albeit without LED support.

libratbag has a [`ratbag_device` class](https://github.com/libratbag/libratbag/blob/master/src/libratbag.c#L186), and more important a [`ratbag_profile` class](https://github.com/libratbag/libratbag/blob/master/src/libratbag.c#L694), which should include all informations about dpi, buttons, leds, etc. The driver should be able to read these informations from the device, and should be able to commit the set of these variables on the device if wished by the user. It's not easy, but maybe you can fiddle around with the etekcity driver and we can help if you have a problem.

FFY00:
I wonder if it's not better to get the devices emulated in https://github.com/libratbag/ratbag-emu and then write the driver.

evrifaessa:
I own a ROG Gladius ii Core and can post any needed diagnostic data you need about it.

- Distribution: Pop!_OS 21.10

- `uname -srmo`
Linux 5.13.0-7619-generic x86_64 GNU/Linux

- `ratbagd --version`
0.15

- `sudo ratbagd --verbose=raw`
```
Initializing libratbag
ratbag debug: New device: Razer Razer USB Audio Controller
ratbag debug: Using data directory '/usr/share/libratbag'
ratbag debug: No data file found for 1532:0520
ratbag debug: New device: ASUSTeK ROG GLADIUS II CORE
ratbag debug: Using data directory '/usr/share/libratbag'
ratbag debug: No data file found for 0b05:18dd
[...]

Issue #632, opened 2019-02-20
Asus rog gladius II and Cooler master masterkeys pro L support ?


startas:
Does this have support for mentioned mouse and keyboard ?

bentiss:
If the USB ids are not in https://github.com/libratbag/libratbag/tree/master/data/devices then I guess not.
We need someone to reverse engineer the protocol, but we are all struggling with the currently supported devices.

Closing this now, but feel free to reopen if you know someone interested in adding the support for those devices.

PR #1250, opened 2021-12-23
Add ASUS driver and fix ratbag-command led setting

kyokenn:
Fixes https://github.com/libratbag/libratbag/issues/632 ("Asus ROG Gladius II" part)
Fixes https://github.com/libratbag/libratbag/issues/925 (some mice may needed .device files, the protocol is almost the same between them)

### DIFF
--- a/src/asus.c
+++ b/src/asus.c
@@ -21,13 +21,14 @@
  * DEALINGS IN THE SOFTWARE.
  */
 
-#include <assert.h>
-
 #include "config.h"
-#include "libratbag-private.h"
-#include "libratbag-data.h"
 
 #include "asus.h"
+
+#include <assert.h>
+
+#include "libratbag-data.h"
+#include "libratbag-private.h"
 
 /* ASUS commands */
 #define ASUS_CMD_GET_LED_DATA		0x0312  /* get all LEDs */
--- a/src/asus.c
+++ b/src/asus.c
@@ -1,0 +1,495 @@
+/*
+ * Copyright (C) 2021 Kyoken, kyoken@kyoken.ninja
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a
+ * copy of this software and associated documentation files (the "Software"),
+ * to deal in the Software without restriction, including without limitation
+ * the rights to use, copy, modify, merge, publish, distribute, sublicense,
+ * and/or sell copies of the Software, and to permit persons to whom the
+ * Software is furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice (including the next
+ * paragraph) shall be included in all copies or substantial portions of the
+ * Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.  IN NO EVENT SHALL
+ * THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+ * FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+ * DEALINGS IN THE SOFTWARE.
+ */
+
+#include <assert.h>
+
+#include "config.h"
+#include "libratbag-private.h"
+#include "libratbag-data.h"
+
+#include "asus.h"
+
+/* ASUS commands */
+#define ASUS_CMD_GET_LED_DATA		0x0312  /* get all LEDs */
+#define ASUS_CMD_GET_SETTINGS		0x0412  /* dpi, rate, button response, angle snapping */
+#define ASUS_CMD_GET_BUTTON_DATA	0x0512  /* get all buttons */
+#define ASUS_CMD_GET_PROFILE_DATA	0x0012  /* get current profile info */
+#define ASUS_CMD_SET_LED		0x2851  /* set single led */
+#define ASUS_CMD_SET_SETTING		0x3151  /* dpi / rate / button response / angle snapping */
+#define ASUS_CMD_SET_BUTTON		0x2151  /* set single button */
+#define ASUS_CMD_SET_PROFILE		0x0250  /* switch profile */
+#define ASUS_CMD_SAVE			0x0350  /* save settings */
+
+/* fields order in _asus_dpiX_data, used for setting with ASUS_CMD_SET_SETTING */
+#define ASUS_FIELD_RATE	0
+#define ASUS_FIELD_RESPONSE	1
+#define ASUS_FIELD_SNAPPING	2
+
+/* order is mandatory because index is mapped from ButtonMapping configuration property */
+static struct asus_button ASUS_BUTTON_MAPPING[] = {
+	{ 0xf0, RATBAG_BUTTON_ACTION_TYPE_BUTTON, 1, 0 },  /* left */
+	{ 0xf1, RATBAG_BUTTON_ACTION_TYPE_BUTTON, 2, 0 },  /* right (button 3 in xev) */
+	{ 0xf2, RATBAG_BUTTON_ACTION_TYPE_BUTTON, 3, 0 },  /* middle (button 2 in xev) */
+	{ 0xe4, RATBAG_BUTTON_ACTION_TYPE_BUTTON, 4, 0 },  /* backward (on left side) */
+	{ 0xe5, RATBAG_BUTTON_ACTION_TYPE_BUTTON, 5, 0 },  /* forward (on left side) */
+	{ 0xe6, RATBAG_BUTTON_ACTION_TYPE_SPECIAL, 0, RATBAG_BUTTON_ACTION_SPECIAL_RESOLUTION_CYCLE_UP },
+	{ 0xe8, RATBAG_BUTTON_ACTION_TYPE_SPECIAL, 0, RATBAG_BUTTON_ACTION_SPECIAL_WHEEL_UP },
+	{ 0xe9, RATBAG_BUTTON_ACTION_TYPE_SPECIAL, 0, RATBAG_BUTTON_ACTION_SPECIAL_WHEEL_DOWN },
+	{ 0xe1, RATBAG_BUTTON_ACTION_TYPE_BUTTON, 4, 0 },  /* backward on right side */
+	{ 0xe2, RATBAG_BUTTON_ACTION_TYPE_BUTTON, 5, 0 },  /* forward on right side */
+	{ 0xff, RATBAG_BUTTON_ACTION_TYPE_NONE, 0, 0 },  /* placeholder */
+	{ 0xff, RATBAG_BUTTON_ACTION_TYPE_NONE, 0, 0 },  /* placeholder */
+	{ 0xff, RATBAG_BUTTON_ACTION_TYPE_NONE, 0, 0 },  /* placeholder */
+	{ 0xff, RATBAG_BUTTON_ACTION_TYPE_NONE, 0, 0 },  /* placeholder */
+	{ 0xff, RATBAG_BUTTON_ACTION_TYPE_NONE, 0, 0 },  /* placeholder */
+	{ 0xff, RATBAG_BUTTON_ACTION_TYPE_NONE, 0, 0 },  /* placeholder */
+};
+
+/* key mapping, the index is actual ASUS code */
+static unsigned char ASUS_KEY_MAPPING[] = {
+/* 00 */	0,		0,		0,		0,
+/* 04 */	KEY_A,		KEY_B,		KEY_C,		KEY_D,
+/* 08 */	KEY_E,		KEY_F,		KEY_G,		KEY_H,
+/* 0C */	KEY_I,		KEY_J,		KEY_K,		KEY_L,
+/* 0E */	KEY_M,		KEY_N,		KEY_O,		KEY_P,
+/* 14 */	KEY_Q,		KEY_R,		KEY_S,		KEY_T,
+/* 18 */	KEY_U,		KEY_V,		KEY_W,		KEY_X,
+/* 1C */	KEY_Y,		KEY_Z,		KEY_1,		KEY_2,
+/* 1E */	KEY_3,		KEY_4,		KEY_5,		KEY_6,
+/* 24 */	KEY_7,		KEY_8,		KEY_9,		KEY_0,
+/* 28 */	KEY_ENTER,	KEY_ESC,	KEY_BACKSPACE,	KEY_TAB,
+/* 2C */	KEY_SPACE,	KEY_MINUS,	KEY_KPPLUS,	0,
+/* 2E */	0,		0,		0,		0,
+/* 34 */	0,		KEY_GRAVE,	KEY_EQUAL,	0,
+/* 38 */	KEY_SLASH,	0,		KEY_F1,		KEY_F2,
+/* 3C */	KEY_F3,		KEY_F4,		KEY_F5,		KEY_F6,
+/* 3E */	KEY_F7,		KEY_F8,		KEY_F9,		KEY_F10,
+/* 44 */	KEY_F11,	KEY_F12,	0,		0,
+/* 48 */	0,		0,		KEY_HOME,	KEY_PAGEUP,
+/* 4C */	KEY_DELETE,	0,		KEY_PAGEDOWN,	KEY_RIGHT,
+/* 4E */	KEY_LEFT,	KEY_DOWN,	KEY_UP,		0,
+/* 54 */	0,		0,		0,		0,
+/* 58 */	0,		KEY_KP1,	KEY_KP2,	KEY_KP3,
+/* 5C */	KEY_KP4,	KEY_KP5,	KEY_KP6,	KEY_KP7,
+/* 5E */	KEY_KP8,	KEY_KP9,	0,
+};
+
+static unsigned int ASUS_LED_TYPE[] = {
+	RATBAG_LED_TYPE_LOGO,  /* Mouse logo */
+	RATBAG_LED_TYPE_WHEEL,  /* Mouse wheel */
+	RATBAG_LED_TYPE_SIDE,  /* Bottom of the mouse */
+};
+
+static unsigned int ASUS_POLLING_RATES[] = { 125, 250, 500, 1000 };
+
+/* search for ASUS button by ratbag types */
+struct asus_button *
+asus_find_button_by_action(struct ratbag_button_action action)
+{
+	struct asus_button *asus_button;
+	for (unsigned int i = 0; i < ARRAY_LENGTH(ASUS_BUTTON_MAPPING); i++) {
+		asus_button = &ASUS_BUTTON_MAPPING[i];
+		if ((action.type == RATBAG_BUTTON_ACTION_TYPE_BUTTON && asus_button->button == action.action.button) ||
+				(action.type == RATBAG_BUTTON_ACTION_TYPE_SPECIAL && asus_button->special == action.action.special))
+			return asus_button;
+	}
+	return NULL;
+}
+
+/* search for ASUS button by ASUS button code */
+struct asus_button *
+asus_find_button_by_code(uint8_t asus_code)
+{
+	for (unsigned int i = 0; i < ARRAY_LENGTH(ASUS_BUTTON_MAPPING); i++) {
+		if (ASUS_BUTTON_MAPPING[i].asus_code == asus_code)
+			return &ASUS_BUTTON_MAPPING[i];
+	}
+	return NULL;
+}
+
+/* search for ASUS key code by Linux key code */
+int
+asus_find_key_code(unsigned int linux_code)
+{
+	for (unsigned int i = 0; i < ARRAY_LENGTH(ASUS_KEY_MAPPING); i++) {
+		if (ASUS_KEY_MAPPING[i] == linux_code)
+			return i;
+	}
+	return -1;
+}
+
+/* convert ASUS key code to Linux key code */
+int
+asus_get_linux_key_code(uint8_t asus_code) {
+	assert(asus_code <= ARRAY_LENGTH(ASUS_KEY_MAPPING));
+	return ASUS_KEY_MAPPING[asus_code];
+}
+
+int
+asus_query(struct ratbag_device *device,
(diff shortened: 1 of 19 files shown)


kyokenn:
Also fixes https://github.com/libratbag/piper/issues/699

whot (changes requested):
Thanks for adding a new driver, always happy to see new devices being supported. There's a bunch of general issues to fix to make the code more readable and more similar to the existing code we have. In particular, the indentation appears to be all messed up, please double-check that.

I've commented on the various issues above, I do strongly suggest that you start using the struct/union approach.

src/driver-asus.c
```diff
@@ -0,0 +1,447 @@
...
+
+static int
+asus_driver_load_profiles(struct ratbag_device *device)
+{
+    int asus_index, error;
+    unsigned int profile_id, profile_count, led_count;
+    struct asus_profile_data *profile_data;
+    struct asus_button_data *button_data;
+    struct asus_resolution_data *resolution_data;
+    struct asus_led_data *led_data;
```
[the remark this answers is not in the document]
kyokenn: Do you want them in drv_data ?

src/asus.c
```diff
@@ -0,0 +1,459 @@
...
+	offset += 1;
+
+	data->led[i].r = response[offset];
+	offset += 1;
+
+	data->led[i].g = response[offset];
+	offset += 1;
+
+	data->led[i].b = response[offset];
+	offset += 1;
```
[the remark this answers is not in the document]
kyokenn: Do you want to automatically deserialize the entire packet into the struct?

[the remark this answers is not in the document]
kyokenn: Done. Migrated to struct with a union.

src/asus.c
```diff
@@ -0,0 +1,459 @@
...
+	return r;
+
+    return 0;
+}
+
+int
+asus_get_profile_data(struct ratbag_device *device, struct asus_profile_data *data)
+{
+    int r;
+    uint8_t request[64];
```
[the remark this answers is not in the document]
kyokenn: Migrated to struct with a union.

src/asus.c
```diff
@@ -0,0 +1,459 @@
...
+
+int
+asus_save_profile(struct ratbag_device *device)
+{
+    int r;
+    uint8_t request[ASUS_PACKET_SIZE];
+    uint8_t response[ASUS_PACKET_SIZE];
+    asus_prepare(request, response);
+    request[0] = 0x50;
+    request[1] = 0x03;
```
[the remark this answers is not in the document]
kyokenn: Migrated to struct with a union.

src/asus.c
```diff
@@ -0,0 +1,459 @@
...
+asus_prepare(uint8_t request[], uint8_t response[])
+{
+    for (int i = 0; i < ASUS_PACKET_SIZE; i++) {
+	request[i] = 0x00;
+	response[i] = 0x00;
+    }
+}
+
+int
+asus_query(struct ratbag_device *device, uint8_t request[], uint8_t response[])
```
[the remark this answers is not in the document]
kyokenn: Migrated to struct with a union.

src/asus.c
```diff
@@ -0,0 +1,459 @@
...
+    return -1;
+}
+
+void
+asus_prepare(uint8_t request[], uint8_t response[])
+{
+    for (int i = 0; i < ASUS_PACKET_SIZE; i++) {
+	request[i] = 0x00;
+	response[i] = 0x00;
+    }
```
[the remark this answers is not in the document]
kyokenn: Migrated to struct with a union.

src/driver-asus.c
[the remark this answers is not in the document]
whot: shouldn't be needed here, you can do this instead:
```c

struct asus_led_data led_data;
....
error = asus_get_led_data(device, &led_data);
```

src/asus-common.h
```diff
@@ -0,0 +1,246 @@
...
+struct _asus_led {
+	uint8_t mode;
+	uint8_t brightness;  // 0-4
+	uint8_t r;
+	uint8_t g;
+	uint8_t b;
+} __attribute__((packed));
+
+struct _asus_led_data {
+	uint8_t filler0[4];
```
whot, line 97:
```suggestion
	uint32_t pad;
```

src/asus-common.h
```diff
@@ -0,0 +1,246 @@
...
+union asus_led_data {
+	struct _asus_led_data data;
+	uint8_t raw[sizeof(struct _asus_led_data)];
+};
+
+// drv data
+
+struct asus_data {
+	uint8_t is_ready;
+} __attribute__((packed));
```
whot, line 110: this doesn't need to be `packed` since it's just a software struct. And no need for this to be in the header, this is a driver-only data and can move to `asus-driver.c` (afaict)

src/asus-defines.h
```diff
@@ -0,0 +1,23 @@
...
+// maximum number of buttons across all ASUS devices
+#define ASUS_MAX_NUM_BUTTON 16
+
+// maximum number of DPI presets across all ASUS devices
+// for 4 DPI devices: 0 - red, 1 - purple, 2 - blue (default), 3 - green
+// for 2 DPI devices: 0 - main (default), 1 - alternative
+#define ASUS_MAX_NUM_DPI 4
+
+// maximum number of LEDs across all ASUS devices
+#define ASUS_MAX_NUM_LED 3
```
whot, line 23: afaict both actual source files include asus-defines and asus-common.h. I dont' think we have any future uses for just these defines here in libratbag so let's just have one header instead.

src/asus.c
```diff
@@ -0,0 +1,429 @@
...
+	RATBAG_LED_TYPE_LOGO,
+	RATBAG_LED_TYPE_WHEEL,
+	RATBAG_LED_TYPE_SIDE,  // bottom back
+};
+
+// search for asus button index by ratbag types
+int
+asus_find_button_index_by_action(struct ratbag_button_action action)
+{
+	for (unsigned int i = 0; i < ARRAY_LENGTH(ASUS_BUTTON_MAPPING); i++) {
```
whot, line 53: just a btw, we also have `ARRAY_FOR_EACH` which would work here. No action needed if you don't want to change it.

src/asus.c
```diff
@@ -0,0 +1,429 @@
...
+	memcpy(data->raw, response.raw, sizeof(data->raw));
+	return 0;
+}
+
+// set button binding using ASUS code of the button
+int
+asus_set_button_action(struct ratbag_device *device, unsigned int asus_index, uint8_t asus_code, uint8_t asus_type)
+{
+	int r;
+	union asus_response response = {0};
```
whot, line 233: unless I overlooked something, you could `memset` the response in `asus_query` which would reduce the chance of a bug being introduced by forgetting the `= {0};`.

src/asus.c
```diff
@@ -0,0 +1,429 @@
...
+asus_set_button_response(struct ratbag_device *device, unsigned int ms)
+{
+	int dpi_count = ratbag_device_data_asus_get_dpi_count(device->data);
+	int r;
+	union asus_response response = {0};
+
+	union asus_request request = {
+		.data.cmd = ASUS_CMD_SET_SETTING,
+		.data.params[0] = dpi_count + 1,  // next entry after rate
+		.data.params[2] = ms / 4 - 1,
```
whot, line 359: might be good to leave an assert here to avoid `ms` being zero.

src/asus.c
```diff
@@ -0,0 +1,429 @@
...
+
+	r = asus_query(device, &request, &response);
+	if (r)
+		return r;
+
+	return 0;
+}
+
+int
+asus_set_angle_snapping(struct ratbag_device *device, unsigned int is_enabled)
```
whot, line 370:
```suggestion
asus_set_angle_snapping(struct ratbag_device *device, bool is_enabled)
```

src/driver-asus.c
```diff
@@ -0,0 +1,516 @@
...
+					if (asus_index_to >= 0) {
+						button->action.type = ASUS_BUTTON_MAPPING[asus_index_to].type;
+						if (ASUS_BUTTON_MAPPING[asus_index_to].type == RATBAG_BUTTON_ACTION_TYPE_BUTTON)
+							button->action.action.button = ASUS_BUTTON_MAPPING[asus_index_to].button;
+						else if (ASUS_BUTTON_MAPPING[asus_index_to].type == RATBAG_BUTTON_ACTION_TYPE_SPECIAL)
+							button->action.action.special = ASUS_BUTTON_MAPPING[asus_index_to].special;
+					} else {
+						log_debug(device->ratbag, "Unknown action code %02x\n", button_data->data.button[asus_index].action);
+					}
+					break;
```
whot, line 165: for this code, I really recommend using temporary variables so you dont need to use `foo->bar[bat->index].baz`. On higher optimisation those are free anyway, so having something like this makes the code easier to read:
```c
   struct foo *mapping = &ASUS_BUTTON_MAPPING[asus_index_to];
   if (mapping->type == ACTION_TYPE_BUTTON)
      button->action.type = mapping->type;

File: src/driver-asus.c [hunk 15]
```diff
@@ -0,0 +1,516 @@
...
+			device->ratbag, "Resolution %d changed to %d\n",
+			resolution->index, resolution->dpi_x);
+
+		error = asus_set_dpi(device, resolution->index, resolution->dpi_x);
+		if (error)
+			return error;
+	}
+
+	// set LEDs
+	if (led_count) {
```
whot, line 358: is this check necessary? `ratbag_profile_for_each_led` should just be an empty loop if we don't have LEDs.

src/driver-asus.c
```diff
@@ -0,0 +1,516 @@
...
+				return error;
+		}
+	}
+
+		// save profile
+		log_debug(device->ratbag, "Saving profile\n");
+		error = asus_save_profile(device);
+		if (error)
+			return error;
+	}
```
whot, line 385: something's wrong with the indentation for this block to be indented - line 270, the profile loop.

Might be easiest to split this function into the one that loops of the profile and one helper function that takes the profile as argument and does the work just for this profile (i.e. the loop body as one function). This way it ends up being:
```c
int asus_driver_save_profiles(...) {
    ratbag_device_for_each_profile(device, profile) {
	if (!profile->dirty)
		continue;
        asus_driver_save_profile(profile);
   }
}
```

src/driver-asus.c
```diff
@@ -0,0 +1,516 @@
...
+
+	// init profiles
+	ratbag_device_init_profiles(
+		device, max(profile_count, 1), dpi_count, button_count, led_count);
+
+	// setup profiles
+	ratbag_device_for_each_profile(device, profile) {
+		if (profile->index == 0) {
+			profile->is_active = true;
+		}
```
whot, line 432: nitpick - no `{}` needed here.

src/driver-asus.c
```diff
@@ -0,0 +1,516 @@
...
+			asus_setup_button(device, button);
+
+		ratbag_profile_for_each_resolution(profile, resolution)
+			asus_setup_resolution(device, resolution);
+
+		if (!led_count)
+			continue;
+
+		ratbag_profile_for_each_led(profile, led)
+			asus_setup_led(device, led);
```
whot, line 446: same as above, this should be an empty loop if the led count is 0 so the extra check shouldn't be needed

src/libratbag-data.c
```diff
@@ -269,6 +283,98 @@ init_data_steelseries(struct ratbag *ratbag,
...
+	_cleanup_(freep) char *str = NULL;
+	_cleanup_(g_strfreevp) char **strv = NULL;
+
+	data->asus.profile_count = 3;
+	data->asus.button_count = 8;
+	data->asus.led_count = 0;
+	data->asus.dpi_count = 2;
+	data->asus.dpi_range = NULL;
+	data->asus.is_wireless = 0;
+	data->asus.quirks = 0;
```
whot, line 303: the above should all be -1 for "unset" and then handled in the driver

src/libratbag-data.c
```diff
@@ -269,6 +283,98 @@ init_data_steelseries(struct ratbag *ratbag,
...
+		data->asus.is_wireless = num;
+	if (error)
+		g_error_free(error);
+
+	error = NULL;
+	strv = g_key_file_get_string_list(keyfile, group, "Quirks", NULL, NULL);
+	if (strv) {
+		i = 0;
+		while (strv[i]) {
+			if (strcmp(strv[i], "DOUBLE_DPI") == 0) {
```
whot, line 366:
```suggestion
			if (streq(strv[i], "DOUBLE_DPI")) {
```

src/libratbag-data.c
```diff
@@ -269,6 +283,98 @@ init_data_steelseries(struct ratbag *ratbag,
...
+				data->asus.quirks |= ASUS_QUIRK_DOUBLE_DPI;
+			} else if (strcmp(strv[i], "STRIX_PROFILE") == 0) {
```
whot, line 368:
```suggestion
			} else if (streq(strv[i], "STRIX_PROFILE")) {
```

src/asus-defines.h
kyokenn: If I move everything into a single .h file the compiler would complain about unused stuff while trying to compile the libratbag-data.

whot: whats the error message you're getting?

kyokenn: It's not an error, but annoying warning:
```
[2/16] Compiling C object libratbag.a.p/src_libratbag-data.c.o
In file included from ../src/libratbag-data.c:33:
../src/asus-common.h:163:21: warning: 'ASUS_POLLING_RATES' defined but not used [-Wunused-variable]
  163 | static unsigned int ASUS_POLLING_RATES[] = { 125, 250, 500, 1000 };
      |                     ^~~~~~~~~~~~~~~~~~
../src/asus-common.h:135:22: warning: 'ASUS_KEY_MAPPING' defined but not used [-Wunused-variable]
  135 | static unsigned char ASUS_KEY_MAPPING[] = {
      |                      ^~~~~~~~~~~~~~~~
```

whot: those arrays need to go into `src/driver-asus.c` - that's where they're being used. These two are very libratbag-driver specific since they just convert (or provide) some format libratbag requires that the device itself doesn't have in that format.

So moving them should get rid of the warning too then.

src/libratbag-data.c
```diff
@@ -269,6 +283,98 @@ init_data_steelseries(struct ratbag *ratbag,
 	}
 }
 
+static void
+init_data_asus(struct ratbag *ratbag,
+		GKeyFile *keyfile,
+		struct ratbag_device_data *data)
+{
+	const char *group = "Driver/asus";
+	int num, i;
```
staticssleever668, line 292: Don't define `i` here, do it inside the loops.

src/libratbag-data.c
```diff
@@ -269,6 +283,98 @@ init_data_steelseries(struct ratbag *ratbag,
...
+	GError *error = NULL;
+	_cleanup_(freep) char *str = NULL;
```
staticssleever668, line 294: Analogous to `strv`: make this a variable for a concrete thing to prevent mistakes from happening in the future.

src/libratbag-data.c
```diff
@@ -269,6 +283,98 @@ init_data_steelseries(struct ratbag *ratbag,
...
+	_cleanup_(g_strfreevp) char **strv = NULL;
```
staticssleever668, line 295: Use a separate variable for each `g_key_file_get_string_list` call or there will be a memory leak.

src/libratbag-data.c
```diff
@@ -269,6 +283,98 @@ init_data_steelseries(struct ratbag *ratbag,
...
+		data->asus.button_mapping[i] = -1;
+
+	error = NULL;
+	num = g_key_file_get_integer(keyfile, group, "Profiles", &error);
+	if (!error && num >= 0)
+		data->asus.profile_count = num;
+	if (error)
+		g_error_free(error);
+
+	error = NULL;
```
staticssleever668, line 314: Use the suggested sample here and everywhere after.
```suggestion
	g_clear_error(&error);
```

src/libratbag-data.c
```diff
@@ -269,6 +283,98 @@ init_data_steelseries(struct ratbag *ratbag,
...
+	num = g_key_file_get_integer(keyfile, group, "Buttons", &error);
+	if (!error && num >= 0 && num <= ASUS_MAX_NUM_BUTTON)
+		data->asus.button_count = num;
+	if (error)
+		g_error_free(error);
+
+	error = NULL;
+	strv = g_key_file_get_string_list(keyfile, group, "ButtonMapping", NULL, NULL);
```
staticssleever668, line 322:
```suggestion
	strv = g_key_file_get_string_list(keyfile, group, "ButtonMapping", &list_length, &error);
```

Define this `list_length` (or whatever name you think would fit better) in the beginning.

src/libratbag-data.c
```diff
@@ -269,6 +283,98 @@ init_data_steelseries(struct ratbag *ratbag,
...
+	if (strv) {
+		i = 0;
+		while (strv[i]) {
+			data->asus.button_mapping[i] = (int8_t) strtoul(strv[i], NULL, 10);
+			i++;
+		}
```
staticssleever668, line 328:
```suggestion
		for (unsigned int i = 0; i <= list_length; i++) {
			data->asus.button_mapping[i] = (int8_t) strtoul(strv[i], NULL, 10);
		}
```
Please, also check if my suggested range `i <= list_length` is correct.

src/libratbag-data.c
```diff
@@ -269,6 +283,98 @@ init_data_steelseries(struct ratbag *ratbag,
...
+
+	error = NULL;
+	num = g_key_file_get_integer(keyfile, group, "Dpis", &error);
+	if (!error && num >= 2 && num <= ASUS_MAX_NUM_DPI)
+		data->asus.dpi_count = num;
+	if (error)
+		g_error_free(error);
+
+	error = NULL;
+	str = g_key_file_get_string(keyfile, group, "DpiRange", NULL);
```
staticssleever668, line 348:
```suggestion
	str = g_key_file_get_string(keyfile, group, "DpiRange", &error);
```

src/libratbag-data.c
```diff
@@ -269,6 +283,98 @@ init_data_steelseries(struct ratbag *ratbag,
...
+
+	error = NULL;
+	num = g_key_file_get_integer(keyfile, group, "Wireless", &error);
+	if (!error && (num == 0 || num == 1))
+		data->asus.is_wireless = num;
+	if (error)
+		g_error_free(error);
+
+	error = NULL;
+	strv = g_key_file_get_string_list(keyfile, group, "Quirks", NULL, NULL);
```
staticssleever668, line 362:
```suggestion
	strv = g_key_file_get_string_list(keyfile, group, "Quirks", &list_length, &error);
```

src/libratbag-data.c
```diff
@@ -269,6 +283,98 @@ init_data_steelseries(struct ratbag *ratbag,
...
+	if (strv) {
+		i = 0;
+		while (strv[i]) {
```
staticssleever668, line 365:
```suggestion
		for (unsigned int i = 0; i <= list_length; i++) {
```
Please, also check if this suggested range is correct.

src/libratbag-data.c
```diff
@@ -269,6 +283,98 @@ init_data_steelseries(struct ratbag *ratbag,
...
+			if (streq(strv[i], "DOUBLE_DPI")) {
+				data->asus.quirks |= ASUS_QUIRK_DOUBLE_DPI;
+			} else if (streq(strv[i], "STRIX_PROFILE")) {
+				data->asus.quirks |= ASUS_QUIRK_STRIX_PROFILE;
+			}
```
staticssleever668, line 370: else: print a warning.

kyokenn:
>Is there a reason for having the whole drv_data->is_ready thing instead of erroring out from start? As far as I see it only complicates the code and gives no real benefit.

The mouse could be sleeping when we start ratbagd or it can go to sleep later. So I want to be able to reinitialize it when it awakes.

staticssleever668 (approved):
Very nice! Thanks for going so fast over the review, too. :)
Would you squash all of the commits if that's fine by you? Doesn't seem like they are in form of incremental changes anyway.

src/driver-asus.c
```diff
@@ -0,0 +1,517 @@
...
+	RATBAG_LED_CYCLE,
+	RATBAG_LED_ON,  /* rainbow wave */
+	RATBAG_LED_ON,  /* reactive - react to clicks */
+	RATBAG_LED_ON,  /* custom - depends on mouse type */
+	RATBAG_LED_ON,  /* battery - battery indicator */
+};
+
+struct asus_data {
+	uint8_t is_ready;
+} __attribute__((packed));
```
staticssleever668, line 72: I don't think you need `__attribute__((packed))` as it's not some data we receive from the mouse, it's something we define and use ourselves.

src/asus.h
```diff
@@ -0,0 +1,212 @@
...
+
+/* maximum number of LEDs across all ASUS devices */
+#define ASUS_MAX_NUM_LED 3
+
+/* base request */
+
+struct _asus_request {
+	uint16_t cmd;
+	uint8_t params[ASUS_PACKET_SIZE - 2];
+} __attribute__((packed));
```
staticssleever668, line 34: I think you need to statically assert size of every packed struct to make sure the compiler doesn't resize anything.
`_Static_assert(sizeof(struct _asus_response) == ASUS_PACKET_SIZE, "The size of `_asus_response` is wrong.");`

src/asus.h
```diff
@@ -0,0 +1,212 @@
...
+
+union asus_response {
+	struct _asus_response data;
+	uint8_t raw[ASUS_PACKET_SIZE];
+};
+
+/* current profile ID and firmware info */
+
+struct asus_profile_data {
+	unsigned int profile_id;
```
staticssleever668, line 56: ..https://github.com/libratbag/libratbag/pull/1250/files#r993419635

Especially here since `unsigned int` may be a different thing depending on target platform.

kyokenn:
> Would you squash all of the commits if that's fine by you? Doesn't seem like they are in form of incremental changes anyway.

I have combined them into single commit.

staticssleever668:
I wonder if the by LGTM bot is actually caused by some changes in this patch. :thinking: 
I'll look into it when I'll have a minute and merge this PR. :+1:

src/asus.h
```diff
@@ -0,0 +1,214 @@
...
+	enum ratbag_button_action_special special;  /* special action, optional */
+};
+
+struct asus_button *
+asus_find_button_by_action(struct ratbag_button_action action);
+
+struct asus_button *
+asus_find_button_by_code(uint8_t asus_code);
+
+extern int
```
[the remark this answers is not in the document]
kyokenn: Yeah, I understood. I just forgot to remove some of them.

staticssleever668:
Dismissed a stale review to unblock merging (I guess it's okay).
While someone new might take a look on this too, it may take months as there are no active maintainers, so let's merge this before it gets stale again. :)

PR #1266, opened 2022-01-20
roccat: name two fields correctly

whot:
In hindsight, these should've been obvious.

cc @bentiss


bentiss (approved):
Happy to say that it looks good to me :)

PR #1268, opened 2022-01-24
fix ratbag-command led setting

whot:
Extracted from #1250 so we don't need to wait until the driver is ready

cc @kyokenn

### DIFF
--- a/tools/ratbagc.py.in
+++ b/tools/ratbagc.py.in
@@ -808,7 +808,10 @@ class RatbagdLed(metaclass=MetaRatbag):
 
         @param color An RGB color, as an integer triplet with values 0-255.
         """
-        libratbag.ratbag_led_set_color(self._led, libratbag.ratbag_color(*color))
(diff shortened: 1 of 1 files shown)


staticssleever668:
What is the problem behind this patch? I can't find it in the linked pull request and fail to see a difference this change makes.

Nit: I don't think variable name `rc` applies here, I think it should only be used for return codes, as the name implies.
Nit: you didn't drop the commented out line. :)

whot:
It was in https://github.com/libratbag/libratbag/pull/1250 and looked correct enough, so I factored it out. @kyokenn can you expand on the need for this patch please?


And yes, I did a sloppy job here, I need to sleep more...

kyokenn:
ratbag-command crashes when trying to set color without this patch:
```
ratbag-command /dev/hidraw4 led 0 set color ff0000
```
```
TypeError: ratbag_color.__init__() takes 1 positional argument but 4 were given
```

PR #1269, opened 2022-01-24
sinowealth: detect LED count at run-time

staticssleever668:
I had to move some stuff around to not read data from the mouse twice.

### DIFF
--- a/src/driver-sinowealth.c
+++ b/src/driver-sinowealth.c
@@ -254,7 +254,6 @@ static int
 sinowealth_print_fw_version(struct ratbag_device *device) {
 	int rc = 0;
 
-	/* Print firmware version */
 	uint8_t version[6] = {SINOWEALTH_REPORT_ID_CMD, SINOWEALTH_CMD_FIRMWARE_VERSION};
 	rc = ratbag_hidraw_set_feature_report(device, SINOWEALTH_REPORT_ID_CMD, version, sizeof(version));
 	if (rc != sizeof(version)) {
--- a/src/driver-sinowealth.c
+++ b/src/driver-sinowealth.c
@@ -266,7 +266,7 @@ sinowealth_print_fw_version(struct ratbag_device *device) {
 		log_error(device->ratbag, "Could not read firmware version: %d\n", rc);
 		return -1;
 	}
-	log_debug(device->ratbag, "firmware version: %.4s\n", version + 2);
+	log_info(device->ratbag, "firmware version: %.4s\n", version + 2);
 
 	return 0;
 }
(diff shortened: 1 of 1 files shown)


src/driver-sinowealth.c
```diff
@@ -361,9 +367,16 @@ sinowealth_read_profile(struct ratbag_profile *profile)
 static int
 sinowealth_init_profile(struct ratbag_device *device)
 {
+	int rc;
 	struct ratbag_profile *profile;
 	struct ratbag_resolution *resolution;
 	struct ratbag_led *led;
+
+	rc = sinowealth_print_fw_version(device);
+	if (rc) {
```
whot, line 399: `{}` shouldn't be needed here

src/driver-sinowealth.c
```diff
@@ -388,6 +388,11 @@ sinowealth_init_profile(struct ratbag_device *device)
 		return rc;
 	}
 
+	rc = sinowealth_read_raw_data(device);
+	if (rc) {
```
whot, line 404: same here, curly braces aren't necessary here

src/driver-sinowealth.c
```diff
@@ -71,7 +71,13 @@ enum rgb_effect {
...
 	RGB_TAIL = 0x4,
 	RGB_RAVE = 0x7,
-	RGB_WAVE = 0x9
+	RGB_WAVE = 0x9,
+
+	/* The value mice with no LEDs have.
+	* Unreliable as non-constant.
+	* Do **not** overwrite it.
+	 */
+	RGB_NOT_SUPPORTED = 0xff,
```
whot, line 80: If it's not constant, how about this approach:
```c
enum rgb_effect {
   ....
   RGB_WAVE = 0x9,
   _RGB_EFFECT_UNSUPPORTED /* Usually 0xff but it's unreliable so let' assume anything not know is unsupoprted */
};
```
and later

```c
if (config->rgb_effect < _RGB_EFFECT_UNSUPPORTED) {
```
This way anything we don't explicitly support becomes an unsupported one without a magic value

whot (approved):
Looks good, thanks, 2 minor nitpicks and a suggestion but feel free to use as-is

src/driver-sinowealth.c
staticssleever668: I like current approach better, because if there happen to be lighting modes we don't know about (and there are such, currently), the user would not be able to change lighting modes at all.

src/driver-sinowealth.c
staticssleever668: Thanks, fixed.

staticssleever668:
Added a small memory leak fix, which is a bit out of scope of this PR, but it would have a merge conflict if I make a new PR right now.

src/driver-sinowealth.c
whot: iirc in other drivers we just transparently map the unknown mode to a known one and pretend everything is ok. As long as you don't touch it it (should) stay as-is even if it's technically listed as wrong mode.

But yeah, we do really need a `RATBAG_LED_UNKNOWN`  enum value for this case...

whot:
Thanks!

PR #1276, opened 2022-01-31
steelseries: fix a memory leak

staticssleever668:
Found by valgrind. I also feel that there is another memory leak of the
same nature a bit higher in `init_data_hidpp10`.

Introduced by me. :^)

### DIFF
--- a/src/libratbag-data.c
+++ b/src/libratbag-data.c
@@ -214,7 +214,8 @@ init_data_steelseries(struct ratbag *ratbag,
 {
 	const char *group = "Driver/steelseries";
 	GError *error = NULL;
-	_cleanup_(freep) char *str = NULL;
+	_cleanup_(freep) char *dpi_range = NULL;
+	_cleanup_(freep) char *quirk = NULL;
 	int num;
 
 	data->steelseries.device_version = -1;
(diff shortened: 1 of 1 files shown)


src/libratbag-data.c
```diff
@@ -251,6 +251,9 @@ init_data_steelseries(struct ratbag *ratbag,
 		str = g_key_file_get_string(keyfile, group, "DpiList", NULL);
 		if (str)
 			data->steelseries.dpi_list = dpi_list_from_string(str);
+
+		free(str);
+		str = NULL;
```
whot, line 256: this shouldn't be necessary because we mark `str` as `_cleanup_(freep)`. The issue here is that we're re-using it so the value gets overwritten but re-using auto-cleanup pointers is bad form. A better approach would be to have a separate variable for the dpi string and the quirk string below, both of hem with `_cleanup_(freep)`

also, indentation looks really messed up here in this block unless that's the github rendering.

whot (changes requested):

src/libratbag-data.c
staticssleever668: Hmm, I see. I don't know much about them because I have never used them myself. Thanks for an advice!
I remade this like you explained here.

About indentation, seems fine to me? Both in the editor and on GitHub.

whot: weird, it looks fine in the snippet above but in the diff view it was off. Anyway, all good then, thanks for the change.

whot (approved):

PR #1277, opened 2022-01-31
Remove old driver methods

staticssleever668:
This basically puts now removed `ratbag_old_write_profile` in every driver that has not yet migrated to the profile-oriented API. The only other change is that now `etekcity` driver only writes profile once. While is might in theory be breaking, I think it's safe to assume it wont as I don't think official software destroys EEPROM of the mouse by repetitive overwriting like it did before this change.


amfern (approved):
LGTM

Issue #1223, opened 2021-11-04
steelseries setting led to off set blue to max


amfern:
Everything behaved the same as before, so this patch is fine.

I, like @klausenbusk, and like i experienced before, had issues setting the led modes to off, i ruled out hardware, and setting the color works fine the whole range, but when i set the leds to off sets blue to max and everything else off, not sure what the cause here is, but maybe mapping off to color 000000 is a good option.

_Originally posted by @perigoso in https://github.com/libratbag/libratbag/issues/1210#issuecomment-944696383_

staticssleever668:
Please note that this intended behavior on V1 devices (for now).
LED `on` and `off`do the same action. https://github.com/libratbag/libratbag/blob/ae53f6123b7924ee1a649243d77ba251c949d6fe/src/driver-steelseries.c#L684
I have not tested how it works in official software, but I think that either there is no `off` option for such mice or it really just maps to black (#000).

PR #1280, opened 2022-02-06
steelseries: fix length of LED cycle output report being incorrect

staticssleever668:
It probably meant to use the macro directly, not `sizeof` of it (which
made it be 4 instead of 64).

Found by clang-tidy. I'm not sure how it worked before considering this, so somebody with a V2 mouse should test this first.


amfern (approved):
Thank you, wonder if this was the reason for the LED bug here https://github.com/libratbag/libratbag/issues/1223

whot (approved):

PR #1298, opened 2022-04-23
Little pull request for helping gentoo and openrc users

FellFromTheSky:
I was looking for why libratbag didn't work for non root users on my gentoo system.
And I found why: 
https://forums.opensuse.org/showthread.php/537468-Piper-not-working
The user need to be added to the games group.
This doesn't seems to be written in the docs.
Feel free to edit this pull request or cancel it.
But please add something similar to the docs or the README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -85,6 +85,9 @@ And finally, to enable the service:
 This places the required symlink into the systemd directory so that dbus
 activation is possible.
 
+Gentoo and OpenRC users only:
+If using a non root user, this user must be added to the games group.
+
 The DBus Interface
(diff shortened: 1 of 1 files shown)


FellFromTheSky:
You could also add that there's a Gentoo packages for libratbag and Piper :
https://packages.gentoo.org/packages/dev-libs/libratbag
https://github.com/libratbag/piper

Thanks a lot for adding the entry.

amfern:
Thank you, I added the libratbag gentoo package link to the installation

PR #1311, opened 2022-05-18
SinoWealth driver improvemnts

staticssleever668:
- Make it easier to add new devices by putting devices data into a struct, instead of working with the values manually. This way a person that doesn't understand coding can easily add new data (though they would have to be familiar with building process).
- Automatically detect sensor type, and if the sensor type was recognized, set corresponding maximum DPI.
- Add support for Marvo Scorpion G961, which has shorter configuration data. Though to use it, a device file for ratbagd would have to be created manually, as another mice with the same VID:PID has been reported not to work (Ant Esports GM500). Incidentally it also has shorter configuration data.
- Minor quality of life improvements.

### DIFF
--- a/src/driver-sinowealth.c
+++ b/src/driver-sinowealth.c
@@ -572,7 +572,7 @@ sinowealth_raw_to_report_rate(uint8_t raw)
 
 /* @return Maximum DPI for sensor `sensor` or 0 on error. */
 static unsigned int
-get_max_dpi_for_sensor(enum sinowealth_sensor sensor)
+sinowealth_get_max_dpi_for_sensor(enum sinowealth_sensor sensor)
 {
 	switch (sensor) {
 	case PWM3327: return 10200;
@@ -610,7 +610,7 @@ sinowealth_dpi_to_raw(struct ratbag_device *device, unsigned int dpi)
 	struct sinowealth_data *drv_data = device->drv_data;
 	enum sinowealth_sensor sensor = drv_data->sensor;
 
-	assert(dpi >= SINOWEALTH_DPI_MIN && dpi <= get_max_dpi_for_sensor(sensor));
+	assert(dpi >= SINOWEALTH_DPI_MIN && dpi <= sinowealth_get_max_dpi_for_sensor(sensor));
 
 	unsigned int raw = dpi / 100;
 
@@ -1453,7 +1453,7 @@ sinowealth_init_profile(struct ratbag_device *device)
 	/* Number of DPIs = all DPIs from min to max (inclusive) and "0 DPI" as a special value
 	 * to signal a disabled DPI step.
 	 */
-	unsigned int num_dpis = (get_max_dpi_for_sensor(drv_data->sensor) - SINOWEALTH_DPI_MIN) / SINOWEALTH_DPI_STEP + 2;
+	unsigned int num_dpis = (sinowealth_get_max_dpi_for_sensor(drv_data->sensor) - SINOWEALTH_DPI_MIN) / SINOWEALTH_DPI_STEP + 2;
 
 	ratbag_device_init_profiles(device, SINOWEALTH_NUM_PROFILES, SINOWEALTH_NUM_DPIS, drv_data->button_count, drv_data->led_count);
 
--- a/src/driver-sinowealth.c
+++ b/src/driver-sinowealth.c
@@ -1464,6 +1464,18 @@ static const struct sinowealth_device_data sinowealth_supported_devices[] = {
 		.button_count = 6,
 	},
 
+	/* 2581:1007 devices. */
+	{
+		.fw_version = "9677",
+		.device_name = "Marvo Scorpion G961",
+		.led_type = LED_RGB,
+		/* Official software allows up to 12000 DPI instead of 10200. */
+		.sensor_type = SINOWEALTH_SENSOR_PMW3327,
+		.vid = 0x258a,
+		.pid = 0x1007,
+		.button_count = 6,
+	},
+
 	/* Dummy device. Make sure it's the last entry. */
 	{
 		.fw_version = "0000",
--- a/src/driver-sinowealth.c
+++ b/src/driver-sinowealth.c
@@ -495,7 +495,7 @@ sinowealth_button_data_is_equal(const struct sinowealth_button_data *lhs, const 
 
 /* Convert a button action to raw data using the `sinowealth_button_map`.
  * NOTE: It does not contain all of the button types, as some of them are
- * better made programatically. See @ref sinowealth_update_buttons_from_profile.
+ * better made programmatically. See @ref sinowealth_update_buttons_from_profile.
  *
  * @param data Struct to write to.
  *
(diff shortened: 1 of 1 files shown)


src/driver-sinowealth.c
```diff
@@ -1465,6 +1465,18 @@ static const struct sinowealth_device_data sinowealth_supported_devices[] = {
 		.button_count = 6,
 	},
 
+	/* 2581:1007 devices. */
+	{
+		.fw_version = "9677",
+		.device_name = "Marvo Scorpion G961",
```
whot, line 1471: regarding the commit message: do we have a name match yet? If not, that could simply be added to avoid that issue.

src/driver-sinowealth.c
```diff
@@ -1796,6 +1796,16 @@ sinowealth_update_config_from_profile(struct ratbag_profile *profile)
...
 		if (!resolution->dpi_x || !resolution->dpi_y)
 			continue;
+
+		/* Limit the resolution if it somehow got higher than allowed. */
+		{
+			const unsigned int max_dpi = sinowealth_get_max_dpi_for_sensor(config->sensor_type);
+			if (resolution->dpi_x > max_dpi)
+				resolution->dpi_x = max_dpi;
+			if (resolution->dpi_y > max_dpi)
+				resolution->dpi_y = max_dpi;
```
whot, line 1819: easier, and from `shared-macro.h`:
```c
resolution->dpi_x = min(resolution->dpi_x, max_dpi);
```

src/driver-sinowealth.c
```diff
@@ -1796,6 +1796,16 @@ sinowealth_update_config_from_profile(struct ratbag_profile *profile)
 	ratbag_profile_for_each_resolution(profile, resolution) {
 		if (!resolution->dpi_x || !resolution->dpi_y)
 			continue;
+
+		/* Limit the resolution if it somehow got higher than allowed. */
+		{
```
whot, line 1814: don't think we need the extra block here, we support C99 anyway and we can probably start using late declarations

src/driver-sinowealth.c
```diff
@@ -1368,6 +1368,130 @@ sinowealth_update_macro_from_action(struct ratbag_profile *profile, struct ratba
...
+static const struct sinowealth_device_data *
+sinowealth_find_device_data(struct input_id ids, const char *fw_version)
+{
+	const struct sinowealth_device_data *device_data = NULL;
+
+	ARRAY_FOR_EACH(sinowealth_supported_devices, device_data) {
+		if (device_data->vid != ids.vendor || device_data->pid != ids.product)
+			continue;
+
+		if (strncmp(fw_version, device_data->fw_version, sizeof(device_data->fw_version)) != 0)
```
whot, line 1501: `streq` or `strneq` is less prone to bugs than `strcmp`

src/driver-sinowealth.c
```diff
@@ -1368,6 +1368,130 @@ sinowealth_update_macro_from_action(struct ratbag_profile *profile, struct ratba
...
+	{
+		.fw_version = "0000",
+		.device_name = "Generic SinoWealth Mouse",
+		.led_type = LED_RBG,
+		.sensor_type = SINOWEALTH_SENSOR_PMW3360,
+		.vid = 0xffff,
+		.pid = 0xffff,
+		.button_count = 6,
+	},
+};
```
whot, line 1490: fwiw, this should really all be in the `.device` files, we do support driver-specific sections (even if the whole thing is a tad awkward).

src/driver-sinowealth.c
```diff
@@ -1090,6 +1090,9 @@ sinowealth_update_profile_from_config(struct ratbag_profile *profile)
 			led->color = sinowealth_raw_to_color(device, config->breathing1_color);
 			sinowealth_set_led_from_rgb_mode(led, config->breathing1_mode);
 			break;
+		case RGB_NOT_SUPPORTED:
+			/* This should be unreachable. */
+			break;
```
whot, line 1110: better:
```c
case RGB_NOT_SUPPORTED:
default:
 ...
```
This way it's clear you didn't just forget about this effect *and* you get the error message if it ever happens

whot (approved):
A few general comments, I'd really encourage you to start using a driver-specific section in the device files though instead of coding it into the driver.

I've given you access now so you can merge these yourself - no point in you waiting idly for *someone* to come along and review the sinowealth patches :)

src/driver-sinowealth.c
staticssleever668: I've actually put it in a block not to pollute the namespace in this scope (personal preference). If you don't think it makes sense I will remove it. :)

src/driver-sinowealth.c
staticssleever668: What do you mean a `name match`? This `device_name` field is not actually reported by the mouse, it's something I added to distinguish between all of these device entries. Used device is actually detected by `fw_version` field.

staticssleever668:
I like the idea of putting device-specific data into device files, but I'm really unsure how to do it. Could you maybe make an example device file for me so that I can imagine how it would look like?

I would imagine it something like so, but I'm not even sure it's possible to iterate over arbitrary section names.

```dosini
[Device]
DeviceMatch=usb:258a:0027
Driver=sinowealth
Name=SinoWealth Generic Mouse (0027)

[Driver/sinowealth/devices/whateverdevicename1]
ButtonCount=6
DeviceName=WhateverCompany WhateverModel
FwVersion=1234
LedType=RGB
SensorType=PWM3360

[Driver/sinowealth/devices/GenesysXenon770]
# There are 5 main buttons.
# Side buttons are detachable.
# There can be either 3 or 9 buttons on the side.
ButtonCount=14
DeviceName=Genesys Xenon 770
FwVersion=3110
LedType=RGB
SensorType=PWM3327
```

src/driver-sinowealth.c
staticssleever668: Fixed :)

src/driver-sinowealth.c
staticssleever668: Fixed :)

src/driver-sinowealth.c
staticssleever668: Fixed :)

staticssleever668:
I will merge this for now (I assume it's okay if you gave me the rights 😃), and rework this to put device-specific data into device files in a separate pull request, I think I have a rough idea on how to do it.

src/driver-sinowealth.c
whot: oh, that name isn't even in the hidraw? What does the device report itself as?

staticssleever668: All of them report themselves as `SINOWEALTH Wired Gaming Mouse`.

whot: well, that's reasonably useless... in that case we'd need some extra matching logic so we can identify the driver based on some matches, but punt the exact match to the driver. Which is going to need some... intermediate plans, so for now it's easier to go with what you have.

PR #1312, opened 2022-05-28
Create sinowealth-0028.device

Meathook33:


staticssleever668:
Hello, thanks for your contribution. Does everything work correctly with that mouse?
If possible, could you attach the output of `sudo /bin/ratbagd --verbose` so that I can add information about this mouse to the driver?

staticssleever668:
Now that #1332 has been merged, you *must* add information about the mouse to the device file.
If you still wish to add support for this mouse, I suggest you using following device file as the base:

```dosini
[Device]
DeviceMatch=usb:258a:0028
Driver=sinowealth
Name=SinoWealth Generic Mouse (0028)

[Driver/sinowealth/devices/TitanwolfLAW]
ButtonCount=12
DeviceName=Titanwolf LAW
FwVersion=YJG23
LedType=RBG
SensorType=PMW3360
``` 

Two important fields are `FwVersion` and `LedType`.

Firmware version is printed when libratbag tries to detect the device.
One way to see it is doing so (assuming your system uses systemd, which is most likely the case):
1) Close Piper
2) Stop the process of libratbag: in the terminal do `sudo systemctl stop ratbagd.service`
3) Follow new messages in the system journal: in the terminal `journalctl -b -f`.
4) Start Piper
Now press `Ctrl+C` in the terminal to stop following the journal.
Look for a similar line:
`Aug 03 12:01:50 pcname ratbagd[21481]: ratbag error: Device with firmware version V103 is not supported`.
`V103` here is the firmware version, put it into the `FwVersion` field of the device file.

To find the LED type you have to change lighting color on the mouse in Piper.
If it's different from what's shown, change `LedType` from `RBG` to `RGB` and restart libratbag and Piper: `sudo systemctl stop ratbagd.service`.
If there are no LEDs at all, set `None`.

PR #1313, opened 2022-05-29
Add device-file for Logitech MX Anywhere 3B

olmari:
This variant stands for Logitech MX Anywhere 3 for Business.

Signed-off-by: Sami Olmari <sami@olmari.fi>

### DIFF
--- a/data/devices/logitech-MX-Anywhere3B.device
+++ b/data/devices/logitech-MX-Anywhere3B.device
@@ -1,0 +1,4 @@
+# Logitech MX Anywhere 3B
+[Device]
+Name=Logitech MX Anywhere 3B
+DeviceMatch=bluetooth:046d:b02d;usb:046d:c548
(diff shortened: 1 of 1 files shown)


olmari:
I took Anywhere 3 device-file and changed ID's to match 3B, seems to work a okay.

whot:
Looking at [the Solaar list of receivers](https://pwr-solaar.github.io/Solaar/devices.html) this is a receiver device, so we can't add it as device file, it would catch all other devices on that same receiver type. I've updated the receiver check in #1314 now but for this to fully work we'll need kernel support. Unfortunately I've paged out everything regarding that, maybe @FFY00 has some memory of it.

olmari:
Well, in mean time adding only bluetooth "side" then, as it works with both bleutooth and the whatever logitech "not bluetooth" dongle it is :)

data/devices/logitech-MX-Anywhere3B.device
```diff
@@ -0,0 +1,4 @@
+# Logitech MX Anywhere 3B
+[Device]
+Name=Logitech MX Anywhere 3B
+DeviceMatch=bluetooth:046d:b02d;usb:046d:c548
```
whot, line 4:
```suggestion
DeviceMatch=bluetooth:046d:b02d
```

whot (approved):
Going to merge this with the USB match removed, better than having nothing :)

PR #1315, opened 2022-06-06
sinowealth: fix all devices being detected as `generic device`

staticssleever668:
Hot-fix.
I did not notice that the logic was inverted when I replaced `strncmp` with `strneq`, and I didn't test it, oops.
I guess that's exactly the reason why `strneq` is more safe. :D


whot (approved):

whot:
merged, thanks!

PR #1320, opened 2022-06-22
Fix typo in tests (resolution-cycledown)

LinusCDE:
Hi,

stumbled upon this when building a bash script to apply my preferred settings using a bash script and used this file for looking up all the special actions.

I'm pretty sure this should not break anything as for example doing the following works, but the button does nothing:

```sh
ratbagctl "Logitech G903" button 7 action set special resolution-cycle-down
```

Also using `cycledown` in the command above doesn't work, as that's not defined somewhere else.

So take this change with a grain of salt, but it should resolve a bug when trying to use this.

---

Currently using this with `ratbagctl` and I'm loving this software btw! Not even the official GHub (in VM) doesn't seem to offer configuring both leds on G903 (UI bug it seems). This software is so great!!!


staticssleever668:
The edited function is only ran in tests, and this particular line doesn't actually ever get used. Still, I think good for QOL. :)

staticssleever668 (approved):

LinusCDE:
Ah cool.

Though the key didn't work on my mouse while the up-version did. So there might be still another typo/bug regarding this.

staticssleever668:
This may be a problem with the driver or maybe your mouse just doesn't support this action. I would suggest you opening a separate issue for that.

LinusCDE:
Good point. However I would be surprised if the mouse only supported `resolution-cycle-up` but not `resolution-cycle-down`. The non-cycle versions also work without issue.

LinusCDE:
Will look into this further and open an issue later if I conclude this should work (testing whether ghub has/allows it).

LinusCDE:
Searched a bit through the code. It seems the test may be the only occurrence and my mouse (Logitech G903 LS) likely just doesn't supports is.

Just tested in a VM and G-Hub only shows an assignable button called DPI Cycle, not any particular direction. So that likely just maps to cycle-up and the down feature doesn't exist for my mouse).

So this is likely just a typo in testing. Feel free to merge this as a typo fix or close it if invalid / not necessary to fix.

LinusCDE:
So my assumption that it should work and a typo like this just doesn't make it work was wrong. There is no separate issue or fix needed. Thanks again for the clarification!

whot:
Merged, thanks!

PR #1322, opened 2022-06-22
RFC: use function defined by DEFINE_TRIVIAL_CLEANUP_FUNC manually as well

staticssleever668:
While adding some new options to device files, I forgot to reset the `error` pointer after freeing it. Once there were `free` calls, I got double free, which took me a while to find.

This way we would never have to think about resetting the pointer, preventing potential double-free errors in the future. This also eliminates some code duplication.

### DIFF
--- a/src/libratbag-data.c
+++ b/src/libratbag-data.c
@@ -124,22 +124,17 @@ init_data_hidpp10(struct ratbag *ratbag,
 	num = g_key_file_get_integer(keyfile, group, "DeviceIndex", &error);
 	if (num != 0 || !error)
 		data->hidpp10.index = num;
-	if (error)
-		g_error_free(error);
+	g_error_freep(&error);
 
-	error = NULL;
 	num = g_key_file_get_integer(keyfile, group, "Profiles", &error);
 	if (num > 0 || !error)
 		data->hidpp10.profile_count = num;
-	if (error)
-		g_error_free(error);
+	g_error_freep(&error);
 
-	error = NULL;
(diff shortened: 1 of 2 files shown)


src/libratbag-data.c
```diff
@@ -124,22 +124,17 @@ init_data_hidpp10(struct ratbag *ratbag,
 	num = g_key_file_get_integer(keyfile, group, "DeviceIndex", &error);
 	if (num != 0 || !error)
 		data->hidpp10.index = num;
-	if (error)
-		g_error_free(error);
+	g_error_freep(&error);
```
whot, line 127: I'm sorry, but very much a NACK from me, this is likely going to mess up things long-term. Those free functions are "reserved" for exactly one use-case, messing with those means that, over time, we'll end up with something suboptimal and harder to read. Having said that, there's no need to even do this: `g_clear_error()` does exactly what we want :)

Alternatively, you could confine the error further in custom blocks:
```c
    int num;
    ...
    {
       g_autoptr(GError) error = NULL;
       num = g_get_file_get_integer(...)
       ...
   }
```

whot (changes requested):

src/libratbag-data.c
staticssleever668: `g_clear_error` seems nice, thanks! For some reason I keep getting lost in GLib's documentation, and I didn't see this function before...

Putting every key retrieval into blocks would be more modern in my opinion, but this pattern is not used in the code base much, so I think the `g_clear_error` fits better.

Issue #1253, opened 2022-01-03
Musl Libc Incompatibility


kwhat:
**Information**
- `ratbagd` version (`ratbagd --version`): 0.16
- Distribution: Gentoo
- Kernel version (ex. `uname -srmo`): `Linux 5.15.11-gentoo x86_64 GNU/Linux`

**Describe the bug**

../libratbag-0.16/tools/hidpp20-reset.c:26:10: fatal error: error.h: No such file or directory

**To Reproduce**
1. Compile against musl libc

**Logs**
N/A

**Additional context**
This may already be fixed in master, if not, it should be a pretty simple fix.  Just need to do something like this.  Having meson detect musl libc is a little more difficult.  I'll leave that part up to someone that knows meson better than I do. 

    #ifdef HAVE_ERROR_H
    #include "error.h"
    #else
    #include "err.h"
    #define error(status, errno, ...) err(status, __VA_ARGS__)
    #endif

whot:
Best I can tell we're not making huge use of `error()` anyway, might as well just rewrite the code to provide our own little function to do this. Pull requests accepted :)

listout:
> Best I can tell we're not making huge use of `error()` anyway, might as well just rewrite the code to provide our own little function to do this. Pull requests accepted :)

Hey, @whot is @kwhat's initial idea still acceptable? It would fix building for musl or other libc that doesn't provide error.h. Or is rewriting the error only acceptable idea for a PR. I've a patch for the first implementation i.e. using meson to detect if error.h is present on system.

whot:
There are only 5 invocations of `error()` in the code, afaict (`tools/hidpp20-dump-page.c` and `tools/hidpp20-reset.c`, both are developer utils, not even the main code). It's just a convenience function, those cases should be rewritten to `fprintf` and `exit`, there's no need to clutter up the code with ifdefs and meson conditions, that's counterproductive IMO.

listout:
@whot what if we check for error.h and if not present then we redefine error to use the err function?

Something like this?

```diff
--- a/meson.build
+++ b/meson.build
@@ -74,8 +74,13 @@ dep_libevdev = dependency('libevdev')
 dep_glib = dependency('glib-2.0')
 dep_json_glib = dependency('json-glib-1.0')
 dep_lm = cc.find_library('m')
+error_exists = cc.has_header('error.h')
 dep_unistring = cc.find_library('unistring')

+if error_exists
[...]

PR #1331, opened 2022-07-29
Use fprintf instead of error

listout:
hidpp files were using the error function from error.h header file.
Unfortunately not every libc comes with this particular header file, so
with this patch we won't be requiring to be rely on error function and
use fprintf function.

Thanks to developer Yaroslav Chvanov <staticssleever668> for the
original suggestion

Closes: https://github.com/libratbag/libratbag/issues/1253

Signed-off-by: listout <listout@protonmail.com>

### DIFF
--- a/tools/hidpp10-dump-page.c
+++ b/tools/hidpp10-dump-page.c
@@ -23,7 +23,6 @@
 
 #include <config.h>
 #include <errno.h>
-#include <error.h>
 #include <fcntl.h>
 
 #include <hidpp10.h>
@@ -94,8 +93,10 @@ main(int argc, char **argv)
 
 	path = argv[argc - 1];
 	fd = open(path, O_RDWR);
-	if (fd < 0)
-		error(1, errno, "Failed to open path %s", path);
+	if (fd < 0) {
+		fprintf(stderr, "Failed to open path '%s': %s", path, strerror(errno));
+		exit(3);
+	}
 
 	hidpp_device_init(&base, fd);
 	rc = hidpp10_device_new(&base, HIDPP_WIRED_DEVICE_IDX, HIDPP10_PROFILE_UNKNOWN, 5, &dev);
--- a/tools/hidpp20-dump-page.c
+++ b/tools/hidpp20-dump-page.c
@@ -23,7 +23,6 @@
 
 #include <config.h>
 #include <errno.h>
-#include <error.h>
 #include <fcntl.h>
 
 #include <hidpp20.h>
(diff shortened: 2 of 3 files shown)


tools/hidpp10-dump-page.c
```diff
@@ -95,7 +94,7 @@ main(int argc, char **argv)
 	path = argv[argc - 1];
 	fd = open(path, O_RDWR);
 	if (fd < 0)
-		error(1, errno, "Failed to open path %s", path);
+		fprintf(stderr, "Failed to open path '%s': %s", path, strerror(errno));
```
staticssleever668, line 97: You also need to add `exit(1);` after every replaced `error` call, as that's what it does - prints a message and exits the program.
So, the code would look like this:
 ```suggestion
		fprintf(stderr, "Failed to open path '%s': %s", path, strerror(errno));
		exit(1);
```
(looks like the indentation is misaligned in the suggestion, I think this is a problem with GitHub)

**Review by perigoso — APPROVED** [review]

File: tools/hidpp10-dump-page.c (hunk 1 above)
Comment (listout) [reply tools/hidpp10-dump-page.c:97]: @staticssleever668 Okay, thanks. I'll update the MR

Comment (listout) [reply tools/hidpp10-dump-page.c:97]: @staticssleever668 should it be `exit(3)`. Referring to this https://linux.die.net/man/3/error

Comment (staticssleever668) [reply tools/hidpp10-dump-page.c:97]: You are right, good catch. 👍🏿

Comment (listout) [reply tools/hidpp10-dump-page.c:97]: Done :)

File: tools/hidpp20-reset.c [hunk 2]
```diff
@@ -87,13 +86,17 @@ main(int argc, char **argv)
...
+		fprintf(stderr, "Failed to open path '%s': %s", path, strerror(errno));
+		exit(3);
+	}
 
 	hidpp_device_init(&base, fd);
 	dev = hidpp20_device_new(&base, 0xff, NULL, 0);
-	if (!dev)
-		error(1, 0, "Failed to open %s as a HID++ 2.0 device", path);
+	if (!dev) {
+		fprintf(stderr, "Failed to open %s as a HID++ 2.0 device: %s", path, strerror(0));
```
staticssleever668, line 97:
```suggestion
		fprintf(stderr, "Failed to open %s as a HID++ 2.0 device", path);
```

tools/hidpp20-dump-page.c
```diff
@@ -111,13 +110,17 @@ main(int argc, char **argv)
...
+		fprintf(stderr, "Failed to open path '%s': %s", path, strerror(errno));
+		exit(3);
+	}
 
 	hidpp_device_init(&base, fd);
 	dev = hidpp20_device_new(&base, 0xff, NULL, 0);
-	if (!dev)
-		error(1, 0, "Failed to open %s as a HID++ 2.0 device", path);
+	if (!dev) {
+		fprintf(stderr, "Failed to open %s as a HID++ 2.0 device: %s", path, strerror(0));
```
staticssleever668, line 121:
```suggestion
		fprintf(stderr, "Failed to open %s as a HID++ 2.0 device", path);
```

staticssleever668 (changes requested):
Didn't notice the use of `strerror(0)`.
It's unneeded as it just prints `Success`.

listout:
> Didn't notice the use of `strerror(0)`. It's unneeded as it just prints `Success`.

Done.

staticssleever668 (approved):

tools/hidpp10-dump-page.c
staticssleever668: It actually meant third section of man pages, not literally `exit(3)`, so it should have indeed been `exit(1)`...

Returning `1` is an informal way of saying there is any kind of error.
Not that it matters in any way, so keeping it at `exit(3)` is okay. :)

listout: Ahh, shoot. My bad. I need to get better at reading man pages. If you want I can definitely send a second PR to fix this.

staticssleever668: No, i's good, as I said, it's not like it really matters, so it's okay. :)

PR #1332, opened 2022-08-01
sinowealth: move device-specific data out of the driver

staticssleever668:
This moves device data out of the driver into device files. However, a
_single_ device file contains data for _several mice_ at the same time.
The reason for this is that different mice by SinoWealth may use the
same VID:PID pair. This is unfortunate, as we only allow a single
device file a given VID:PID pair.

@whot, what do you think of such a way? I've looked into making it possible to split it further so that each device has it's own device file by allowing several devices with the same VID:PID pair, but it turned out to be quite complex. Doing it like so on contrary was pretty easy.

Unrelated: also move `sinowealth_led_format` enum into a header file and rename
its constants to fit in outside of the driver.

### DIFF
--- a/data/devices/sinowealth-1007.device
+++ b/data/devices/sinowealth-1007.device
@@ -1,0 +1,20 @@
+[Device]
+DeviceMatch=usb:258a:1007
+Driver=sinowealth
+Name=SinoWealth Generic Mouse (1007)
+
+# This does NOT work. See libratbag/libratbag#1296.
+# For future: ID of A3050 sensor is 0x2, and maximum DPI is 4000.
+#[Driver/sinowealth/devices/AntEsportsGM500]
+#ButtonCount=7
+#DeviceName=Ant Esports GM500
+#FwVersion=0910
+#LedType=RBG
+#SensorType=A3050
+
+[Driver/sinowealth/devices/MarvoScorpionG961]
+ButtonCount=6
+DeviceName=Marvo Scorpion G961
+FwVersion=9677
+LedType=RGB
+SensorType=PMW3327
--- a/src/driver-sinowealth.c
+++ b/src/driver-sinowealth.c
@@ -1529,11 +1529,11 @@ sinowealth_init_profile(struct ratbag_device *device)
 	 */
 	struct sinowealth_config_report *config = &drv_data->configs[0];
 
-	char fw_version[SINOWEALTH_FW_VERSION_LEN] = { 0 };
+	char fw_version[SINOWEALTH_FW_VERSION_LEN + 1] = { 0 };
 	rc = sinowealth_get_fw_version(device, fw_version);
 	if (rc)
 		return rc;
-	log_info(device->ratbag, "firmware version: %.4s\n", fw_version);
+	log_info(device->ratbag, "firmware version: %s\n", fw_version);
 
 	rc = sinowealth_read_raw_config(device);
 	if (rc)
--- a/src/driver-sinowealth.c
+++ b/src/driver-sinowealth.c
@@ -21,6 +21,9 @@
  * DEALINGS IN THE SOFTWARE.
  */
 
+#include "driver-sinowealth.h"
+
+#include "libratbag-data.h"
 #include "libratbag-private.h"
 #include "libratbag-hidraw.h"
 #include "shared-macro.h"
@@ -187,12 +190,6 @@ union sinowealth_dpis {
 	struct sinowealth_xy_independent_dpi independent[8];
 };
 _Static_assert(sizeof(union sinowealth_dpis) == sizeof(uint8_t[16]), "Invalid size");
-
-enum sinowealth_led_format {
-	LED_NONE,
-	LED_RBG,
-	LED_RGB,
-};
 
 /* Configuration data the way it's stored in mouse memory.
  * When we want to change a setting, we basically copy the entire mouse
@@ -651,12 +648,12 @@ sinowealth_raw_to_color(struct ratbag_device *device, struct sinowealth_color ra
 	switch (drv_data->led_type) {
 	/* Fall back to RBG if the LED type is incorrect. */
 	default:
-	case LED_RBG:
+	case SINOWEALTH_LED_TYPE_RBG:
 		color.red = raw_color.data[0];
 		color.green = raw_color.data[2];
 		color.blue = raw_color.data[1];
 		break;
-	case LED_RGB:
+	case SINOWEALTH_LED_TYPE_RGB:
 		color.red = raw_color.data[0];
 		color.green = raw_color.data[1];
 		color.blue = raw_color.data[2];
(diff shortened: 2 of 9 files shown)


src/driver-sinowealth.c
```diff
@@ -1373,140 +1370,37 @@ sinowealth_update_macro_from_action(struct ratbag_profile *profile, struct ratba
...
-		.device_name = "Generic SinoWealth Mouse",
-		.led_type = LED_RBG,
-		.sensor_type = SINOWEALTH_SENSOR_PMW3360,
-		.vid = 0xffff,
-		.pid = 0xffff,
-		.button_count = 6,
-	},
-};
 
+// @return Supported device data for the device or `NULL`.
```
whot, line 1374: I think we use `/* */` comments exclusively, please change to that.

src/driver-sinowealth.c
```diff
@@ -1373,140 +1370,37 @@ sinowealth_update_macro_from_action(struct ratbag_profile *profile, struct ratba
...
 
 		return device_data;
 	}
 
-	/* Return the last entry in `sinowealth_device_data` - a dummy. */
-	return device_data;
+	const struct sinowealth_device_data *fallback_device_data = ratbag_device_data_sinowealth_get_fallback_device(data);
+	if (fallback_device_data != NULL) {
+		log_info(device->ratbag, "Your mouse is not supported officially, but still seems to work fine.\n");
+		log_info(device->ratbag, "Consider reporting model name and above firmware version to make it official.\n");
```
whot, line 1398: I wouldn't use the term `officially` here, since we're not sinowealth. How about something like:
```
This device is not known to libratbag, using fallback behavior. Please report model name and above firmware version to the libratbag project.
```
Or something along these lines anyway.

I'd also be careful with the term `supported` since we don't really support anything. We know of it and maybe find time to help fix it, but that's about it :)

src/libratbag-data.c
```diff
@@ -78,6 +79,12 @@ struct data_hidpp10 {
 	int led_count;
 };
 
+struct data_sinowealth {
+	/* @note May be `NULL` if fallback is disallowed. */
```
whot, line 83: No need for `@note` here, this isn't doxygen-parsed anyway.

src/libratbag-data.c
```diff
@@ -207,6 +215,69 @@ init_data_hidpp20(struct ratbag *ratbag,
 	}
 }
 
+static void
+init_data_sinowealth(struct ratbag *ratbag,
+		     GKeyFile *keyfile,
+		     struct ratbag_device_data *data)
```
whot, line 221: probably just github's rendering, but the indentation looks off here.

data/devices/sinowealth-1007.device
```diff
@@ -0,0 +1,23 @@
+[Device]
+DeviceMatch=usb:258a:1007
+Driver=sinowealth
+Name=SinoWealth Generic Mouse (1007)
+
+[Driver/sinowealth]
+DisableFallback=1
```
whot, line 7: One lesson I learned [1]: negative options get confusing quickly. `DisableFallback=false` takes too much brain power to be reliably implemented. I'd go for an `EnableFallback` instead and you'll find the code becomes a lot easier to read.


[1] you can specify "NoDontZap false" and it's valid configuration in xorg.conf  :)

src/libratbag-data.c
```diff
@@ -207,6 +215,69 @@ init_data_hidpp20(struct ratbag *ratbag,
...
+
+	int disable_fallback = g_key_file_get_integer(keyfile, group, "DisableFallback", &error);
+	if (!disable_fallback) {
+		data->sinowealth.fallback_device = zalloc(sizeof(struct sinowealth_device_data));
+		struct sinowealth_device_data *device = data->sinowealth.fallback_device;
+
+		device->button_count = 6;
+		device->device_name = "Generic SinoWealth Mouse";
+		device->fw_version = "0000";
+		device->led_type = SINOWEALTH_LED_TYPE_RBG;
```
whot, line 239: feel free to ignore this: do you want to specify the fallback device in the respective device data file too? Because there's a chance they may need to differ anyway (eventually) between the two USB PIDs

src/libratbag-data.c
```diff
@@ -207,6 +215,69 @@ init_data_hidpp20(struct ratbag *ratbag,
...
+	}
+	g_clear_error(&error);
+
+	list_init(&data->sinowealth.supported_devices);
+
+	for (size_t i = 0; i < group_count; ++i) {
+		const char *device_group = groups[i];
+		if (!strneq(device_group, devices_group, strlen(devices_group)))
```
whot, line 247: fwiw, you don't need to use `strneq()` for "nice" strings, you can use `streq()` instead. `strneq()` is just for where you have at least one string that isn't NULL-terminated or where you just want to compare a prefix.

src/driver-sinowealth.c
```diff
@@ -1527,7 +1529,7 @@ sinowealth_init_profile(struct ratbag_device *device)
 	 */
 	struct sinowealth_config_report *config = &drv_data->configs[0];
 
-	char fw_version[4];
+	char fw_version[SINOWEALTH_FW_VERSION_LEN];
```
whot, line 1426:
```c
char fw_version[SINOWEALTH_FW_VERSION_LEN + 1] = {0};
```
and bam, you have a guaranteed null-terminated string and don't need to worry about the exact length anymore.

whot (approved):
Yes! this is exactly how the device data files should be used. A few nitpicks as usual, but overall this is a big improvement, should be possible to add new devices now without having to modify code.

src/libratbag-data.c
staticssleever668: Here I iterate over *all* groups in the device file and check if the current one is a part of `Driver/sinowealth/devices/`, not if they are completely the same, so we *do* have to use `strneq`.
Maybe there is a GLib method to get only needed groups, but I couldn't find anything. Do you know anything like that?

src/libratbag-data.c
staticssleever668: I don't think it's needed with the fact some of the mice with the same PIDs have different parameters, so we would have to do a lucky guess in every single device file.

data/devices/sinowealth-1007.device
staticssleever668: Actually, I think I'll just get rid of the fallback thing.
Although unlikely, some devices may look like they support the driver, but actually don't. If a user wants to experiment with the mouse and check if it can be supported, they can add the entry to the device file themselves. This would also encourage them to report device data here.

src/driver-sinowealth.c
staticssleever668: Super nice!

src/libratbag-data.c
staticssleever668: I'm not sure how to indent it manually, I just used clang-format file from the Linux project, which more or less works fine in libratbag. Is it okay now?

src/libratbag-data.c
whot: ah, I misread then, apologies. I don't think there's a method for that in the GLib code (we're using .ini files a bit outside what they're originally designed for) but we *do* have `startswith()` in `shared-macro.h` which will make the code more readable for this case.

src/libratbag-data.c
whot: > I'm not sure how to indent it manually, 

I.. wait, what? :smile: Just use tabs and then spaces to align the two?

staticssleever668: I think it's aligned now.
If it doesn't look right in the diff, you can open the file and see that it's correct.

src/libratbag-data.c
staticssleever668: Great, thanks!

src/libratbag-data.c
whot: ah, right. the `.editorconfig` states how things should expand, pretty sure these days tabs other than 8 spaces is going to look ugly in most projects. Unfortunately, github renders tabs as 4 spaces which makes everything a bit messy. Looks correct now though, thanks!

PR #1355, opened 2022-10-10
Add Support for G502 X over USB

andigandi:
Just received my G502 X and saw it is not yet supported. This trivial change made ratbagd recognize it. Thus pull request addresses the "basic" version of the G502 X connected via USB.

### DIFF
--- a/data/devices/logitech-g502-x.device
+++ b/data/devices/logitech-g502-x.device
@@ -1,0 +1,4 @@
+[Device]
+Name=Logitech G502 X
+DeviceMatch=usb:046d:c099
+Driver=hidpp20


staticssleever668 (approved):

staticssleever668:
Nice!

YamiYukiSenpai:
what version would have the G502 X?

staticssleever668:
@YamiYukiSenpai, it's going to be in the next release, which may be months from now.
For now you can make this file locally. It depends on the distro, but usually it goes to `/usr/share/libratbag` or `/usr/local/share/libratbag` where other `*.device` files are.

YamiYukiSenpai:
Sorry.  I got confused and I was actually inquiring about G502 X Plus

PR #1364, opened 2022-11-04
Implement Mars Gaming MM4 protocol 

### LINKED ISSUE #1199: Support for Mars Gaming MM4
**Describe the solution you'd like**
Add support for Mars Gaming MM4, with enough flexibility for possible support for other Mars Gaming mice.

**Describe alternatives you've considered**
I have reverse engineered the communication protocol between the mouse and the windows configurator.
I have implemented most of the functionality at [Qjammer/libratbag@marsgaming-mm4](https://github.com/Qjammer/libratbag/tree/marsgaming-mm4). 

**Additional context**
I will be opening a PR soon with support for this mouse.
Related: https://github.com/libratbag/piper/issues/676

Qjammer:
closes #1199

This PR implements most of the protocol for Mars Gaming MM4 mice.

This includes:
Querying
- Profiles
- Report rate
- Resolution
- Buttons (excluding full macros)
- LEDs

Writing
- Profiles
- Report rate
- Resolution
- Buttons (including macros up to 2 keys + mods)
- LEDs


Qjammer:
The pipeline complains about `ERROR: Could not detect Ninja v1.8.2 or newer`. It seems to be installing v1.11.1, which should provide this dependency. It doesn't seem to make sense. How can I fix this?

staticssleever668:
Sorry for taking so long, I'll check this first thing when I have a minute for a thorough review.

Without looking in-depth, I noticed following:

1) Usage of `typedef` to define enums and structs. This is perfectly fine, but we don't really use it in libratbag (`grep "typedef" src/* ratbagd/*` yields only 7 results, of which only 2 are used in the same context as in this PR). Please, replace it with fully qualified `enum marsgaming_x` and `struct marsgaming_x`.
2) There is `static_assert` used, which was only added in C11, while we use C99 (GNU flavor), so, please, replace them with `_Static_assert`.
3) We have a stylistic preference to put function return types on a separate line, like in the Linux project.
E.g.:
This
```c
static void do_stuff()
{
  // stuff
}
```

Will become this
```c
static void
do_stuff()
{
  // stuff
}
```

4) There is a lot of function forward declarations, as I see it, this is because the implementations are in the header files. Could you move them to separate source files? (So every header file in `src/driver-marsgaming` folder will have a matching `*.c` file.

5) IMO we don't need the new `.gitignore` as it only contains environment-specific file paths. I think things like these should go in your local gitignore file, by default it's in `~/.config/git/ignore`.

Qjammer:
I tried to modify the code as per requested. Would you be able to give some more guidance on how to proceed?

src/driver-marsgaming/marsgaming-buttons.c
```diff
@@ -0,0 +1,425 @@
...
+	const struct marsgaming_button_action_mapping *button_mapping;
+	ARRAY_FOR_EACH(marsgaming_mm4_button_action_mapping, button_mapping) {
+		if (button_mapping->marsgaming_action_id == button_info->action)
+			return button_mapping->ratbag_action;
+	}
+	return (struct ratbag_button_action)BUTTON_ACTION_UNKNOWN;
+}
+
+static int
+ratbag_button_macro_from_marsgaming_combo_keycode(struct ratbag_button *button, unsigned int key1, unsigned int key2, unsigned int modifiers);
```
staticssleever668, line 39: These function names are misleading as they are provided by the marsgaming driver and not libratbag.
This one could be `marsgaming_ratbag_button_macro_from_combo_keycode` for example.

src/driver-marsgaming/marsgaming-buttons.c
```diff
@@ -0,0 +1,425 @@
...
+			}
+
+		};
+	}
+	// TODO: Implement macros
+	return (struct marsgaming_button_info)MARSGAMING_BUTTON_ACTION_NONE;
+}
+
+static int
+ratbag_action_keycodes_from_macro(struct ratbag_button_action *action, unsigned int *key0_out, unsigned int *key1_out, unsigned int *modifiers_out)
```
staticssleever668, line 220: Note to self: take a look at this function again later.
Also, return value of this function is confusing. it would make sense if the output was written to an array, but here two separate arguments are used.

src/driver-marsgaming/marsgaming-buttons.c
```diff
@@ -0,0 +1,425 @@
...
+	if (e == 1) { // We can convert this to single/combo key
+		return (struct marsgaming_button_info){
+			.action = MARSGAMING_MM4_ACTION_COMBO_KEY,
+			.action_info.combo_key = {
+				.modifiers = mods,
+				.keys[0] = ratbag_hidraw_get_keyboard_usage_from_keycode(button->profile->device, key0),
+			}
+
+		};
+	}
```
staticssleever668, line 201: Is this whole case even needed? Look like `ratbag_action_keycodes_from_macro` would do the same thing in the end.

src/driver-marsgaming/marsgaming-buttons.c
```diff
@@ -0,0 +1,425 @@
...
+	{ MARSGAMING_MM4_ACTION_FORWARD, marsgaming_button_action_lookup },
+	{ MARSGAMING_MM4_ACTION_DPI_SWITCH, marsgaming_button_action_lookup },
+	{ MARSGAMING_MM4_ACTION_DPI_MINUS, marsgaming_button_action_lookup },
+	{ MARSGAMING_MM4_ACTION_DPI_PLUS, marsgaming_button_action_lookup },
+	{ MARSGAMING_MM4_ACTION_PROFILE_SWITCH, marsgaming_button_action_lookup },
+	//{ MARSGAMING_MM4_ACTION_MEDIA, marsgaming_button_action_media },
+	{ MARSGAMING_MM4_ACTION_COMBO_KEY, marsgaming_button_action_key },
+	{ MARSGAMING_MM4_ACTION_SINGLE_KEY, marsgaming_button_action_key },
+	//{ MARSGAMING_MM4_ACTION_MACRO, marsgaming_button_action_macro }, // TODO: Implement macros
+	//{ MARSGAMING_MM4_ACTION_FIRE, marsgaming_button_action_fire },
```
staticssleever668, line 320: There are no `marsgaming_button_action_media`, `marsgaming_button_action_macro`  and `marsgaming_button_action_fire`. Maybe stub them out and add TODOs in there with short description of what it should be?

They could look like this for now:

```c
static struct ratbag_button_action
marsgaming_button_action_media(struct ratbag_button *button,
			     const struct marsgaming_button_info *button_info)
{
	// TODO: something-something description of what should be done here.
	return (struct ratbag_button_action)BUTTON_ACTION_UNKNOWN;
}
```

src/driver-marsgaming/marsgaming-buttons.c
```diff
@@ -0,0 +1,425 @@
...
+unsigned
+marsgaming_button_type(unsigned button_index)
+{
+	const struct marsgaming_button_mapping *mapping;
+	ARRAY_FOR_EACH(marsgaming_mm4_button_mapping, mapping) {
+		if (mapping->button_id == button_index)
+			return mapping->type;
+	}
+	return RATBAG_BUTTON_TYPE_UNKNOWN;
+}
```
staticssleever668, line 394: `enum ratbag_button_type` is deprecated so all of this is not useful, sorry.

src/driver-marsgaming/marsgaming-command.c
```diff
@@ -0,0 +1,92 @@
...
+	struct marsgaming_report_resolutions report = marsgaming_profile_get_drv_data(profile)->resolutions_report;
+	report.report_type = MARSGAMING_MM4_REPORT_TYPE_WRITE;
+	report.unknown_6 = 0xfa;
+	report.unknown_7 = 0xfa;
+
+	for (int i = 0; i < report.count_resolutions; ++i) {
+		// Handle endianness
+		set_unaligned_le_u16((uint8_t*)&report.resolutions[i].x_res, report.resolutions[i].x_res);
+		set_unaligned_le_u16((uint8_t*)&report.resolutions[i].y_res, report.resolutions[i].y_res);
+	}
```
staticssleever668, line 79: This looks out of place. I think this should be done where the value is modified, in `marsgaming_commit_profile_resolutions`.

Also, is this really needed? (Never handled endianness so I'll just go with what you think is right)

src/driver-marsgaming/marsgaming-commit.c
```diff
@@ -0,0 +1,144 @@
...
+	return 0;
+}
+
+static int
+marsgaming_commit_profiles(struct ratbag_device *device)
+{
+	uint8_t current_profile = marsgaming_query_current_profile(device);
+	struct ratbag_profile *profile;
+	ratbag_device_for_each_profile(device, profile) {
+		profile->is_active = current_profile == profile->index; // TODO: Check if we should be modifying this
```
staticssleever668, line 121: Looks like this will simply mark every single profile as active, which is incorrect, so probably this is not needed.

src/driver-marsgaming/marsgaming-leds.h
```diff
@@ -0,0 +1,29 @@
...
+	MARSGAMING_LED_BRIGHTNESS_MEDIUM = 0x02,
+	MARSGAMING_LED_BRIGHTNESS_HIGH = 0x03
+};
+
+#define MARSGAMING_LED_BREATHING_OFF 0
+
+struct marsgaming_led_color {
+	uint8_t red; // 0xff - red
+	uint8_t green; // 0xff - green
+	uint8_t blue; // 0xff - blue
```
staticssleever668, line 17: These comments are confusing, they seem like description of the value, like, when it's 0xff, the LED is bright, but if I understand correctly, it's the other way around.

src/driver-marsgaming/marsgaming-leds.h
```diff
@@ -0,0 +1,29 @@
...
+};
+
+struct marsgaming_led {
+	struct marsgaming_led_color color;
+	uint8_t brightness; // marsgaming_led_brightness_intensity
+	uint8_t breathing_speed; // marsgaming_led_breathing_speed
```
staticssleever668, line 23: I think these should have these enums as the type, although I'm not 100%.
(would be nice if someone more experienced in C could confirm this).

staticssleever668: Also, there is no `enum marsgaming_led_breathing_speed`.

src/driver-marsgaming/marsgaming-probe.c
```diff
@@ -0,0 +1,159 @@
...
+		}
+	}
+}
+
+static void
+marsgaming_probe_button(struct ratbag_button *button, struct marsgaming_button_info *button_info)
+{
+	ratbag_button_enable_action_type(button, RATBAG_BUTTON_ACTION_TYPE_BUTTON);
+	ratbag_button_enable_action_type(button, RATBAG_BUTTON_ACTION_TYPE_SPECIAL);
+	ratbag_button_enable_action_type(button, RATBAG_BUTTON_ACTION_TYPE_KEY); // Deprecated
```
staticssleever668, line 51: Why `Deprecated`?

src/driver-marsgaming/marsgaming-probe.c
```diff
@@ -0,0 +1,159 @@
...
+	ratbag_button_enable_action_type(button, RATBAG_BUTTON_ACTION_TYPE_MACRO);
```
staticssleever668, line 52: I can't find if these two are handled anywhere, if they aren't, these two lines should be removed.

src/driver-marsgaming/marsgaming-protocol.h
```diff
@@ -0,0 +1,90 @@
...
+
+enum marsgaming_report_type {
+	MARSGAMING_MM4_REPORT_TYPE_UNKNOWN_1 = 0x01,
+	MARSGAMING_MM4_REPORT_TYPE_WRITE = 0x02,
+	MARSGAMING_MM4_REPORT_TYPE_READ = 0x03,
+	MARSGAMING_MM4_REPORT_TYPE_UNKNOWN_4 = 0x04,
+	MARSGAMING_MM4_REPORT_TYPE_UNKNOWN_6 = 0x06
+};
+
+struct marsgaming_report_resolution_info {
```
staticssleever668, line 13: All the `_info`  sufixes are a mouthful, maybe they are not needed?
(No need for action, just a suggestion)

src/driver-marsgaming/marsgaming-protocol.h
```diff
@@ -0,0 +1,90 @@
...
+	uint8_t profile_id;
+	uint8_t unknown_4; // 0x4d
+	uint8_t unknown_5; // 0x00
+	uint8_t unknown_6; // 0x00 from device | 0xfa from host
+	uint8_t unknown_7; // 0x00 from device | 0xfa from host
+	uint8_t button_count;
+	union {
+		struct marsgaming_button_info buttons[253];
+		uint8_t padding[1015];
+	};
```
staticssleever668, line 58: Same as the above, these fields have different sizes, and it also seems like this doesn't have to be a union at all.

staticssleever668 (changes requested):
Looks good and clean, thank you and sorry for the delay in reviewing this. 

By the way, there are so many commits almost to the point of spam, like almost 10 instances `Marsgaming Return type on new line`, does this imply you are expecting the commits squashed? (They don't appear to be in form of incremental changes anyway)

src/driver-marsgaming/marsgaming-buttons.c
Qjammer: Should I not set ratbag_button::type then?

src/driver-marsgaming/marsgaming-leds.h
Qjammer: I tried using enums, but I didn't find any way of specifying the backing type for the enum, so using it in usb reports directly might make this code platform-dependent.

staticssleever668: You need `packed` attribute and an assertion to do a size check.
Here is an example of an enum from the sinowealth driver:
```c
enum sinowealth_report_id {
	SINOWEALTH_REPORT_ID_CONFIG = 0x4,
	SINOWEALTH_REPORT_ID_CMD = 0x5,
	SINOWEALTH_REPORT_ID_CONFIG_LONG = 0x6,
} __attribute__((packed));
_Static_assert(sizeof(enum sinowealth_report_id) == sizeof(uint8_t), "Invalid size");
```

src/driver-marsgaming/marsgaming-buttons.c
staticssleever668: Yes.
I believe it's not used anywhere anyway.

src/driver-marsgaming/marsgaming-protocol.h
Qjammer: The padding is so that the object size is exactly 1024. An array of `marsgaming_button_info` does not fit evenly inside the report, there would be 3 bytes left over, but the report still has to be 1024 to avoid reading or writing to unallocated memory. Is there any other way of doing this?

src/driver-marsgaming/marsgaming-probe.c
Qjammer: I was told [here](https://github.com/libratbag/libratbag/issues/1199#issuecomment-923089817) that `key` functionality was deprecated. I will remove this line, since it isn't even implemented.

src/driver-marsgaming/marsgaming-leds.h
Qjammer: > By the way, there are so many commits almost to the point of spam, like almost 10 instances `Marsgaming Return type on new line`, does this imply you are expecting the commits squashed? (They don't appear to be in form of incremental changes anyway)

Yes, I was expecting the commits to be squashed.

src/driver-marsgaming/marsgaming-command.c
Qjammer: Hmm it does look weird. I think my intention was to handle endianness only when sending it through the hidraw api, to keep most of the program in native endianness. I'll make a copy of the report so that the original in drv_data is not modified.

src/driver-marsgaming/marsgaming-commit.c
Qjammer: I don't think that's the case, no? Only the profile whose index matches `current_profile` will be set as active, right?

The TODO was not about this question though. The issue comes when the user modifies the active profile after probe but before commit. Only the led of the active profile should be changed in `marsgaming-command.c:43`, but I don't know if modifying this here can cause any issues.

src/driver-marsgaming/marsgaming-leds.h
Qjammer: Your understanding is correct. I will modify the comments to make it clearer.

src/driver-marsgaming/marsgaming-command.c
Qjammer: Whether it is needed or not is debatable. Big endian systems are so rare that it might not be worth it to care.

src/driver-marsgaming/marsgaming-buttons.c
Qjammer: I renamed all functions that started with `ratbag_` to `marsgaming_` to avoid confusion.

src/driver-marsgaming/marsgaming-protocol.h
staticssleever668: You can remove the union and just add padding at the end:
```c
	struct marsgaming_button_info buttons[253];
	uint8_t padding[3];
```

By the way, I think this struct is also missing ` __attribute__((packed))` :)

src/driver-marsgaming/marsgaming-protocol.h
```diff
@@ -0,0 +1,90 @@
...
+	uint8_t unknown_4; // 0x2a
+	uint8_t unknown_5; // 0x00
+	uint8_t unknown_6; // 0x00 from device | 0xfa from host
+	uint8_t unknown_7; // 0x00 from device | 0xfa from host
+	uint8_t count_resolutions;
+	uint8_t current_resolution;
+	union {
+		struct marsgaming_report_resolution_info resolutions[6];
+		uint8_t padding[54];
+	};
```
staticssleever668, line 40:
```c
	struct marsgaming_report_resolution_info resolutions[6];
	uint8_t padding[6];
```
Here too, you can put the padding after the data.

https://github.com/libratbag/libratbag/pull/1364/files/14a2a85dd0f40a18afac41761aeff69339608552#r1090238728

src/driver-marsgaming/marsgaming-protocol.h
Qjammer: I could replace them with `_payload` or `_contents`, but I want to make it explicit that it's not the whole report, just the data relating to resolutions. I'm choosing to leave it as is.

src/driver-marsgaming/marsgaming-buttons.c
Qjammer: I deleted this function, I'm resolving this conversation.

src/driver-marsgaming/marsgaming-commit.c
staticssleever668: > Looks like this will simply mark every single profile as active, which is incorrect, so probably this is not needed.

I was wrong here.

---

> The TODO was not about this question though

Could you put this explanation to the TODO itself? Otherwise it's not really clear what it means.

src/driver-marsgaming/marsgaming-probe.c
```diff
@@ -0,0 +1,150 @@
...
+		resolution->dpi_y = queried_resolution.y_res * MARSGAMING_MM4_RES_SCALING;
+		ratbag_resolution_set_cap(resolution, RATBAG_RESOLUTION_CAP_SEPARATE_XY_RESOLUTION);
+	}
+}
+
+static void
+marsgaming_probe_profile_report_rate(struct ratbag_profile *profile)
+{
+	static unsigned int rates[] = { 125, 250, 500, 1000 };
+	ratbag_profile_set_report_rate_list(profile, rates, ARRAY_LENGTH(rates));
```
staticssleever668, line 81: Since then `ratbag_profile_set_report_rate_list` was modified to take a const pointer to rates, so you can make `rates` const.

```suggestion
	static const unsigned int rates[] = { 125, 250, 500, 1000 };
	ratbag_profile_set_report_rate_list(profile, rates, ARRAY_LENGTH(rates));
```

src/driver-marsgaming/marsgaming-query.c
```diff
@@ -0,0 +1,71 @@
...
+	ratbag_hidraw_get_feature_report(device, 0x02, readdata, ARRAY_LENGTH(readdata));
+
+	return readdata[8];
+}
+
+struct marsgaming_report_resolutions
+marsgaming_query_profile_resolutions(struct ratbag_profile *profile)
+{
+	uint8_t writedata[16] = { 0x02, 0x03, 0x4f, /* profile */ 0x00, 0x2a, 0x00, 0xfa, 0xfa } /* zeros */;
+	writedata[3] = profile->index;
```
staticssleever668, line 19: IMO it's more clear to set profile index directly in the initializer:
```suggestion
	uint8_t writedata[16] = {
		0x02, 0x03, 0x4f, profile->index, 0x2a, 0x00, 0xfa, 0xfa,
	};
```
Also, what does `zeros` comment mean?

src/driver-marsgaming/marsgaming-query.c
```diff
@@ -0,0 +1,71 @@
...
+	}
+
+	return report;
+}
+
+struct marsgaming_report_buttons
+marsgaming_query_profile_buttons(struct ratbag_profile *profile)
+{
+	uint8_t writedata[16] = { 0x02, 0x03, 0x90, /*profile*/ 0x00, 0x4d, 0x00, 0xfa, 0xfa };
+	writedata[3] = (uint8_t)profile->index;
```
staticssleever668, line 38: Like above, I think it's more clear to have profile index directly in the initializer.

src/driver-marsgaming/marsgaming-query.c
```diff
@@ -0,0 +1,71 @@
...
+	ratbag_hidraw_get_feature_report(profile->device, 0x02, readdata, ARRAY_LENGTH(readdata));
+
+	return readdata[8];
+}
+
+struct marsgaming_report_led
+marsgaming_query_profile_led(struct ratbag_profile *profile)
+{
+	uint8_t writedata[16] = { 0x02, 0x03, 0xf1, /*profile*/ 0x00, 0x06, 0x00 /*led_id???*/, 0xfa, 0xfa };
+	writedata[3] = (uint8_t)profile->index;
```
staticssleever668, line 64: Here too, "I think it's more clear to have profile index directly in the initializer.".

staticssleever668 (commented):
Just some nit picking, the only blocker is https://github.com/libratbag/libratbag/pull/1364#discussion_r1165484567
That and I want to take a look at this later with -Wconversions and it's good to go
Thank you for your work!

src/driver-marsgaming/marsgaming-probe.c
```diff
@@ -0,0 +1,150 @@
...
+}
+
+static void
+marsgaming_probe_profile_leds(struct ratbag_profile *profile)
+{
+	struct marsgaming_profile_drv_data *profile_data = marsgaming_profile_get_drv_data(profile);
+	profile_data->led_report = marsgaming_query_profile_led(profile);
+	struct ratbag_led *led;
+	ratbag_profile_for_each_led(profile, led) {
+		led->type = RATBAG_LED_TYPE_LOGO;
```
staticssleever668, line 21: Remove this line, libratbag doesn't have LED types anymore and it causes a compilation error.
```suggestion
```

data/devices/marsgaming-mm4.device
```diff
@@ -0,0 +1,4 @@
+[Device]
```
staticssleever668, line 1: Add `DeviceType=mouse` somewhere here, otherwise libratbag will ignore this device file (since #931). This also causes test failures.

src/driver-marsgaming/marsgaming-buttons.c
```diff
@@ -0,0 +1,377 @@
...
+			case KEY_RIGHTSHIFT: modifiers &= ~MODIFIER_RIGHTSHIFT; break;
+			case KEY_RIGHTALT: modifiers &= ~MODIFIER_RIGHTALT; break;
+			case KEY_RIGHTMETA: modifiers &= ~MODIFIER_RIGHTMETA; break;
+			default:
+				// As soon as we release a key that we pressed, return what we have processed so far
+				if (event.event.key == key0 || event.event.key == key1) {
+					*key0_out = key0;
+					*key1_out = key1;
+					*modifiers_out = modifiers;
+					return keys_pressed;
```
staticssleever668, line 297:
```suggestion
					return (int)keys_pressed;
```

src/driver-marsgaming/marsgaming-query.c
staticssleever668: `profile->index` in `unsigned int`, so also cast it to uint8_t:
```suggestion
	uint8_t writedata[16] = {
		0x02, 0x03, 0x4f, (uint8_t)profile->index, 0x2a, 0x00, 0xfa, 0xfa,
	};
```

Qjammer: I agree it is more clear. I'll change it. The `zeros` comment means that the rest of the bytes that are not explicitly initialized are set to zero. I can remove it

src/driver-marsgaming/marsgaming-buttons.c
```diff
@@ -0,0 +1,377 @@
...
+
+static const struct marsgaming_modifier_mapping marsgaming_modifier_mapping[] = {
+	{ MODIFIER_LEFTCTRL, KEY_LEFTCTRL },
+	{ MODIFIER_LEFTSHIFT, KEY_LEFTSHIFT },
+	{ MODIFIER_LEFTALT, KEY_LEFTALT },
+	{ MODIFIER_LEFTMETA, KEY_LEFTMETA },
+	{ MODIFIER_RIGHTCTRL, KEY_RIGHTCTRL },
+	{ MODIFIER_RIGHTSHIFT, KEY_RIGHTSHIFT },
+	{ MODIFIER_RIGHTALT, KEY_RIGHTALT },
+	{ MODIFIER_RIGHTMETA, KEY_RIGHTMETA },
```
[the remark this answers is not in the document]
Qjammer: Done. I created a separate enum for the marsgaming modifiers.

staticssleever668:
Could you also take a look at https://github.com/libratbag/libratbag/pull/1364#discussion_r1165490225?

staticssleever668 (approved):

src/driver-marsgaming/marsgaming-commit.c
Qjammer: I have expanded on the comment to make the intent clearer. The doubt still remains, although during development I did not see any unexpected behavior.

staticssleever668:
Great, thank you!!!

PR #1372, opened 2022-11-14
Device firmware version

staticssleever668:
Supersedes and closes #1068.
Following whot/ratbag-python#27.

### DIFF
--- a/ratbagd/ratbagd-device.c
+++ b/ratbagd/ratbagd-device.c
@@ -220,10 +220,27 @@ ratbag_device_get_model(sd_bus *bus,
 	return sd_bus_message_append(reply, "s", model);
 }
 
+static int
+ratbagd_device_get_firmware_version(sd_bus *bus,
+				    const char *path,
+				    const char *interface,
+				    const char *property,
+				    sd_bus_message *reply,
+				    void *userdata,
+				    sd_bus_error *error)
+{
+	struct ratbagd_device *device = userdata;
+	struct ratbag_device *lib_device = device->lib_device;
+	const char* version = ratbag_device_get_firmware_version(lib_device);
+
+	return sd_bus_message_append(reply, "s", version);
+}
+
 const sd_bus_vtable ratbagd_device_vtable[] = {
 	SD_BUS_VTABLE_START(0),
 	SD_BUS_PROPERTY("Model", "s", ratbag_device_get_model, 0, SD_BUS_VTABLE_PROPERTY_CONST),
 	SD_BUS_PROPERTY("Name", "s", ratbagd_device_get_device_name, 0, SD_BUS_VTABLE_PROPERTY_CONST),
+	SD_BUS_PROPERTY("FirmwareVersion", "s", ratbagd_device_get_firmware_version, 0, SD_BUS_VTABLE_PROPERTY_CONST),
 	SD_BUS_PROPERTY("Profiles", "ao", ratbagd_device_get_profiles, 0, SD_BUS_VTABLE_PROPERTY_CONST),
 	SD_BUS_METHOD("Commit", "", "u", ratbagd_device_commit, SD_BUS_VTABLE_UNPRIVILEGED),
 	SD_BUS_SIGNAL("Resync", "", 0),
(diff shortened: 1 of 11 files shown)


src/driver-steelseries.c
```diff
@@ -421,14 +423,15 @@ steelseries_probe(struct ratbag_device *device)
...
-		log_debug(device->ratbag, "SteelSeries firmware version %d.%d\n",
-			drv_data->firmware_major,
-			drv_data->firmware_minor);
+	{
+		int firmware_minor = 0;
+		int firmware_major = 0;
+		rc = steelseries_get_firmware_version(device, &firmware_minor, &firmware_major);
+		if (rc == 0) {
+			device->firmware_version = zalloc(STEELSERIES_FIRMWARE_VERSION_STRING_LENGTH + 1);
+			snprintf(device->firmware_version, STEELSERIES_FIRMWARE_VERSION_STRING_LENGTH + 1, "%d.%d", firmware_major, firmware_minor);
```
whot, line 432: this could be a simpler:
```c
device->firmware_version = asprintf_safe("%d.%d", major, minor);
```

(`asprintf_safe` should probably  be changed to return `char*` but that's separate)

staticssleever668: In such cases would it be preferable to set `device->firmware_version` directly to not do a redundant string copy, or should the `ratbag_device_set_fw_version` API be used to stay on higher abstraction layer?

src/driver-sinowealth.c
```diff
@@ -1417,11 +1417,11 @@ sinowealth_init_profile(struct ratbag_device *device)
 	 */
 	struct sinowealth_config_report *config = &drv_data->configs[0];
 
-	char fw_version[SINOWEALTH_FW_VERSION_LEN + 1] = { 0 };
+	device->firmware_version = zalloc(SINOWEALTH_FW_VERSION_LEN + 1);
```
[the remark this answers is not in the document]
staticssleever668: Thank you!

ratbagd/ratbagd-device.c
```diff
@@ -220,10 +220,27 @@ ratbag_device_get_model(sd_bus *bus,
...
 }
 
+static int
+ratbagd_device_get_firmware_version(sd_bus *bus,
+			   const char *path,
+			   const char *interface,
+			   const char *property,
+			   sd_bus_message *reply,
+			   void *userdata,
+			   sd_bus_error *error)
```
staticssleever668, line 230: Incorrect indentation.

src/driver-steelseries.c
whot: the cost of doing this is miniscule - you're doing this exactly once during device startup and a string copy isn't expensive in the grand scheme of things. So something like this is best, IMO:
```c
_cleanup_free_ char *fw = asprintf_safe("%d.%d", major, minor);
ratbag_device_set_fw_version(device, fw);
```
(or whatever the right cleanup thing is)

alternatively and I only realised this now: how about making `ratbag_device_set_fw_version()` a printf-like function? Then most drivers can just do `ratbag_device_set_fw_version(device, "%d.%d", major, minor)` and be done with it.

staticssleever668: > So something like this is best, IMO

That's what I will do, then. :)

Still, I'm not a fan of doing any wasteful operations, but here it really isn't heavy at all and having the code more maintainable is more important.

> how about making `ratbag_device_set_fw_version()` a printf-like function

I don't really like this, as this would make the function do several things at the same time. Having a separate formatting and data setting is good as is in my opinion.

src/driver-steelseries.c
```diff
@@ -199,9 +194,8 @@ button_defaults_for_layout(struct ratbag_button *button, int button_count)
 }
 
 static int
-steelseries_get_firmware_version(struct ratbag_device *device)
+steelseries_get_firmware_version(struct ratbag_device *device, int *out_major, int *out_minor)
```
staticssleever668, line 197: For consistency with the codebase: out_major -> major_out ; out_minor -> minor_out

whot (approved):

whot:
LGTM now, thanks!

PR #1377, opened 2022-11-24
ASUS: reworked ButtonMapping config property and added Spatha X mouse

kyokenn:
Some mice uses button order which differs from the default one. So storing button indices won't work for them. I can't rely on the hardcoded buttons table. I have changed ButtonMapping to store the actual button codes in the order that they appear in the packet. So I can get the current bindings and rebind any button.
I also added extra buttons for Spatha X mouse and it's device file. We did some research and tests with device owner in this issue - https://github.com/kyokenn/rogdrv/issues/50.

### DIFF
--- a/src/driver-asus.c
+++ b/src/driver-asus.c
@@ -55,6 +55,7 @@ static int8_t ASUS_CONFIG_BUTTON_MAPPING[] = {
 	-1,  /* placeholder */
 	-1,  /* placeholder */
 	-1,  /* placeholder */
+	-1,  /* placeholder */
 };
 
 static unsigned int ASUS_LED_MODE[] = {
--- a/src/driver-asus.c
+++ b/src/driver-asus.c
@@ -207,36 +207,35 @@ asus_driver_save_profile(struct ratbag_device *device, struct ratbag_profile *pr
 			continue;
 		}
 
-		int button_asus_code = drv_data->button_mapping[asus_index];
+		int asus_code_src = drv_data->button_mapping[asus_index];
+		int asus_code_dst;
+		struct asus_button *asus_button;
 		log_debug(device->ratbag, "Button %d (%02x) changed\n",
-			  button->index, (uint8_t) button_asus_code);
+			  button->index, (uint8_t) asus_code_src);
 
 		switch (button->action.type) {
 		case RATBAG_BUTTON_ACTION_TYPE_NONE:
 			rc = asus_set_button_action(
-				device, button_asus_code,
-				ASUS_BUTTON_CODE_DISABLED,
+				device, asus_code_src, ASUS_BUTTON_CODE_DISABLED,
 				ASUS_BUTTON_ACTION_TYPE_BUTTON);
 			break;
 
 		case RATBAG_BUTTON_ACTION_TYPE_KEY:
 			/* Linux code to ASUS code */
-			int asus_code = asus_find_key_code(button->action.action.key.key);
-			if (asus_code >= 0)
+			asus_code_dst = asus_find_key_code(button->action.action.key.key);
+			if (asus_code_dst >= 0)
 				rc = asus_set_button_action(
-					device, button_asus_code,
-					asus_code,
+					device, asus_code_src, asus_code_dst,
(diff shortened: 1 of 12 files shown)


src/driver-asus.c
```diff
@@ -189,51 +192,51 @@ static int
...
-		int asus_index = bmi[button->index] != -1 ? bmi[button->index] : ASUS_BUTTON_MAPPING_INDEX[button->index];
-		int asus_code;
+		int asus_index = drv_data->button_indices[button->index];
 		if (asus_index == -1) {
 			log_debug(device->ratbag, "No mapping for button %d\n", button->index);
 			continue;
 		}
 
-		log_debug(device->ratbag, "Button %d (%d) changed\n", button->index, asus_index);
+		int asus_code_src = drv_data->button_mapping[asus_index];
```
staticssleever668, line 211: This is missing handling of the sentinel `-1` value.
Also, it would be nice to make this a u8 from start, so perhaps something like this would be good:
```suggestion
		uint8_t asus_code_src = 0;
		{
			rc = drv_data->button_mapping[asus_index];
			if (rc == -1) {
				log_debug(device->ratbag, "No mapping for button %d\n", button->index);
				continue;
			}
			asus_code_src = (uint8_t)rc;
		}

```

src/driver-asus.c
```diff
@@ -189,51 +192,51 @@ static int
...
-				ASUS_BUTTON_CODE_DISABLED,
+				device, asus_code_src, ASUS_BUTTON_CODE_DISABLED,
 				ASUS_BUTTON_ACTION_TYPE_BUTTON);
 			break;
 
 		case RATBAG_BUTTON_ACTION_TYPE_KEY:
 			/* Linux code to ASUS code */
-			asus_code = asus_find_key_code(button->action.action.key.key);
-			if (asus_code >= 0)
+			asus_code_dst = asus_find_key_code(button->action.action.key.key);
```
staticssleever668, line 226: Same thing as with `asus_code_src` - the sentinel `-1` is not handled.
Also, because of this a few lines below there is an implicit conversion from int to u8.

src/driver-asus.c
```diff
@@ -37,16 +37,17 @@
...
+/* ButtonMapping configuration property defaults */
+static int8_t ASUS_CONFIG_BUTTON_MAPPING[] = {
+	0xf0,  /* left */
+	0xf1,  /* right (button 3 in xev) */
+	0xf2,  /* middle (button 2 in xev) */
+	0xe4,  /* backward */
+	0xe5,  /* forward */
+	0xe6,  /* DPI */
+	0xe8,  /* wheel up */
+	0xe9,  /* wheel down */
```
staticssleever668, line 49: This problem was here even before this PR, but these values do not fit in 8 bytes, see compiler warning:

```
../src/driver-asus.c:42:9: warning: conversion from ‘int’ to ‘signed char’ changes value from ‘240’ to ‘-16’ [-Wconversion]
   42 |         0xf0,  /* left */
      |
```

src/driver-asus.c
```diff
@@ -423,6 +436,31 @@ asus_driver_probe(struct ratbag_device *device)
...
+	/* setup a lookup table for all defined buttons */
+	unsigned int button_index = 0;
+	for (unsigned int i = 0; i < ARRAY_LENGTH(ASUS_BUTTON_MAPPING); i++) {
+		struct asus_button *asus_button = &ASUS_BUTTON_MAPPING[i];
+
+		/* search for this button in the ButtonMapping by it's ASUS code */
+		for (unsigned int j = 0; j < button_count; j++) {
+			if (drv_data->button_mapping[j] == (int8_t) asus_button->asus_code) {
+				/* add button to indices array */
+				drv_data->button_indices[button_index] = j;
```
staticssleever668, line 456: With `-Wconversion` this gives an implicit conversion warning.
```suggestion
				drv_data->button_indices[button_index] = (int)j;
```

staticssleever668 (approved):

staticssleever668:
Nice!

PR #1380, opened 2022-12-08
asus: allow empty cells in bindings table and fix pugio button bindings

kyokenn:
I forgot that some mice have zero byte elements in the bindings array. So those elements should be skipped while searching for the button code. It also fixes button bindings info for the Pugio mouse.

### DIFF
--- a/data/devices/asus-rog-pugio.device
+++ b/data/devices/asus-rog-pugio.device
@@ -9,4 +9,4 @@ Buttons=10
 Leds=3
 Dpis=2
 DpiRange=100:7200@100
-ButtonMapping=f0;f1;f2;e4;e5;e6;;e8;e9;e1;e2
+ButtonMapping=f0;f1;f2;e4;e5;e6;0;e8;e9;e1;e2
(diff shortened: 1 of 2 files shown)


data/devices/asus-rog-pugio.device
```diff
@@ -9,4 +9,4 @@ Buttons=10
 Leds=3
 Dpis=2
 DpiRange=100:7200@100
-ButtonMapping=f0;f1;f2;e4;e5;e6;e8;e9;e1;e2
+ButtonMapping=f0;f1;f2;e4;e5;e6;;e8;e9;e1;e2
```
staticssleever668, line 12: Maybe this should be a `0` explicitly?

Because otherwise we rely on `strtoul` in libratbag-data.c:376 converting an empty string to `0`, which is not completely obvious.
Example:
```c
#include <stdio.h>
#include <stdlib.h>

int main() {
  printf("%lu", strtoul("", NULL, 16));
  // > `0` printed.
}
```

kyokenn: Yes. I'm fine with those changes.

staticssleever668 (approved):

PR #1389, opened 2022-12-20
New additional usb id for Logitech G102

mistepien:
New additional usb id for Logitech G102: usb:046d:c09d


staticssleever668:
Hi, thank you for you contribution!
Just for future, could you post the output of `lsusb -vd 046d:c09d`?

Issue #1168, opened 2021-04-29
API for changing debounce time, lift off distance and angle snapping


staticssleever668:
Some mice, for example Sinowealth-based, support changing debounce time, lift off distance and angle snapping in their configuration software. It would be nice if this was possible to do in libratbag.

staticssleever668:
Angle snapping and debounce time were implemented.
For lift-off distance I don't have any use-case and won't work on it myself, but if anyone sends a PR, I'd be happy to review it.

PR #1394, opened 2022-12-22
add angle snapping and debounce time

kyokenn:
Adds API for https://github.com/libratbag/libratbag/issues/1168 and implementation for ASUS mice.

I also made UI with new options for Piper.

### DIFF
--- a/doc/dbus.rst
+++ b/doc/dbus.rst
@@ -264,6 +264,37 @@ org.freedesktop.ratbag1.Profile
         Provides the object paths of all LEDs in this profile, see
         :ref:`led`.
 
+.. attribute:: AngleSnapping
+
+        :type: i
+        :flags: read-write, mutable
+
+        Sensor angle snapping boolean value as an int (1 or 0), or `-1` to
+        indicate that the device doesn't support reading and/or writing this
+        value.
+
+.. attribute:: Debounce
+
+        :type: i
+        :flags: read-write, mutable
+
+        Int for the button debounce time in milliseconds assigned to this
+        profile. This time must be one of those listed in
+        :attr:`Debounces`, or `-1` to indicate that changing debounce time
+        for this device is not allowed.
+
+.. attribute:: Debounces
+
+        :type: au
+        :flags: read-write, constant
+
+        A list of permitted button debounce times. Values in this list may
+        be used in the :attr:`Debounce` property. This list is always sorted
+        ascending, the lowest debounce time is the first item in the list.
+
+        This list may be empty if the device does not support reading and/or
+        writing the debounce time.
+
 .. attribute:: ReportRate
 
         :type: u
--- a/ratbagd/ratbagd-profile.c
+++ b/ratbagd/ratbagd-profile.c
@@ -455,6 +455,40 @@ ratbagd_profile_get_report_rate(sd_bus *bus,
 }
 
 static int
+ratbagd_profile_get_angle_snapping(sd_bus *bus,
+				   const char *path,
+				   const char *interface,
+				   const char *property,
+				   sd_bus_message *reply,
+				   void *userdata,
+				   sd_bus_error *error)
+{
+	struct ratbagd_profile *profile = userdata;
+	struct ratbag_profile *lib_profile = profile->lib_profile;
+	int value;
+
+	value = ratbag_profile_get_angle_snapping(lib_profile);
+	return sd_bus_message_append(reply, "i", value);
+}
+
+static int
+ratbagd_profile_get_debounce(sd_bus *bus,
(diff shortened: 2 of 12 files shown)


staticssleever668:
This is useful, but it will have to wait until we have a way to tell if a device supports these actions.

This would require us adding a new API for device capabilities, which I'm thinking of working on.
We also already have such API for profiles and resolutions/DPIs.

Related: https://github.com/libratbag/libratbag/pull/931#discussion_r449992893

staticssleever668:
> This would require us adding a new API for device capabilities

I was looking at this again and noticed that angle snapping and debounce time here are implemented per-profile. I assumed it was per-device because Sinowealth devices don't seem to support setting this for each profile, only one setting for the whole device.

Either way, this needs to add some capabilities to `enum ratbag_profile_capability`.

---

Well then, we have to think what to do with devices that are like Sinowealth.

Two ways I can think of right now:

1) Add another similar API but for the whole device.
This would be both complex and confusing, but may be worth it in the end.

2) Go with per-profile API, and in drivers for devices like Sinowealth update every single profile when only one is altered.
This may have a problem with D-Bus clients and ratbagd getting out of sync, I don't think we have a nice way to force a resync from drivers right now (although we can do it by triggering an error).

kyokenn:
>Go with per-profile API, and in drivers for devices like Sinowealth update every single profile when only one is altered.

I like this way. 

>I don't think we have a nice way to force a resync from drivers right now (although we can do it by triggering an error).

Maybe this could be solved in future with a new API.

>Add another similar API but for the whole device. This would be both complex and confusing, but may be worth it in the end.

This one could be too complicated and out of ratbag design, where all the settings are bound to the profile.

staticssleever668:
@whot, sorry to bother you, but do you think such API fits libratbag? If I remember right, in some issue related to ratbag-python you mentioned that such APIs are highly device-specific, and you would rather have a way for drivers to have their own additional API, than making something like this generic. Although I don't think there is a lot of such features, so having proper support for all of them won't be bad. Not to mention there isn't yet a way for drivers to add their own features.
In defense for this change, debounce time adjustment is pretty popular in Minecraft community to easen the so called "drag clicking" - a way of clicking mouse buttons which allows for very fast inputs, like 30 clicks per second. I can't defend angle snapping, but I don't any problems in supporting it.

staticssleever668:
> 2\. Go with per-profile API, and in drivers for devices like Sinowealth update every single profile when only one is altered.

Thinking about it again, we could just enable these options for the first profile only.
This would be a bit less convenient to the user, but it would also not be deceiving, so that they won't think that this is per-profile and be confused that the value remains the same when they switch profiles.

staticssleever668:
> Either way, this needs to add some capabilities to `enum ratbag_profile_capability`.

Having sentinel values like you did is good too :+1:

staticssleever668:
Implemented debounce time getting/setting for SinoWealth

TODO:
- Check if indentation is correct (like 66a3c6ade75b4e9730e7026feef88bffea74bd96)
- Add error handling to ratbagctl (right now it will happily ignore any errors from the driver)
- Update D-Bus docs

whot:
> @whot, sorry to bother you, but do you think such API fits libratbag? 

First, a few things to keep in mind: libratbag was started in 2015, nearly 8 years ago. And it was written in C to provide better compatibility with everything, with the goal of it being a library used by multiple actual implementations. That didn't really stand the test of time since we now just export a DBus daemon and some cli tools.

C is a horrible language for many things which is why the ratbag API is very restrictive and mostly focused on the features available in multiple vendors' devices. Also, both bentiss and me are not gaming mice users as such, we had fun making libratbag but I don't actually *need* any of the features it provides. Which is also why our guesses on what is needed were sometimes wrong. These mice tend to have marketing features (looks good on the box, no-one really needs/uses them though) and useful features. Distinguishing between the two is hard when you're not using the device that way yourself.

But really, in many cases C is the single reason things are terrible - e.g. returning a list of values in C is painful.

Hence my attempt (except `ETIME`) to get ratbag-python off the ground to be able to make a bunch of things a lot easier. This would also allow for much more flexible APIs - we can push a lot of the feature set handling to the callers which are presumably written in more sensible languages and thus able to filter better.


Anyway. Long story short: if it's a useful feature and it's available on a bunch of mice, add it to the API. That's pretty much the summary. The answer to "does this fit the API" is often "no. no. no. not yet. ok, now it is" spread over a number of years. :wink:

staticssleever668:
Rebased, fixed some indentation issues. For sinowealth made debounce time reading/writing optional as it looks like some devices may not support it.

staticssleever668:
Great, thank you!

PR #1414, opened 2023-01-27
driver-hidpp10: fix integer overflow when profile count is unset

staticssleever668:
This showed up while working on #1244.

### DIFF
--- a/src/driver-hidpp10.c
+++ b/src/driver-hidpp10.c
@@ -656,7 +656,15 @@ hidpp10drv_probe(struct ratbag_device *device)
 		else if (strcasecmp("G9", typestr) == 0)
 			type = HIDPP10_PROFILE_G9;
 
-		profile_count = ratbag_device_data_hidpp10_get_profile_count(device->data);
+		rc = ratbag_device_data_hidpp10_get_profile_count(device->data);
+		if (rc == -1) {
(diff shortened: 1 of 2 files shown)


src/driver-hidpp10.c
```diff
@@ -656,7 +656,15 @@ hidpp10drv_probe(struct ratbag_device *device)
 		else if (strcasecmp("G9", typestr) == 0)
 			type = HIDPP10_PROFILE_G9;
 
-		profile_count = ratbag_device_data_hidpp10_get_profile_count(device->data);
+		rc = ratbag_device_data_hidpp10_get_profile_count(device->data);
+		if (rc == -1) {
+			log_error(device->ratbag,
+				  "Device %s has no profile count set, even though profiles are enabled. "
+				  "Please adjust the .device file.\n",
+				  device->name);
```
amfern, line 664: Should it fall back to `profile_count = 1` here?

staticssleever668: It's initialized with `1` above
https://github.com/libratbag/libratbag/blob/931ec44ff27fd2ef78b830bbe20acd2e26576a9f/src/driver-hidpp10.c#L639

amfern (approved):

PR #1424, opened 2023-02-15
Create genesis-xenon210.device

nyxnyx:
This is gaming mouse with 8 buttons and rgb led


staticssleever668:
Hello, I assume you are running a slightly older libratbag version, as the device file format for the `sinowealth` driver has changed in v0.17.
Could you post the output of `ratbagd --verbose=raw` so that I can update this file for you?

PR #1430, opened 2023-02-25
Remove deprecated reatbag_led_type and ratbag_button_type

staticssleever668:


staticssleever668:
There is one actual logic change is in 5548da21d335d525955b7b7337fdb836685d89fc, this is something that would be nice to test before making a new release.

Issue #992, opened 2020-05-24
Logitech M720 issues -- macros broken and bluetooth ID


acegallagher:
- `ratbagd` version (`ratbagd --version`):
0.14 (git)
- Distribution: 
Arch 
- Kernel version (ex. `uname -srmo`): 
Linux-CK_ 5.6.13-1 (but also tested on arch and lts)

Device info (if applicable):
```
$ ratbagctl <device> info
cheering-viscacha - M720 Triathlon
             Model: bluetooth:046d:b015:0
 Number of Buttons: 9
    Number of Leds: 0
Number of Profiles: 1
Profile 0: (active)
  Name: n/a
  Button: 0 is mapped to 'button 1'
  Button: 1 is mapped to 'button 2'
  Button: 2 is mapped to 'button 3'
  Button: 3 is mapped to 'button 4'
  Button: 4 is mapped to 'button 5'
  Button: 5 is mapped to 'wheel-left'
  Button: 6 is mapped to 'wheel-right'
  Button: 7 is mapped to 'button 4'
  Button: 8 is mapped to UNKNOWN
```

**Describe the bug**

**To Reproduce**

**Logs**
Please start the daemon with verbosity and reproduce the issue.

First make sure it isn't running already.
```
$ ratbagd --verbose=raw
Initializing libratbag
ratbag debug: New device: Logitech M720 Triathlon
ratbag debug: Using data directory '/usr/share/libratbag'
ratbag debug: device assigned driver hidpp20
ratbag debug: hidraw info: bus 0x03 vendor 0x46d product 0x405e
ratbag debug: Logitech M720 Triathlon is device '/dev/hidraw2'.
ratbag debug: Parsing HID report descriptor
ratbag debug: Parsing HID report descriptor
ratbag debug: - HID report ID 01
ratbag debug: - HID report ID 0e
ratbag debug: - HID report ID 02
ratbag debug: - HID report ID 03
ratbag debug: - HID report ID 04
ratbag debug: - HID report ID 10
ratbag debug: - HID report ID 11
ratbag debug: - HID report ID 20
ratbag debug: - HID report ID 21
ratbag debug: hidpp: device supports short reports
ratbag debug: hidpp: device supports long reports
ratbag raw: hidpp write:  10 ff 00 18 00 00 00
ratbag raw: hidpp read:   10 01 8f 00 18 09 00
ratbag debug:     HID++ error from the device (1): ERR_UNSUPPORTED (09)
[...]

PR #1433, opened 2023-03-01
Fix ratbagctl crash if there is no active profile or no active resolution

staticssleever668:
This fixes cryptic libratbag crashes seen in #992.


staticssleever668:
This could probably be structured better, but I can't do better and there is no one to review this right now, while I want to get this in before doing some more destructive changes.

PR #1446, opened 2023-03-05
Add new D-Bus API to check if a profile is dirty

staticssleever668:


staticssleever668:
This doesn't add use of new ratbag_profile_is_dirty API to ratbagc.py, but it never sets this property to anything other than False, and it's not used anywhere either. Probably something to address later: update the `dirty` property in RatbagdProfile in ratbagc.py, and show whether a profile is dirty in ratbagctl.

Issue #1413, opened 2023-01-24
Asus ROG Chakram X support doesn't work


ProspectPyxis:
**Information**
- `ratbagd` version (`ratbagd --version`): 0.17 (Arch Linux libratbag-git 0.17.r85.6dcd17d-1)
- Distribution: Arch Linux
- Kernel version (ex. `uname -srmo`): `Linux 6.1.7-arch1-1 x86_64 GNU/Linux`

**Describe the bug**
Ratbag doesn't seem to work with the Asus ROG Chakram X correctly, for multiple reasons.

First, the pid:vid seems to be incorrect - while the built-in device file has it listed as `0b05:1a1a`, the mouse shows up as `0b05:1a18` on my device, and thus, it isn't detected.

Second, even after manually changing the vid, the mouse will fail to work, with multiple errors of `'Connection timed out' (-110)` and occasionally `'Protocol Error' (-71)` when checking `systemctl status ratbagd.service`.

**To Reproduce**
1. Plug in the mouse via USB.
2. Mouse is not detected by ratbagd.
3. Manually change the device pid:vid in the corresponding device file from `0b05:1a1a` to `0b05:1a18`.
4. Restart `ratbagd.service`.
5. The mouse shuts down - the RGB lights beneath the click buttons are frozen green, all other RGB lights turn off, and no input works.

**Logs**
Note that this is the output from after editing the pid:vid.

```
$ ratbagd --verbose=raw
Initializing libratbag
ratbag debug: New device: LIZHI Flash IC USB Keyboard
ratbag debug: Using data directory '/usr/share/libratbag'
ratbag debug: No data file found for 1c4f:0084
ratbag debug: New device: LIZHI Flash IC USB Keyboard
ratbag debug: Using data directory '/usr/share/libratbag'
ratbag debug: No data file found for 1c4f:0084
ratbag debug: New device: ASUSTeK ROG CHAKRAM X
ratbag debug: Using data directory '/usr/share/libratbag'
ratbag debug: device assigned driver asus
ratbag debug: hidraw info: bus 0x03 vendor 0xb05 product 0x1a18
ratbag debug: ASUSTeK ROG CHAKRAM X is device '/dev/hidraw2'.
ratbag debug: Parsing HID report descriptor
ratbag debug: Parsing HID report descriptor
ratbag debug: Button 0 is mapped to 0xf0
ratbag debug: Button 1 is mapped to 0xf1
ratbag debug: Button 2 is mapped to 0xf2
[...]

PR #1468, opened 2023-03-30
asus: check for hid report and fix chakram x

kyokenn:
Fixes https://github.com/libratbag/libratbag/issues/1413

Every ASUS mouse have multiple interfaces for mouse, keyboard, configuration and other stuff. There was some error messages because libratbag tried to initialize on the wrong interfaces of the mouse. It tries every interface and fails on the wrong one with error message. This PR fixes it and I have tested this part.

### DIFF
--- a/src/driver-asus.c
+++ b/src/driver-asus.c
@@ -434,6 +434,7 @@ asus_driver_probe(struct ratbag_device *device)
 	unsigned int profile_count, dpi_count, button_count, led_count;
 	const struct asus_button *asus_button;
 	struct asus_data *drv_data;
+	struct asus_profile_data profile_data;
 	struct ratbag_profile *profile;
 	struct ratbag_button *button;
 	struct ratbag_resolution *resolution;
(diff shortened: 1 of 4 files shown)


src/driver-asus.c
```diff
@@ -443,6 +443,11 @@ asus_driver_probe(struct ratbag_device *device)
 	if (rc)
 		return rc;
 
+	/* HID device descriptor must have input and output */
+	if (!ratbag_hidraw_has_report(device, HID_INPUT_REPORT) ||
+	    !ratbag_hidraw_has_report(device, HID_OUTPUT_REPORT))
+		return -1;
```
staticssleever668, line 449: Could you change this to `-ENODEV`?
The `probe` function is supposed to only return POSIX error codes on error, moreover, if it's something other than ENODEV it's going to print it to stderr.

staticssleever668 (changes requested):
One little change and it's good (you could also split this into two commits but whatever). Thank you so much for your work on the asus driver! :)

kyokenn:
"ratbag_hidraw_has_report" doesn't work for me anymore for some reason. Maybe I have updated some packages. So I replaced it with profile request, it should work even when the mouse is sleeping or turned off.

staticssleever668 (approved):

staticssleever668:
> "ratbag_hidraw_has_report" doesn't work for me anymore for some reason.
Well, if you find a way to use it instead, feel free to open another PR to avoid duplicate work.

PR #1471, opened 2023-04-08
Feat simple button action modifiers

### LINKED ISSUE #1470: Add modifiers to single key bindings
We need a way to add to modifiers to single key bindings (`RATBAG_BUTTON_ACTION_TYPE_KEY`, not `RATBAG_BUTTON_ACTION_TYPE_MACRO`).
There is a few FIXMEs in the code for this:
https://github.com/libratbag/libratbag/blob/70f2e96a66ab4b5822ebb17c7eabeb8857b20446/src/libratbag-private.h#L351
https://github.com/libratbag/libratbag/blob/70f2e96a66ab4b5822ebb17c7eabeb8857b20446/src/libratbag.c#L1389
https://github.com/libratbag/libratbag/blob/70f2e96a66ab4b5822ebb17c7eabeb8857b20446/src/libratbag.c#L1409
These functions take in/out modifiers like this:
Pointer to an array of modifiers, and pointer to a variable with size of the array. Currently it ignores whatever is in the array, and just resets the variable with size to zero.

However, some other functions, like ratbag_button_macro_new_from_keycode(), use a mask for modifiers.
https://github.com/libratbag/libratbag/blob/70f2e96a66ab4b5822ebb17c7eabeb8857b20446/src/libratbag.c#L1907-L1910

Here's what I plan on doing:
1) Add array of modifiers and it's size to ratbag_button_action.action.key
1) Use it in ratbag_button_set_key and ratbag_button_get_key.
It should check that there is only key (e.g. `KEY_A KEY_B` not allowed), and that there are no duplicate modifiers.
1) Add tests for it.
1) Not well thought off: modify the interface in ratbagctl to use it like this:
`ratbagctl Glorious button 5 action set key KEY_LEFTSHIFT KEY_LEFTMETA KEY_A`
1) Check if there was a proper reason for this, if not, change ratbag_button_macro_new_from_keycode() to use an array instead of a mask. There should be only one API for this (maintenance burden), and I think an array is more straightforward.

Continuation of #1251, I think the only blocker for #1388.

staticssleever668:
Closes #1470.
ratbagctl is not implemented and the API itself is not tested yet.


staticssleever668:
I wonder what the D-Bus API for this should look like.
Here is what I have in mind right now:
1) Add additional field of data - an array - to the existing methods ratbagd_button_get_key() and ratbagd_button_set_key().
2) If above is not possible, make the passed data an array instead of an int, where the first index is the key and the rest is modifiers.

staticssleever668:
I think we also need to separately indicate that modifiers are supported, so that Piper can safely fall-back to using macros when needed.

staticssleever668:
I decided not to add modifiers to simple button events. If such functionality is needed, macros can be used instead.
Example of transforming simple macro into a key action on driver side: 33f8791dfd045a33.
I will make a new PR with enhancements from here.

staticssleever668:
Superseded by #1523.

PR #1520, opened 2023-09-20
Create logitech-g102.device

himark1977:
Add support for Logitech G102 c09d


staticssleever668:
Hello! Thank you, but it's already in logitech-g102-g203.device and does the same thing as #1389.
This will be available OOTB in the next release.

PR #1538, opened 2023-10-05
Update logitech-g513.device

fritteli:
The G513 is actually a keyboard, not a mouse (at least mine is ...).


staticssleever668:
Thank you, good catch! No need to change it's "Name" though, can you only leave the "DeviceType" change, please?

staticssleever668 (approved):

staticssleever668:
Thank you!

PR #1546, opened 2023-10-27
This device operates exactly like the MX 3S

bithead2k:
There is no technical difference between this mouse and the Logitech MX Master 3S.   However, it still gets its own device id and unique name for no apparent reason.

### DIFF
--- a/data/devices/logitech-MX-Master-3S-for-Business.device
+++ b/data/devices/logitech-MX-Master-3S-for-Business.device
@@ -1,0 +1,5 @@
+[Device]
+DeviceMatch=bluetooth:046d:b035
+DeviceType=mouse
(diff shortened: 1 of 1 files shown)


staticssleever668 (approved):
Thank you! Could you also open a PR for this device on Piper side?
Add this to data/svgs/svg-lookup.ini.
```dosini
[Logitech MX Master 3S For Business]
DeviceMatch=bluetooth:046d:b035
Svg=logitech-mx-master-3s.svg
```

staticssleever668:
Rebased and changed 'For' -> 'for' (https://www.logitech.com/en-us/products/mice/mx-master-3s-business-wireless-mouse.910-006581.html).
Thank you!

PR #1556, opened 2023-11-19
Create logitech-g304.device

BaconCondensate:
Add support to Logitech G304 (Asian market version of G305)

### DIFF
--- a/data/devices/logitech-g304.device
+++ b/data/devices/logitech-g304.device
@@ -1,0 +1,9 @@
+[Device]
+Name=Logitech Gaming Mouse G304
+DeviceMatch=usb:046d:c53f
+Driver=hidpp20
+DeviceType=mouse
+
+[Driver/hidpp20]
+Leds=0
+Quirk=G304
--- a/data/devices/logitech-g304.device
+++ b/data/devices/logitech-g304.device
@@ -6,4 +6,4 @@ DeviceType=mouse
 
 [Driver/hidpp20]
 Leds=0
-Quirk=G304
+Quirk=G305


data/devices/logitech-g304.device
```diff
@@ -0,0 +1,9 @@
+[Device]
+Name=Logitech Gaming Mouse G304
+DeviceMatch=usb:046d:c53f
```
staticssleever668, line 3: 046d:c53f is receiver's ID, which doesn't usually work with libratbag. Not sure why tests don't fail for that. 
Perhaps `lsusb` shows another Logitech device?
If not, try connecting it in wired mode.

data/devices/logitech-g304.device
```diff
@@ -0,0 +1,9 @@
...
+Driver=hidpp20
+DeviceType=mouse
+
+[Driver/hidpp20]
+Leds=0
+Quirk=G304
```
staticssleever668, line 9: 'G304' is not a valid quirk, it should could either be 'G305' or not set at all. Our tests don't catch that right now, which I opened an issue for: #1557.

staticssleever668 (changes requested):
Hello! Thank you for your PR, but it needs some changes

data/devices/logitech-g304.device
BaconCondensate: Yeah, I'll need to change it to G305 then. It's my first time trying to do this so pardon my skill issue

BaconCondensate:
After a lot of testing, I have concluded that G304 have the exact same PID as G305, and the reason it was not detected is not due to the lack of device file, but due to a bug on my end. Closing MR

PR #1560, opened 2023-11-23
MX Master 2S: Add Logitech Unifying Receiver

ZephyraSilentis:
added the USB ID for Logitech Unifying Receivers.


staticssleever668:
Hi! libratbag internationally doesn't have device files for receivers as it usually doesn't work, especially if there are several devices connected to it.
We expect the kernel to reexport devices separately (which needs a recent enough kernel for some newer devices).

PR #1561, opened 2023-11-24
Driver for mice with Holtek 8-bit MCUs

lubasowo0:
Hi! This pull request introduces a new driver for mice with Holtek 8-bit MCUs (such as HT68FB550 and HT68FB560) with OEM firmware. Unfortunately the ABI between devices is not stable but with supporting just 2 API types we are able to support a good fraction of devices. Devices which use these MCUs but with their own proprietary firmware such as Redragon are obviously not supported. I only have one device of each type so the rest is obviously untested and their device file are not yet added but I could verify with an usb emulator and official softwares of many devices that a lot of the functionality is common between them.

This includes reading and writing of profiles, resolutions, rates, buttons and macros and all of this is implemented in this driver. This doesn't include LEDs (illumination modes are different but with a bit of work it may be possible to represent in device file), debounce times (ranges are different but probably possible to have in device file) and liftoff distance (not planned, ABI are completly unstable and few devices support it).

Device file loading is based on sinowealth's because especially devices of api ver.A share a common vid:pid pair and devices are recognised based on their firmware version, which is conveniently stored in bcdDevice field of usb device descriptor.

Configuartion interface doesn't use numbered reports at all, instead exchanging data via constant length chunks, which are interpreted based on previously sent SET feature report with a command and parameters. This can lead to unsynchronized reads/writes which are known to soft brick devices even using official software, so we actually do this part better, checking if the device really is expecting data.

I decided to have both apis as seperate drivers so I could handle their behavior more clearly which led to some code duplication but I tried to handle as much shared behavior in a shared file as possible.


staticssleever668:
Hi! Thank you for your PR. I didn't look at the changes yet, but I noticed you didn't update the device file test. Could you update it too?
It's at 'test/data-parse-test.py'.
Add an 'elseif' here, you could look at the check for sinowealth above as an example, but call out to a function, which would verify keys *and* values (sinowealth check is lacking in this regard).
https://github.com/libratbag/libratbag/blob/6db5cc374352367937bf00098ba365e50d645ee5/test/data-parse-test.py#L321

lubasowo0:
I rewrote the test a bit for sinowealth and holtek so they can have their own check_section functions

src/driver-holtek.h
```diff
@@ -0,0 +1,29 @@
...
+	HOTLEK8_SENSOR_PAW3333,
+	HOTLEK8_SENSOR_PMW3320,
+};
+
+struct holtek8_device_data {
+	char *fw_version;
+	char *device_name;
+	enum holtek8_sensor sensor;
+	int button_count;
+	uint8_t password[6];
```
staticssleever668, line 23: Make it [7] and you can get nice null termination ;)

src/driver-holtek.h
```diff
@@ -0,0 +1,29 @@
...
+#define HOLTEK8_FW_VERSION_LEN 4
+
+enum holtek8_sensor {
+	HOTLEK8_SENSOR_UNKNOWN = 0,
+	HOTLEK8_SENSOR_PAW3333,
+	HOTLEK8_SENSOR_PMW3320,
+};
+
+struct holtek8_device_data {
+	char *fw_version;
```
staticssleever668, line 19: (this was probably based on the sinowealth driever but) you can probably make this an array too

src/driver-holtek.h
```diff
@@ -0,0 +1,29 @@
...
+#pragma once
+
+#include "libratbag-util.h"
+
+#define HOLTEK8_FW_VERSION_LEN 4
+
+enum holtek8_sensor {
+	HOTLEK8_SENSOR_UNKNOWN = 0,
+	HOTLEK8_SENSOR_PAW3333,
+	HOTLEK8_SENSOR_PMW3320,
```
staticssleever668, line 15: "HOTLEK" vs "HOLTEK", typo?

src/driver-holtek/holtek8-shared.h
```diff
@@ -0,0 +1,280 @@
...
+#define HOLTEK8_FEATURE_REPORT_SIZE 9
+#define HOLTEK8_BUTTON_SIZE 4
+#define HOLTEK8_MACRO_EVENT_SIZE 2
+#define HOLTEK8_MACRO_DATA_SIZE 128
+
+#define HOLTEK8_MAX_MACRO_EVENTS 256
+
+enum holtek8_api_version {
+	HOTLEK8_API_A = 1,
+	HOTLEK8_API_B = 2,
```
staticssleever668, line 20: "HOTLEK" vs "HOLTEK" too

staticssleever668, line 20: > Unfortunately the ABI between devices is not stable but with supporting just 2 API types we are able to support a good fraction of devices.

If 'A' and 'B' are the names you chose yourself, add a comment that it's not some official names and they are just for libratbag, please, to avoid confusion

src/driver-holtek/holtek8-shared.c
```diff
@@ -0,0 +1,1301 @@
...
+
+
+	return 0;
+}
+
+/*
+ * Converts ratbag button to raw device button data.
+ * If the ratbag's button action is macro, writes an encoded macro
+ * to the device memory and returns a data already pointing to the
+ * just written macro.
```
staticssleever668, line 753: Looks like it returns an int and writes to 'data' a copy of the written macro.
This should be like '@param[out] data Description' instead.

src/driver-holtek/holtek8-shared.c
```diff
@@ -0,0 +1,1301 @@
...
+				rc = holtek8_macro_to_events(button, macro_events);
+				//log_buf_error(device->ratbag, "MACRO: ", (uint8_t*) macro_events, 128);
+				rc = holtek8_write_macro_data(device, macro_events);
+				if (rc < 0)
+					return rc;
+
+				data->type = HOLTEK8_BUTTON_TYPE_MACRO;
+				data->macro.mode = HOLTEK8_BUTTON_MACRO_REPEAT_COUNT;
+				data->macro.index = rc;
+				data->macro._padding = 0;
```
staticssleever668, line 804: Consider memsetting 'data' in the beginning of the function instead.

staticssleever668 (commented):
Looked through about 60% of changes, looks good and clean so far, with just a few minor suggestions! Good job!
Also pushed a few commits:
- some const correctness; otherwise mutableness spreads like plague sometimes, making it hard to write new const-correct code
- Don't use 'default' for exhaustive switch cases; when adding new fields you'd be able to get compile warning for unhandled values, and right now those should be unreachable anyway, although enums in C are a landmine
- fix some implicit conversion; another landmine, I will get another look at it with -Wconversions later

src/driver-holtek/holtek8-shared.c
```diff
@@ -0,0 +1,1301 @@
...
+		case RATBAG_BUTTON_ACTION_TYPE_MACRO: {
+			unsigned int modifiers;
+			unsigned int key1;
+			unsigned int key2;
+
+			rc = holtek8_keycodes_from_ratbag_macro(&button->action, &key1, &key2, &modifiers);
+			if (rc == -EPROTO) {
+				struct holtek8_macro_event macro_events[HOLTEK8_MAX_MACRO_EVENTS] = {0};
+
+				rc = holtek8_macro_to_events(button, macro_events);
```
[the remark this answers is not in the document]
lubasowo0: Yes, this should definitely be checked. Fixed!

src/driver-holtek/holtek8-shared.c
```diff
@@ -0,0 +1,1301 @@
...
+				//log_buf_error(device->ratbag, "MACRO: ", (uint8_t*) macro_events, 128);
```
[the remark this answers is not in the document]
lubasowo0: Oh, this is just a leftover from development. --verbose=raw prints it much better on commit. Removed.

src/driver-holtek/holtek8-shared.c
lubasowo0: Reworded a bit and described param data

src/driver-holtek.h
lubasowo0: I somehow missed it 44 times... Fixed :D

src/driver-holtek/holtek8-shared.h
lubasowo0: Yes, these are completely arbitrary. Added a comment and a fun fact in the commit message.

src/driver-holtek.h
lubasowo0: done!

src/driver-holtek.h
lubasowo0: This one was kind of on purpose because at this point I started treating it as an array of 6 bytes instead of a string, but on a second thought these passwords are always ASCII so we may store a null terminator to make for example a debug printf much easier with %s. Let me know which way you prefer.

lubasowo0:
Thank you for the review so far! I admit I'm a bit terrible at catching things like implicit conversions so I may need to experiment with various warnings enabled too.

staticssleever668:
Sorry for taking long, I'll try taking a look at it again soon. I want to review it thoroughly and make a release of libratbag after merging it.

lubasowo0:
Sure! If you plan a release soon, you can take a look at my other PR too as that may break distro builds with meson test enabled and Python 3.12 (happened to me on Fedora already).

src/driver-holtek.h
staticssleever668: It's uint8_t[] so it's clear it's not a string, so it's good. :)

src/driver-holtek/holtek8-shared.c
```diff
@@ -0,0 +1,1291 @@
...
+
+	if (single_page_macros && pages_to_write > 1)
+		return -ENOMEM;
+
+	if (pages_to_write > free_pages)
+		return -ENOMEM;
+
+	while (events_i < events_to_write) {
+		struct holtek8_macro_event *ev = &macro_data.event[data_i];
+		uint8_t *ev_data = (uint8_t*) ev;
```
staticssleever668, line 317: This looks scary. Consider making holtek8_macro_event a union with uint8_t array or maybe even another struct (considering `ev_data[1] = HOLTEK8_MACRO_CMD_JUMP;` below).

https://github.com/libratbag/libratbag/blob/a52ffcdb45a120aa3b36a5620244012d0bb0c4d8/src/driver-steelseries.c#L116-L125

src/driver-holtek/holtek8-shared.c
```diff
@@ -0,0 +1,1291 @@
...
+			if (++i >= HOLTEK8_MAX_MACRO_EVENTS) {
+				rc = -ENODATA;
+				goto err;
+			}
+
+			delay += get_unaligned_be_u16((uint8_t*) &raw[i]) * 2;
+			continue;
+		}
+		if (event->key == HOLTEK8_MACRO_CMD_MOUSE) {
+			//no support in ratbag for mouse movements in macros
```
staticssleever668, line 545: Log here, maybe also add a TODO.

src/driver-holtek/holtek8-shared.c
```diff
@@ -0,0 +1,1291 @@
...
+		}
+	}
+
+	return -EINVAL;
+}
+
+/*
+ * Converts raw macro events to ratbag macro and writes to a button
+ *
+ * @param raw A buffer of macro events at least HOLTEK8_MAX_MACRO_EVENTS of size
```
staticssleever668, line 496: I would prefer passing buffer size along with the pointer, I think it's a bit less error-prone.

src/driver-holtek/holtek8-shared.c
```diff
@@ -0,0 +1,1291 @@
...
+	ratbag_button_copy_macro(button, macro);
+err:
+	ratbag_button_macro_unref(macro);
+	return rc;
+}
+
+/*
+ * Reads ratbag macro from a button and converts to raw macro events
+ *
+ * @param raw A buffer of macro events at least HOLTEK8_MAX_MACRO_EVENTS of size
```
staticssleever668, line 586: Same about buffer size:
[Jan 28, 2024](https://github.com/libratbag/libratbag/pull/1561/files#r1468903332)
I would prefer passing buffer size along with the pointer, I think it's a bit less error-prone.

src/driver-holtek/holtek8-shared.c
```diff
@@ -0,0 +1,1291 @@
...
+	if (rc < 0)
+		return rc;
+
+	rc = holtek8_read_chunked(device, (uint8_t*) &macro_data, HOLTEK8_MACRO_DATA_SIZE, NULL);
+	if (rc < 0)
+		return rc;
+
+	while (events_i < HOLTEK8_MAX_MACRO_EVENTS) {
+		struct holtek8_macro_event *ev = &macro_data.event[data_i];
+		uint8_t *ev_data = (uint8_t*) ev;
```
staticssleever668, line 229: Like below:
https://github.com/libratbag/libratbag/pull/1561/files#r1468901974
Consider making holtek8_macro_event a union with uint8_t array or maybe even another struct>> (considering the `ev_data[1] == HOLTEK8_MACRO_CMD_JUMP *here*)

src/driver-holtek/holtek8-shared.c
```diff
@@ -0,0 +1,1291 @@
...
+/*
+ * Reads macro events from the device memory.
+ *
+ * Macros in memory are divided into pages. If the device supports jumping between
+ * pages, tries to follow jumps to read multipaged macros.
+ *
+ * @return 0 on success or a negative errno.
+ */
+static int
+holtek8_read_macro_data(struct ratbag_device *device, struct holtek8_macro_event *macro_events, uint8_t macro_idx)
```
staticssleever668, line 200: Please, pass along 'macro_events' size instead of expecting it to be equal to 'HOLTEK8_MAX_MACRO_EVENTS'.

src/driver-holtek/holtek8-shared.c
```diff
@@ -0,0 +1,1291 @@
...
+		case HOLTEK8_API_B:
+			report.command = HOLTEK8B_CMD_WRITE_MACRO_DATA;
+			max_macro_index = HOLTEK8B_MAX_MACRO_INDEX;
+			break;
+	}
+
+	report.arg[1] = HOLTEK8_MACRO_DATA_SIZE;
+
+	while (events_to_write < HOLTEK8_MAX_MACRO_EVENTS && *((uint16_t*) &macro_events[events_to_write]) != 0)
+		events_to_write += 1;
```
staticssleever668, line 298: `*((uint16_t*) &macro_events[events_to_write])` - this is dangerous, it's probably unportable and I think in C++ would have even been UB.
I think something like this would be better even though it's noisier:
```c
	for (size_t i = 0; i < macro_events_size_pls_add_this_as_function_argument; ++i) {
		struct holtek8_macro_event *macro_event = &macro_events[i];
		if (macro_event->key == 0 && macro_event->delay == 0 && macro_event->release == 0) {
			break;
		}
		events_to_write += 1;
	}
```
(also yeah `macro_events_size_pls_add_this_as_function_argument`)

src/driver-holtek/holtek8-shared.c
```diff
@@ -0,0 +1,1291 @@
...
+
+			holtek8_button_macro_new_from_keycodes(button, key, 0, 0);
+			break;
+		}
+		case HOLTEK8_BUTTON_TYPE_MACRO: {
+			struct holtek8_macro_event macro_events[HOLTEK8_MAX_MACRO_EVENTS] = {0};
+
+			rc = holtek8_read_macro_data(device, macro_events, data->macro.index);
+			if (rc == -EOVERFLOW)
+				return 0;
```
staticssleever668, line 725: I don't really understand this specific check.
HOLTEK8_MAX_MACRO_EVENTS and MAX_MACRO_EVENTS are both 256, and they store more or less the same information, so in the end:
1) I would expect this check to never be true.
2) Even if would be, this would silently ignore the error. I would at least expect a warning like 'there are more macro events than libratbag can handle, showing empty macro instead'.

src/driver-holtek/holtek8-shared.c
```diff
@@ -0,0 +1,1291 @@
...
+
+/*
+ * Clears hidraw's read buffer.
+ * Prevents reading incorrect data if there was another
+ * configuration program before us.
+ *
+ * @return Number of chunks of data cleared or a negative errno.
+ */
+static int
+holtek8_clear_read_buffer(struct ratbag_device *device)
```
staticssleever668, line 907: Not a blocker, but if this is needed, it should probably be a helper in libratbag namespace instead.

staticssleever668 (changes requested):
Sorry for taking so long. Didn't look through everything, I hope I finish until the end of upcoming week. :sweat_smile:

src/driver-holtek/holtek8-shared.c
lubasowo0: This check is after reading from memory and before parsing. It can overflow if the macro which already is in device memory has too many jumps (or even infinite loops), although I'm not sure if official software can generate macro this big, but we never know. Added a log here.

lubasowo0:
Thanks for the review! I addressed all suggestions and also fixed one bug with macro writes which required some changes to macro page handling.

gadefox:
@lubasowo0 

I've got an `ASUS Echelon Laser` mouse, although `lsusb` shows `Holtek Semiconductor, Inc. USB Gaming Mouse` device name. Any idea which sensor type should I use? The mouse has 8 buttons, one LED (for DPI) and `high precision-laser` sensor. Thanks.

`cat /usr/local/share/libratbag/asus-echelon-laser.device`:
```
[Device]
Name=Asus Echelon Laser
DeviceMatch=usb:04d9:a073
DeviceType=mouse
Driver=holtek8b

[Driver/holtek8/devices/0101]
Buttons=8
DeviceName=Holtek Semiconductor, Inc. USB Gaming Mouse
SensorType=[???]

```

gadefox:
Okay, we'll see..
```
Bus 006 Device 002: ID 04d9:a073 Holtek Semiconductor, Inc. USB Gaming Mouse
Negotiated speed: Full Speed (12Mbps)
Device Descriptor:
  bLength                18
  bDescriptorType         1
  bcdUSB               2.00
  bDeviceClass            0 [unknown]
  bDeviceSubClass         0 [unknown]
  bDeviceProtocol         0 
  bMaxPacketSize0         8
  idVendor           0x04d9 Holtek Semiconductor, Inc.
  idProduct          0xa073 USB Gaming Mouse
  bcdDevice            1.00
  iManufacturer           1 Ione
  iProduct                2 USB Gaming Mouse
  iSerial                 0 
  bNumConfigurations      1
  Configuration Descriptor:
    bLength                 9
    bDescriptorType         2
    wTotalLength       0x0054
    bNumInterfaces          3
    bConfigurationValue     1
    iConfiguration          0 
    bmAttributes         0xa0
      (Bus Powered)
      Remote Wakeup
    MaxPower              100mA
    Interface Descriptor:
      bLength                 9
      bDescriptorType         4
      bInterfaceNumber        0
      bAlternateSetting       0
      bNumEndpoints           1
      bInterfaceClass         3 Human Interface Device
      bInterfaceSubClass      1 Boot Interface Subclass
      bInterfaceProtocol      1 Keyboard
      iInterface              0 
        HID Device Descriptor:
          bLength                 9
          bDescriptorType        33
          bcdHID               1.10
          bCountryCode            0 Not supported
          bNumDescriptors         1
          bDescriptorType        34 (null)
          wDescriptorLength      63
          Report Descriptors: 
            ** UNAVAILABLE **
      Endpoint Descriptor:
        bLength                 7
        bDescriptorType         5
        bEndpointAddress     0x81  EP 1 IN
        bmAttributes            3
          Transfer Type            Interrupt
          Synch Type               None
          Usage Type               Data
        wMaxPacketSize     0x0008  1x 8 bytes
        bInterval               2
    Interface Descriptor:
      bLength                 9
      bDescriptorType         4
      bInterfaceNumber        1
      bAlternateSetting       0
      bNumEndpoints           1
      bInterfaceClass         3 Human Interface Device
      bInterfaceSubClass      1 Boot Interface Subclass
      bInterfaceProtocol      2 Mouse
      iInterface              0 
        HID Device Descriptor:
          bLength                 9
          bDescriptorType        33
          bcdHID               1.10
          bCountryCode            0 Not supported
          bNumDescriptors         1
          bDescriptorType        34 (null)
          wDescriptorLength     175
          Report Descriptors: 
            ** UNAVAILABLE **
      Endpoint Descriptor:
        bLength                 7
        bDescriptorType         5
        bEndpointAddress     0x82  EP 2 IN
        bmAttributes            3
          Transfer Type            Interrupt
          Synch Type               None
          Usage Type               Data
        wMaxPacketSize     0x0008  1x 8 bytes
        bInterval               2
    Interface Descriptor:
      bLength                 9
      bDescriptorType         4
      bInterfaceNumber        2
      bAlternateSetting       0
      bNumEndpoints           1
      bInterfaceClass         3 Human Interface Device
      bInterfaceSubClass      0 [unknown]
      bInterfaceProtocol      0 
      iInterface              0 
        HID Device Descriptor:
          bLength                 9
          bDescriptorType        33
          bcdHID               1.10
          bCountryCode            0 Not supported
          bNumDescriptors         1
          bDescriptorType        34 (null)
          wDescriptorLength      26
          Report Descriptors: 
            ** UNAVAILABLE **
      Endpoint Descriptor:
        bLength                 7
        bDescriptorType         5
        bEndpointAddress     0x83  EP 3 IN
        bmAttributes            3
          Transfer Type            Interrupt
          Synch Type               None
          Usage Type               Data
        wMaxPacketSize     0x0008  1x 8 bytes
        bInterval              16
```

lubasowo0:
@gadefox Yeah this driver won't work. It may be one of those mice which don't have internal memory at all and need some daemon to be running and rebind vendor specific reports for extra buttons etc. but it would need a whole seperate driver, possibly outside of libratbag if there is no way to send configuration to the device.

gadefox:
Makes sense.. Thanks a lot!
> daemon to be running and rebind vendor specific reports for extra buttons

This is exactly what I'm looking for - at least to bind those 4 extra buttons, that's all.

deadYokai:
hi, i'm don't know how work with hid devices, but i have an `hator pulsar pro 2` with macro support
VID:PID `04d9:a09f`
have an software, but for windows only, i want to somewhat test it or try to communicate
and maybe useful info
```
VID=04D9	
PID=A09F
Sensor=3335 (PAW3335 i think)
Password=FB1512
```

lsusb output:
```
Bus 007 Device 006: ID 04d9:a09f Holtek Semiconductor, Inc. E-Signal LUOM G10 Mechanical Gaming Mouse
Negotiated speed: Full Speed (12Mbps)
Device Descriptor:
  bLength                18
  bDescriptorType         1
  bcdUSB               2.00
  bDeviceClass            0 [unknown]
  bDeviceSubClass         0 [unknown]
  bDeviceProtocol         0 
  bMaxPacketSize0         8
  idVendor           0x04d9 Holtek Semiconductor, Inc.
  idProduct          0xa09f E-Signal LUOM G10 Mechanical Gaming Mouse
  bcdDevice            3.03
  iManufacturer           1 E-Signal
  iProduct                2 Hator Pulsar
  iSerial                 0 
  bNumConfigurations      1
  Configuration Descriptor:
    bLength                 9
    bDescriptorType         2
    wTotalLength       0x0054
    bNumInterfaces          3
    bConfigurationValue     1
    iConfiguration          0 
    bmAttributes         0xa0
      (Bus Powered)
      Remote Wakeup
    MaxPower              100mA
    Interface Descriptor:
      bLength                 9
      bDescriptorType         4
      bInterfaceNumber        0
      bAlternateSetting       0
      bNumEndpoints           1
      bInterfaceClass         3 Human Interface Device
      bInterfaceSubClass      1 Boot Interface Subclass
      bInterfaceProtocol      2 Mouse
      iInterface              0 
        HID Device Descriptor:
          bLength                 9
          bDescriptorType        33
          bcdHID               1.10
          bCountryCode            0 Not supported
          bNumDescriptors         1
          bDescriptorType        34 (null)
          wDescriptorLength      71
          Report Descriptors: 
            ** UNAVAILABLE **
      Endpoint Descriptor:
        bLength                 7
        bDescriptorType         5
        bEndpointAddress     0x81  EP 1 IN
        bmAttributes            3
          Transfer Type            Interrupt
          Synch Type               None
          Usage Type               Data
        wMaxPacketSize     0x0010  1x 16 bytes
        bInterval               1
    Interface Descriptor:
      bLength                 9
      bDescriptorType         4
      bInterfaceNumber        1
      bAlternateSetting       0
      bNumEndpoints           1
      bInterfaceClass         3 Human Interface Device
      bInterfaceSubClass      0 [unknown]
      bInterfaceProtocol      0 
      iInterface              0 
        HID Device Descriptor:
          bLength                 9
          bDescriptorType        33
          bcdHID               1.10
          bCountryCode            0 Not supported
          bNumDescriptors         1
          bDescriptorType        34 (null)
          wDescriptorLength     117
          Report Descriptors: 
            ** UNAVAILABLE **
      Endpoint Descriptor:
        bLength                 7
        bDescriptorType         5
        bEndpointAddress     0x82  EP 2 IN
        bmAttributes            3
          Transfer Type            Interrupt
          Synch Type               None
          Usage Type               Data
        wMaxPacketSize     0x0020  1x 32 bytes
        bInterval               2
    Interface Descriptor:
      bLength                 9
      bDescriptorType         4
      bInterfaceNumber        2
      bAlternateSetting       0
      bNumEndpoints           1
      bInterfaceClass         3 Human Interface Device
      bInterfaceSubClass      0 [unknown]
      bInterfaceProtocol      0 
      iInterface              0 
        HID Device Descriptor:
          bLength                 9
          bDescriptorType        33
          bcdHID               1.10
          bCountryCode            0 Not supported
          bNumDescriptors         1
          bDescriptorType        34 (null)
          wDescriptorLength      28
          Report Descriptor: (length is 28)
            Item(Global): Usage Page, data= [ 0x00 0xff ] 65280
                            (null)
            Item(Local ): (null), data= [ 0x00 0xff ] 65280
                            (null)
            Item(Main  ): (null), data= [ 0x01 ] 1
                            Application
            Item(Global): (null), data= [ 0x00 ] 0
            Item(Global): (null), data= [ 0xff 0x00 ] 255
            Item(Global): (null), data= [ 0x08 ] 8
            Item(Global): (null), data= [ 0x20 ] 32
            Item(Local ): (null), data= [ 0x21 ] 33
                            (null)
            Item(Main  ): (null), data= [ 0x02 ] 2
                            Data Variable Absolute No_Wrap Linear
                            Preferred_State No_Null_Position Non_Volatile Bitfield
            Item(Local ): (null), data= [ 0x22 ] 34
                            (null)
            Item(Global): (null), data= [ 0x08 ] 8
            Item(Main  ): (null), data= [ 0x02 ] 2
                            Data Variable Absolute No_Wrap Linear
                            Preferred_State No_Null_Position Non_Volatile Bitfield
            Item(Main  ): (null), data=none
      Endpoint Descriptor:
        bLength                 7
        bDescriptorType         5
        bEndpointAddress     0x03  EP 3 OUT
        bmAttributes            3
          Transfer Type            Interrupt
          Synch Type               None
          Usage Type               Data
        wMaxPacketSize     0x0040  1x 64 bytes
        bInterval               4
Device Status:     0x0000
  (Bus Powered)
  ```

lubasowo0:
> hi, i'm don't know how work with hid devices, but i have an `hator pulsar pro 2 (HT-520)` with macro support VID:PID `04d9:a09f` have an software, but for windows only, i want to somewhat test it or try to communicate and maybe useful info

Hi! This is one of those fun devices I explained in this commit message: https://github.com/libratbag/libratbag/pull/1561/commits/bccc02a9db9faa13f55aa9f7b0c39683e561459c

It won't be hard to add support for them in this driver. The annoying issue is that it will require a small patch to the linux kernel to probe this non hid compliant interface. I can send a small patch I wrote for myself if you want to experiment but I cannot guarantee that it will continue to work as the kernel updates - that's how it is for out-of-tree kernel kernel changes until someone upstreams them. 

These devices also don't support downloading configurations from them so you'll need to make a configuration from scratch each time you open libratbag, which is not a big deal if you don't reconfigure the device very often. The windows driver solves it by storing configurations in a file but libratbag doesn't currently support it.

PR #1569, opened 2023-12-17
Fixes for ratbagctl test

lubasowo0:
Random fixes for bugs I encountered during the driver development.
The ratbagctl test is always skipped on github actions because it needs to run as root but with python 3.12 the imp module was removed so it always fails even before it can be skipped due to import error.


staticssleever668 (approved):
Tested on Arch Linux, indeed fixes failing tests. Thanks!
---
Also a note: on Arch automatic installing of the devel dbus policy breaks dbus, which is why I never ran these tests on my system...

staticssleever668:
Rebased and put description of the last commit inline.

PR #1574, opened 2023-12-31
data: add Logitech MX Anywhere 3S

vtyulb:
I add device file for logitech mx anywhere 3s (it's silent version of mx anywhere 3). Tested it on arch linux. Not sure why anywhere3.device contains usb filter, it works fine without it.


staticssleever668:
Could probably use the same SVG as Logitech MX Anywhere 3 on Piper side

PR #1580, opened 2024-01-18
add rog harpe ace wireless

dleviminzi:
Adds rog harpe ace wireless. For now it only works plugged in on USB mode. I will update if I can get it to work with the OMNI receiver.


staticssleever668 (approved):
Thanks!
cc @kyokenn for ASUS

Issue #1119, opened 2021-01-15
Logitech G PRO Superlight missing


CyanBook:
**Is your feature request related to a problem? Please describe.**
the Logitech G PRO Superlight mouse is missing. Can you add that?

**Describe the solution you'd like**
Add the mouse to the supported ones.

FFY00:
I opened #1122 which should fix this.

I also sent patches adding support for the receiver to the kernel driver.

https://www.spinics.net/lists/linux-input/msg71493.html
https://www.spinics.net/lists/linux-input/msg71495.html

They have not been merged yet, I am not sure in which Linux version this will be fixed. In the meantime, the PR still allows to configure the device via the USB connection, just not the receiver, which needs the kernel changes.

cc @brianegan

seantcanavan:
Is basic functionality for the superlight supported in general however now that #1122 is merged? It seems like the core issue for the kernel is related to using the mouse with the powerplay mat.

FFY00:
My first patch, which fixed support for the Powerplay mat, was merged, my second one, which added support for the new Lightspeed receiver that comes with the G Pro Wireless Superlight, wasn't.

Currently, with #1122:
  - device connected via USB: can be configured by libratbag
  - device connected via Powerplay receiver: can be configured by libratbag
  - device connected via Lightspeed receiver that comes with the device: **cannot be configured** by libratbag (missing kernel support)

I should fix up the patch that adds support for the new Lightspeed receiver and re-submit, but haven't had that much time.

GloriousEggroll:
I compiled the kernel with your patches @FFY00 but unfortunately it's still not showing up in ratbagctl output nor the piper gui. Is there anything additional needed on the ratbag or piper ends to work with the kernel patches?

berglh:
Having the same issue as @GloriousEggroll. Running libratbag 0.16 and piper 0.6 on 5.15.6-051506-generic and I'm not able to configure this by the Lightspeed receiver. @FFY00 is there a working kernel version for this now?

idle5:
Any updates to this? Would really appreciate if anyone could get this to work with the receiver. Would be happy to provide any information from the terminal. Thanks in advance!

berglh:
[...]

PR #1586, opened 2024-02-10
Add G705 Mouse support

Nadeflore:
This device needs the led effect settings for the first (and only) led cluster to be written in the second led slot in the profile data.

I wasn't sure how to do this without breaking the compatibility with other devices, so I conditioned it the profile type, but maybe there's a better way.


Nadeflore:
In order to get the mouse to work while connected with the lightspeed dongle, 
I had to add this device file :
```
[Device]
Name=Logitech G705
DeviceMatch=usb:046d:c547
Driver=hidpp20
DeviceType=mouse

[Driver/hidpp20]
DeviceIndex=1
```

But this makes the test `receiver-id-test` fails. Because apparently this is a generic dongle.

As as understand it, the kernel is supposed to support logitech dongles, so that each device connected to the dongle appears as a wired one, but that's not happening with this device.

src/hidpp20.h
```diff
@@ -619,6 +619,12 @@ enum hidpp20_color_led_zone_effect {
 #define HIDPP20_DPI_COUNT				5
 #define HIDPP20_LED_COUNT				2
 
+#define HIDPP20_ONBOARD_PROFILES_PROFILE_TYPE_G402	0x01
+#define HIDPP20_ONBOARD_PROFILES_PROFILE_TYPE_G303	0x02
+#define HIDPP20_ONBOARD_PROFILES_PROFILE_TYPE_G900	0x03
+#define HIDPP20_ONBOARD_PROFILES_PROFILE_TYPE_G915	0x04
+#define HIDPP20_ONBOARD_PROFILES_PROFILE_TYPE_G705	0x05
```
staticssleever668, line 626: If you are gonna move this, might as well make this an enum along the way.

```suggestion
enum hidpp20_onboard_profiles_profile_type {
	HIDPP20_ONBOARD_PROFILES_PROFILE_TYPE_G402 = 0x01,
	HIDPP20_ONBOARD_PROFILES_PROFILE_TYPE_G303 = 0x02,
	HIDPP20_ONBOARD_PROFILES_PROFILE_TYPE_G900 = 0x03,
	HIDPP20_ONBOARD_PROFILES_PROFILE_TYPE_G915 = 0x04,
	HIDPP20_ONBOARD_PROFILES_PROFILE_TYPE_G705 = 0x05,
};
```

src/hidpp20.h
```diff
@@ -916,6 +922,7 @@ struct hidpp20_profiles {
 	uint8_t sector_count;
 	uint16_t sector_size;
 	uint8_t active_profile_index;
+	uint8_t profile_format_id;
```
staticssleever668, line 925: Following previous suggestion, use the above enum:
```suggestion
	enum hidpp20_onboard_profiles_profile_type profile_format_id;
```

staticssleever668: Actually, since, if I understand right, this is read from the mouse, there may be a value we don't yet know of, and so this won't necessarily fit an enumeration, so using an int should be better.

data/devices/logitech-g705-wireless.device
```diff
@@ -0,0 +1,8 @@
+[Device]
+Name=Logitech G705
+DeviceMatch=usb:046d:c547
+Driver=hidpp20
+DeviceType=mouse
+
+[Driver/hidpp20]
+DeviceIndex=1
```
staticssleever668, line 8: > As as understand it, the kernel is supposed to support logitech receiver, so that each device connected to the receiver appears as a wired one, but that's not happening with this device.

You are right, and this file has deleted, sadly.

src/hidpp20.c
```diff
@@ -2959,6 +2957,10 @@ hidpp20_onboard_profiles_write_profile(struct hidpp20_device *device,
 
 	memcpy(pdata->profile.name.txt, profile->name, sizeof(profile->name));
 
+	// This field needs to be zero for leds to work in G705
+	// If a non zero value is set, custom animation will be used instead
+	pdata->profile.free[0] = 0x00;
```
staticssleever668, line 2962: I guess `free` implies it was assumed to be unused.
Could you then change the `free` member of the struct to only be one-byte wide, and add additional field like `set_to_zero_for_leds_for_g705`?
Be wary of endianness, the new field may need to go after the now-one-byte-wide 'free' field.

staticssleever668 (changes requested):
Hi! Thank you for your contribution! I left a few suggestions, please, look at them.

src/hidpp20.h
Nadeflore: Yes, just like this is the first device with a 0x05 profile format, there might be new devices with other profile format in the future.

data/devices/logitech-g705-wireless.device
Nadeflore: I removed it and added the pid the device reports when the receiver is actually supported by the kernel.

I was able to get it to work by applying the patch from this issue to kernel module hid-logitech-dj
https://github.com/libratbag/libratbag/issues/1119 (Issue about the same receiver (pid 0xc547))

Apparently this patch was included at some point in kernel 6.6 but was reverted because it caused issues with other devices. (https://github.com/torvalds/linux/commit/5b4ffb176d7979ac66b349addf3f7de433335e00)

src/hidpp20.c
Nadeflore: I named it `custom_animation_g705` since it appears to control the custom animation.

And this field is definitely only one byte long since the other `free` byte is actually the first byte of the crc in this device, because it has a profile sector size of 255 bytes not 256

Nadeflore:
Hi !
Thanks for you feedback.
I made the changes requested.
Let me know if there is anything else I nee to do

src/hidpp20.c
```diff
@@ -2959,6 +2958,10 @@ hidpp20_onboard_profiles_write_profile(struct hidpp20_device *device,
 
 	memcpy(pdata->profile.name.txt, profile->name, sizeof(profile->name));
 
+	// This field needs to be zero for leds to work in G705
+	// If a non zero value is set, custom animation will be used instead
```
staticssleever668, line 2962: I think this comment should go above the `custom_animation_g705` struct field instead.

src/hidpp20.c
```diff
@@ -2959,6 +2958,10 @@ hidpp20_onboard_profiles_write_profile(struct hidpp20_device *device,
...
+	pdata->profile.custom_animation_g705 = 0x00;
```
staticssleever668, line 2963: Could you put this under a check for G705? Just to make sure we don't accidentally introduce regressions for other devices. I don't have any experience with HID++ myself, so we should be extra careful.

unicornsandcandies:
Hi, I'm sorry for writing here, I just don't know where to turn. What's the state of the support for this device?

src/hidpp20.c
lat9nq: Hi, I'm attempting to take this branch and add the requested changes, but I'm not sure how to identify the device from the context. Would you mind pointing me in the right direction?

lat9nq: Oh excellent! I'll get right on that then. Thank you thank you

lat9nq:
@staticssleever668 Would it be preferable if I wrote a new PR to supersede this one? Or are you capable of merging a commit on to this branch?

PR #1621, opened 2024-06-09
Create logitech-M190

hightesto:
I added the wireless office Mouse Logitech M190 to the devices.

### DIFF
--- a/data/devices/logitech-M190
+++ b/data/devices/logitech-M190
@@ -1,0 +1,6 @@
+# Logitech M190
+[Device]
+Name=Logitech M190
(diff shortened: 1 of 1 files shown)


hightesto:
> `046d:c542` is a receiver (Nano, says Solaar https://pwr-solaar.github.io/Solaar/devices.html), which won't work correctly with libratbag. We expect Linux kernel to reexport devices connected through receivers as separate devices. Does 'lsusb' show any device? Although, it looks like this receiver is not supported by Linux.

'lsusb' shows this. The mouse is wireless, therefore the USB 2.4G reciever is plugged in.

PR #1679, opened 2025-01-22
Additional ID for ASUS ROG Keris Wireless AimPoint

bliepp:
Hey guys,

I received my ASUS ROG Keris Wireless AimPoint today and after not working with piper/llibratbag I figured out that it had a different ID, so I added mine to the corresponding device file (PID 1a66 instead of 1a68).

bliepp

### DIFF
--- a/data/devices/asus-rog-keris-wireless-aimpoint.device
+++ b/data/devices/asus-rog-keris-wireless-aimpoint.device
@@ -1,6 +1,6 @@
 [Device]
 Name=ASUS ROG Keris Wireless AimPoint
-DeviceMatch=usb:0b05:1a68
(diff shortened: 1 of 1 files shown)


bliepp:
> Does it work? Looks like libratbag may crash when it's in such mode #1683

Frankly, i cannot tell. I sent the mouse back after a week because it was causing pain in my wrist, so I have no idea if it's broken by now.

kyokenn:
I think this could be safely merged. Both USB ids look like belong to the same mouse:
https://linux-hardware.org/?id=usb:0b05-1a66
https://linux-hardware.org/?id=usb:0b05-1a68

PR #1680, opened 2025-01-26
Adding Logitech G502 X LIGHTSPEED to device list

arne-fuchs:
Only works while mouse is being connected via usb-c. Dongle not supported.
Tested via [Piper](https://github.com/libratbag/piper).


arne-fuchs:
Oh, it wasn't available until I added it manually. Are the repos on pop-os outdated?

PR #1698, opened 2025-04-29
Add Tronsmart TG007 support

Dhterech:
Adds the device indentification so it indentifies the Tronsmart TG007. ([Link for the official website](https://www.tronsmart.com/products/tg007-rgb-gaming-mouse))

I'm unsure about the button quantity though, official site says 9 (Left Click, Right Click, Scroll wheel, DPI adjustment, Light switch, Forward, Backward, Sniper button and Polling rate at the bottom of the device.)

### DIFF
--- a/data/devices/sinowealth-1007.device
+++ b/data/devices/sinowealth-1007.device
@@ -19,3 +19,9 @@ Buttons=6
 DeviceName=Marvo Scorpion G961
 LedType=RGB
 SensorType=PMW3327
+
+[Driver/sinowealth/devices/Y129]
+Buttons=7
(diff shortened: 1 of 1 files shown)


Dhterech:
Unfortunately yes, there are 9 buttons assignable and all are re-assignable, however, libratbag fails to detect any. If you force it (by changing in the code) changing of the buttons functionality and macros does work. Some specific button functions (cycle LED mode, cycle poll rate, "dpi lock" as they call it or temporarily change dpi while pressed) are shown as mapped to unknown.

I'm planning to work on it after learning a bit more how to make the proper changes if possible.

PR #1699, opened 2025-05-02
Add support for Logitech G705

nueidris:
What does not work:
- LEDs

Everything else looks fine.


nueidris:
Hi, @staticssleever668, thank you for your reply. With my PR LEDs are shown but shut down.

PR #1705, opened 2025-05-10
Add G705 Mouse support

lat9nq:
I've rebased #1586 and addressed its final PR review with help from @staticssleever668 (tyvm!). I would have preferred to PR to the original author's fork, but due to its age I just wrote a new PR. See original PR description for details.

Supersedes #1586


lat9nq:
I kept the quirk named G502X Plus for now. I was also bothered by that, but I'm of course even less sure what to do with it. I thought about adding a new G705 quirk, but I'm not sure how to handle its relation to the G502's.

Issue #710, opened 2019-05-27
Logitech G600 G-Key Can't be set to CTRL, SHIFT, or ALT


Madserge11:
https://github.com/libratbag/piper/issues/342

terencode:
Also affects G502

guppy42:
Any chance support for this will be added ?

FFY00:
> Any chance support for this will be added ?

Yes, but it will be awhile. Right now we have more pressing matters.

AtomToast:
I noticed that using keybindings such as `Shift + A` do work. It only gives an error when it's only the modifier key

guppy42:
> I noticed that using keybindings such as `Shift + A` do work. It only gives an error when it's only the modifier key

The 'only modifier' case is also where it's interesting, using a standard WASD setup reaching ALT and CTRL can be quite anoying / requires extensive finger gymnastics. Having them on the mouse would be very, very helpfull for games such as MMOs with lots of keybinds.

Hope the feature gets added in the foreseeable future, currently researching virtual HID devices but it seems like a very hacky way to do it

gerardlouw:
I have a workaround for this:
```
ratbagctl singing-gundi profile 0 button 5 action set macro +KEY_LEFTALT KEY_RESERVED -KEY_LEFTALT
```

I'm not sure what `KEY_RESERVED` is, but it doesn't seem to do anything for me. If this actually triggers some unwanted action for you, looking at [input.h](https://github.com/spotify/linux/blob/master/include/linux/input.h), some other options might be `KEY_UNKNOWN`, `KEY_F24`, or similar.

So far I've tested using the G-Key as Alt both for my KDE shortcuts (e.g. G-Key + F2 now opens KRunner) and in World of Warcraft (I am able to to self-cast by pressing G-Key with another shortcut).

guppy42:
@gerardlouw can't make that work at all 

also it seems that it would just press left alt 'unkown' and release it really quickly ?
Where as what we are trying to achive is to have say mouse 3 behave as ALT - so while it's being pressed +KEY_LEFTALT is being sendt, and when released it will send -KEY_LEFTALT (once)

Tried with both ALT and CTRL and neigther ALT+F2 or CTRL-W would trigger in cinnamon when using the mouse :(

gerardlouw:
Here is a demo of what `xev` reports when I press down my G-Key with the above settings. I did both quick and slow press-and-release for demonstration purposes.

[...]

PR #1709, opened 2025-05-18
Support single modifier key macros

Haiquan-27:
Support single modifier key macros
support:
```bash
ratbagctl
        "$device_name"
button $btn
action set macro
        KEY_LEFTALT   |
        KEY_LEFTCTRL  |
        KEY_LEFTMETA  |
        KEY_LEFTSHIFT |
        KEY_RIGHTALT  |
        KEY_RIGHTCTRL |
        KEY_RIGHTMETA |
        KEY_RIGHTSHIFT
```

> The original version seems to only support key combinations such as `CTRL+C`/`META+W`, specifically, it can only support macro combinations of multiple modifier keys and normal keys (detected using `ratbag_action_keycode_from_macro` and `ratbag_key_is_modifier`).
However for setting a single modifier key macro, it will prompt an error `"failed to update button (-22)"` (I used `ratbagctl - v "$my_device_name" button $btn action set macro KEY_LEFTALT`).
I think as a key macro, it should support a single modifier key macro, so I modified the function `ratbag_action_keycode_from_macro`, which will not affect normal key combinations and can work together with piper

> This is a correction for this [submit](https://github.com/libratbag/libratbag/pull/1630)

### DIFF
--- a/src/libratbag-private.h
+++ b/src/libratbag-private.h
@@ -460,6 +460,9 @@ ratbag_button_action_match(const struct ratbag_button_action *action,
 int
 ratbag_action_macro_num_keys(const struct ratbag_button_action *action);
 
+bool
+ratbag_action_is_single_modifier_key(const struct ratbag_button_action *action);
+
 int
 ratbag_button_macro_new_from_keycode(struct ratbag_button *button,
 				     unsigned int key,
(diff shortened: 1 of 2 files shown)


Haiquan-27:
Now, single modifier key macros and multi key macros (containing one or more modifier keys) can work properly

Haiquan-27:
see [issues](https://github.com/libratbag/libratbag/issues/710)

PR #1728, opened 2025-07-11
data: add G502 X Plus in wireless mode (usb:046d:4099)

Kvalme:
To work correctly following driver is required:
https://github.com/Kvalme/hid-logitech-dj

### DIFF
--- a/data/devices/logitech-g502-x-plus.device
+++ b/data/devices/logitech-g502-x-plus.device
@@ -1,6 +1,6 @@
 [Device]
 Name=Logitech G502 X PLUS
-DeviceMatch=usb:046d:4099,usb:046d:c095
+DeviceMatch=usb:046d:4099;usb:046d:c095
 DeviceType=mouse
 Driver=hidpp20
 


Reginald-Gillespie:
Doesn't the `X Plus` (the full, RBG mouse) use `usb:046d:c098`, or is the ID different when it's plugged in (as that's the only way I could get the ID off my kernel)?

Kvalme:
This mouse uses 2 different IDs. One for wired more and that one is 0xc095. And second one when it's connected wirelessly and it's 0x4099.
Looks like that's true for other models (502X Wireless is 0xc099 and 502X is 0xc098)

Reginald-Gillespie:
Got it - thanks. I was able to get the G502 X Plus working wirelessly on the latest arch kernal, with no other drivers, but I had to put in the ID of the receiver, `usb:046d:c547` - which surprisingly worked for me.

Reginald-Gillespie:
Yup. 
And on top of that, if I remove your driver, and put the _receiver ID_ in as the mouse ID in the ratbag .device file, it will pick up the mouse (even though receivers normally don't work like this and aren't supported) AND show me the correct macro in piper - so I'll see that `Ctrl + C` is registered under the button it's on. 

Swapping to this patch breaks macros.

Is it possible that this is a bug in this module? I was reading the patch and the discussion is from 2021 - maybe this is a whole separate issue that was since patched, but this module reverts? If not, it's either a bug that this patch causes or it's something that Arch in particular patched in their kernal. I'd love to look more into it, I'm just hardly experienced in kernal stuff.

Kvalme:
With latest changes to repo (update driver to Linux 6.19 PR (https://lore.kernel.org/lkml/q111o3n3-4so9-5nr3-4q63-51rp645n8420@xreary.bet/)) It seems to finally works fine.

staticssleever668:
@Kvalme am I right in understand that the mouse shows up as 046d:4099 with the latest kernel, without any additional kernel modules?

data/devices/logitech-g502-x-plus-wireless.device
```diff
@@ -0,0 +1,8 @@
+[Device]
+Name=Logitech G502 X PLUS
+DeviceMatch=usb:046d:4099
```
staticssleever668: nit: I think it's better to add this ID to `logitech-g502-x-plus.device` instead.
I couldn't find any separate "G502 X PLUS Wireless" mouse.
Regular G502 has separate files for wireless editions as those seem to be actually different products ([1]).
But G502 X PLUS doesn't have any separate wireless variation.  
```suggestion
DeviceMatch=usb:046d:4099,usb:046d:c095
```
[1] <https://www.logitechg.com/en-us/search?q=g502+x>

Kvalme:
> @Kvalme am I right in understand that the mouse shows up as 046d:4099 with the latest kernel, without any additional kernel modules?

It will once following PR will be merged to mainline: https://lore.kernel.org/lkml/q111o3n3-4so9-5nr3-4q63-51rp645n8420@xreary.bet/

As of today - there are no kernel that contains this. Hopefully it will happen in 6.19.

```
 logitech-djreceiver 0003:046D:C547.0006: hidraw8: USB HID v1.11 Mouse [Logitech USB Receiver] on usb-0000:75:00.0-1.1.3/input0
```

data/devices/logitech-g502-x-plus-wireless.device
Kvalme: I've made this in separate file because there are already:
[logitech-g502-x.device](https://github.com/libratbag/libratbag/blob/master/data/devices/logitech-g502-x.device)
[logitech-g502-x-wireless.device](https://github.com/libratbag/libratbag/blob/master/data/devices/logitech-g502-x-wireless.device)
[logitech-g502-x-plus.device](https://github.com/libratbag/libratbag/blob/master/data/devices/logitech-g502-x-plus.device)
[logitech-g700.device](https://github.com/libratbag/libratbag/blob/master/data/devices/logitech-g700.device)
[logitech-g700-wireless.device](https://github.com/libratbag/libratbag/blob/master/data/devices/logitech-g700-wireless.device)

But if you think it's better to have in same file - sure, let's use single file.
It just looks like for wireless mode of mouse always separate file was added

Reginald-Gillespie: Yall's naming conventions for all the previous g502 mice are screwed up, I was looking at it before. You probably want to merge it and fix the naming of all of the g502 line in another PR.

staticssleever668: As I was trying to say, G502 X Lightspeed (named G502 X Wireless in the device file) and G502 X (wired only) are actually different, separately sold mice. Meanwhile G502 X Plus only has a single version that can be connected as both wired and wireless.
G700 is like G502 X Plus in this regard, but [logitech-g700-wireless.device](https://github.com/libratbag/libratbag/blob/master/data/devices/logitech-g700-wireless.device) specifies an option not in [logitech-g700.device](https://github.com/libratbag/libratbag/blob/master/data/devices/logitech-g700.device), justifying use of separate files.

Kvalme: Done

data/devices/logitech-g502-x-plus.device
```diff
@@ -1,6 +1,6 @@
 [Device]
 Name=Logitech G502 X PLUS
-DeviceMatch=usb:046d:c095
+DeviceMatch=usb:046d:4099,usb:046d:c095
```
staticssleever668: My mistake
```suggestion
DeviceMatch=usb:046d:4099;usb:046d:c095
```

staticssleever668:
Thank you!

PR #1732, opened 2025-07-20
data: add ASUS ROG Keris (Wired)

kgmt0:
Apart from the ID, this is basically the same as the existing wireless version.


0CCULTIST:
Seems to be working fine on my ASUS Rog Keris Wired. Although there was a "ghost"/dummy device in the piper GUI & ratbagctl output that also just said ASUSTeK ROG KERIS but failed to do anything. After restarting without the device file both devices disappeared, a second restart with the device file shows only the working one. Rate, resolution & LED all work.

PR #1735, opened 2025-07-24
Add DM4 Evo (sinowealth-1007)

fenuks:
Mouse is available at https://dreammachines.pl/en/dm4evo.

Changing settings in Piper works fine, but there are few missing
bits/differences compared to the official Windows app (it can be found
at the mouse page):
- official app has an option to set DPI independently for X and Y axis
- official app allows to set DPI up to 26000 DPI although mouse
  specification says it has 16000 DPI sensor, so maybe it is a mistake
- Piper allows setting 4 modes of RGB lighting:
  + solid
  + cycle
  + breathing
  + off

  Official app has many more:
  - colourful tail
  - neon
  - colourful solid
  - flicker
  - tail
  - wave
- official app allows setting sensitivity, scrolling speed and two-click
  speed
- there is a macro editor, but I don't know how to use it, and probably
  macros are not stored directly on the mouse, but I'm adding it to the
  list for the completeness sake

I don't know if (some) of these omission are deliberate or not. I can
try to capture packets sent by the official app if needed.

### DIFF
--- a/data/devices/sinowealth-1007.device
+++ b/data/devices/sinowealth-1007.device
@@ -19,3 +19,11 @@ Buttons=6
 DeviceName=Marvo Scorpion G961
 LedType=RGB
 SensorType=PMW3327
+
+# https://dreammachines.pl/en/dm4evo
+[Driver/sinowealth/devices/M303]
+Buttons=7
+Profiles=1
+DeviceName=DM4 Evo
+LedType=RGB
+SensorType=PMW3327


staticssleever668:
Have you tried reassigning mouse buttons in Piper/ratbagctl?
(please run ratbagd in verbose mode beforehand, like `sudo systemctl stop ratbagd.service && sudo ratbagd --verbose` so that we have the original mouse configuration memory in case something goes wrong)

enkore:
> official app has an option to set DPI independently for X and Y axis

All sinowealth mice have this option but we just never surfaced it since it seems kinda useless.

> official app allows to set DPI up to 26000 DPI although mouse
specification says it has 16000 DPI sensor, so maybe it is a mistake

Could just be a typo, though there is a PixArt 26k DPI sensor around these days and iirc the SW mice all use PixArt sensors. The value is encoded in 100 DPI steps (I assume the Windows app also only allows 100 DPI steps?), so hypothetically the struct goes to 255k DPI.

fenuks:
> Have you tried reassigning mouse buttons in Piper/ratbagctl?

I'm not sure, it was a while ago. For sure, I checked lighting and DPI setting.
Will (re)check.

> All sinowealth mice have this option but we just never surfaced it since it
> seems kinda useless.

I see.

> I assume the Windows app also only allows 100 DPI steps?

Will check that later, I have virtual machine with Windows on the other computer.

PR #1707, opened 2025-05-14
Logitech G Pro X Superlight 2 and G Pro 2 Lightspeed

kyokenn:
I'm just rebased https://github.com/libratbag/libratbag/pull/1676 from the latest master, but I don't know how this driver works.
Added Logitech G Pro 2 Lightspeed.
Changed profile id, which is mentioned in this comment - https://github.com/libratbag/libratbag/issues/1572#issuecomment-2798861770 because I had the same error: "Profile layout not supported: 0x07".
Added some workarounds for ratbad-command, so it doesn't crash with "KeyError: 0" when trying to display macros.
I have tested the driver on Logitech G Pro 2 Lightspeed, most of the basic stuff works like LED, DPI setting and button mapping.

PR #1743, opened 2025-08-06
Logitech G Pro X Superlight 2 and G Pro 2 Lightspeed (CI Fixed)

daemon-mouse:
This pull request is the same as #1707, with the Python contributions formatted according to project standards. The change to the affected commit  (90fd945b55008ad4551f55f199b6d8862cc2f611) has been squashed to keep a clean commit history.

I don't possess the hardware to test this patch; but, the code has *not* changed semantically in any way from #1707, so it should work if that one did.


mcoffin:
Thank you lots to @nathanrossi , @kyokenn , and now @daemon-mouse for keeping up with this!

I bought a Pro 2 LIGHTSPEED a week or so ago, and I've been using this with great success!

One question though -- according to [nathan's comment here](https://github.com/libratbag/libratbag/issues/1572#issuecomment-2798711127), the 0x06 => 0x07 profile layout change was made between firmware versions, so it seems to me like simply replacing the ID would maybe cause issues for those unable to upgrade that have older devices?

I had done the following instead (include both IDs in code path checks for this family of profile formats), but I assume there's a reason *not* to do that and I just don't know? 6238691625129e2ed4a5cb7e29929af04f5d0dbb

For those with hardware stumbling across, I also confirmed expected operation of the sysfs battery interface via `hid_logitech_hidpp` with the relevant IDs added, and [submitted them](https://patchwork.kernel.org/project/linux-input/patch/20250820074951.1120397-1-mcoffin13@gmail.com/) for upstream.

Since others don't seem to all have hardware, here's my test list at the moment.

- [x] DPI profile changes / inc. disable/enable
- [x] LED
- [ ] ~~Button mapping~~ (leaving this out because mine *seems* to be reporting 2 more buttons than I think it has, but maybe I just don't know what I'm talkin' about)
- [x] Battery feedback
- [x] Profile management

ginzberg:
Following this.. @nathanrossi from your review, this would also cover the DEX variant of this mouse right?

ginzberg:
Following this.. @nathanrossi from your review, this would also cover the DEX variant of this mouse right?

nathanrossi:
> Following this.. @nathanrossi from your review, this would also cover the DEX variant of this mouse right?

@ginzberg The profile variations apply to the following mice, including the G Pro X Superlight 2 DEX:

* G Pro X Superlight 2
* G Pro X Superlight 2 SE
* G Pro X Superlight 2 DEX
* G Pro 2 Lightspeed

daemon-mouse:
After carefully reviewing the new patch series on #1676, it is clear that that @nathanrossi's PR contains all of the fixes in this one, resolves the discussions raised here (including the need to support the old and new device IDs), and has better documentation for the changes in the commit messages to boot.

In light of this, I believe the best thing to do is close this PR and enthusiastically recommend #1676 for merging.

For posterity, there are two meaningful differences between #1676 and this PR:

In **data/devices/logitech-g-pro-2-lightspeed.device**:
```ini
[Driver/hidpp20]
DeviceIndex=1
```

If the hardware works without this setting, we probably don't need this. Testing is advisable.

In **tools/ratbagd.py** and **tools/ratbagc.py.in**:
```python
diff --git a/tools/ratbagd.py b/tools/ratbagd.py
--- a/tools/ratbagd.py	(date 1757207404709)
+++ b/tools/ratbagd.py	(revision ea3140e4967561d38ecf1a153250fa165fc70597)
@@ -1042,7 +1042,10 @@
                         idx += 1
             except IndexError:
                 pass
-            keys.append(self._MACRO_DESCRIPTION[t](v))
+            if t not in self._MACRO_DESCRIPTION:
+                keys.append("?")
+            else:
+                keys.append(self._MACRO_DESCRIPTION[t](v))
             idx += 1
         return " ".join(keys)
 ```

This check by may be worth keeping. I do believe that foreseeable exceptions, especially KeyError, should not be floated to the user. We should strive instead to print user-friendly and brief error messages. Also, in this case we might still usefully print the macros for which descriptions do exist.

On the other hand, a missing key in this dict is *always* a bug; further, this substitution may not indicate strongly enough that one has occurred. Allowing the exception to be raised may help the project discover and fix these issues.

I've rewritten this behavior to print a more explicit and translatable message to stderr, based it on #1676, and pushed it to [daemon-mouse:logitech-gpro-x-sl2-without-keyerror](/daemon-mouse/libratbag/commit/08622cbdf673209dae9edf63e9d2dfe7c39c8e3c). I can't quite decide if it's worth pulling, due to my inexperience with the project. Ultimately, I think it depends on how likely we are to observe this behavior in real usage -- if missing macro descriptions are commonplace, the extra LOC to detect them are justified. I'll leave it up for discussion, or it can be easily added to #1676 if desired.

nathanrossi:
> In **data/devices/logitech-g-pro-2-lightspeed.device**:
> 
> ```ini
> [Driver/hidpp20]
> DeviceIndex=1
> ```
> 
> If the hardware works without this setting, we probably don't need this. Testing is advisable.

This workaround is only required if the kernel support is not available. Support for the G Pro 2 Lightspeed has been merged in the 6.17 kernel. Though agreed, testing would be ideal.

> In **tools/ratbagd.py** and **tools/ratbagc.py.in**:
> 
> ```python
> diff --git a/tools/ratbagd.py b/tools/ratbagd.py
> --- a/tools/ratbagd.py	(date 1757207404709)
> +++ b/tools/ratbagd.py	(revision ea3140e4967561d38ecf1a153250fa165fc70597)
> @@ -1042,7 +1042,10 @@
>                          idx += 1
>              except IndexError:
>                  pass
> -            keys.append(self._MACRO_DESCRIPTION[t](v))
> +            if t not in self._MACRO_DESCRIPTION:
> +                keys.append("?")
> +            else:
> +                keys.append(self._MACRO_DESCRIPTION[t](v))
>              idx += 1
>          return " ".join(keys)
> ```
> 
> This check by may be worth keeping. I do believe that foreseeable exceptions, especially KeyError, should not be floated to the user. We should strive instead to print user-friendly and brief error messages. Also, in this case we might still usefully print the macros for which descriptions do exist.
> 
> On the other hand, a missing key in this dict is _always_ a bug; further, this substitution may not indicate strongly enough that one has occurred. Allowing the exception to be raised may help the project discover and fix these issues.
> 
> I've rewritten this behavior to print a more explicit and translatable message to stderr, based it on #1676, and pushed it to [daemon-mouse:logitech-gpro-x-sl2-without-keyerror](/daemon-mouse/libratbag/commit/08622cbdf673209dae9edf63e9d2dfe7c39c8e3c). I can't quite decide if it's worth pulling, due to my inexperience with the project. Ultimately, I think it depends on how likely we are to observe this behavior in real usage -- if missing macro descriptions are commonplace, the extra LOC to detect them are justified. I'll leave it up for discussion, or it can be easily added to #1676 if desired.

I did not include that change in my update since it was unclear what macro description was missing or what exactly was being fixed by instead emitting a "?". I don't think missing macro descriptions are expected since its an internal libratbag api, https://github.com/libratbag/libratbag/blob/master/src/libratbag-enums.h#L274. The change is likely masking an underlying issue, e.g. corrupt memory/pointers/etc. as opposed to missing hardware support, so more information as to how/why the error is triggered would be helpful in determining the cause and resulting fix.

PR #1789, opened 2025-12-22
Fix size of struct `hidpp20_internal_profile`

### LINKED ISSUE #1788: Struct `hidpp20_internal_profile` has wrong size
**Information**
- `ratbagd` version (`ratbagd --version`): 0.18
- Distribution: Arch Linux
- Kernel version (ex. `uname -srmo`): Linux 6.18.1-arch1-2 x86_64 GNU/Linux

**Describe the bug**
Struct `hidpp20_internal_profile` which is read from on-board memory of logitech mice currently has size of `0x100`.
But on my Logitech G304 (same model with G305) each on-board profile has size of `0xff`. This mismatch causes `crc` field of `hidpp20_internal_profile` to have wrong value.

Actually, this bug does not affect any other part of library since `crc` field of `hidpp20_internal_profile` is not used by any other part of code.

**Additional context**
I'm not sure whether this bug is specific to my device or other device has different on-board profile size.

Quasar-Kim:
Fixes #1788.

Previously struct `hidpp20_internal_profile` had size of 0x100, but actual profile read from logitech devices has size of 0xff. This mimatch causes `crc` field of `hidpp20_internal_profile` to be corrupted.

This PR simply changes size of the struct to 0xff.

### DIFF
--- a/src/hidpp20.c
+++ b/src/hidpp20.c
@@ -1791,7 +1791,7 @@ int hidpp20_adjustable_report_rate_set_report_rate(struct hidpp20_device *device
 #define CMD_ONBOARD_PROFILES_GET_CURRENT_DPI_INDEX	0xb0
 #define CMD_ONBOARD_PROFILES_SET_CURRENT_DPI_INDEX	0xc0
 
-#define HIDPP20_PROFILE_SIZE		256
+#define HIDPP20_PROFILE_SIZE		255
 #define HIDPP20_BUTTON_HID		0x80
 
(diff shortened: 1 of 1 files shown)


staticssleever668:
> Actually, this bug does not affect any other part of library since crc field of hidpp20_internal_profile is not used by any other part of code.

Thanks, but if it doesn't introduce any visible improvements, I'd rather not get this in since there is a chance of breaking things.

PR #1808, opened 2026-01-29
data: add G502 X Wireless support via Lightspeed receiver

jszmajda:
## Summary

- Adds support for the Logitech G502 X Wireless (Lightspeed) when connected via the c547 Lightspeed receiver
- Adds wireless PID `409f` to DeviceMatch
- Adds `DeviceIndex=1` to communicate with the mouse on the receiver

## Background

The c547 Lightspeed receiver doesn't expose paired devices as separate HID devices in the kernel - it only shows the receiver itself. This means libratbag needs to:
1. Match on the receiver ID (`c547`) 
2. Use `DeviceIndex=1` to communicate with the paired mouse rather than the receiver

This is similar to how `logitech-g-pro-wireless.device` handles the same situation.

## Test plan

- [x] Verified `ratbagctl list` detects the device
- [x] Verified `ratbagctl info` shows all 11 buttons, 5 profiles, and DPI settings
- [ ] Note: `test/receiver-check.py` will fail because c547 is a known receiver ID - this may need an exception added

## Device info

```
hooting-chinchilla - Logitech USB Receiver
             Model: usb:046d:c547:0
       Device Type: Mouse
 Number of Buttons: 11
    Number of Leds: 0
Number of Profiles: 5
```

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/data/devices/logitech-g502-x-wireless.device
+++ b/data/devices/logitech-g502-x-wireless.device
@@ -1,5 +1,8 @@
 [Device]
 Name=Logitech G502 X Wireless
-DeviceMatch=usb:046d:c098
+DeviceMatch=usb:046d:c098;usb:046d:409f;usb:046d:c547
 DeviceType=mouse
 Driver=hidpp20
+
+[Driver/hidpp20]
+DeviceIndex=1


jszmajda:
Regarding the `receiver-check.py` test failure:

The c547 Lightspeed receiver behaves differently from Unifying receivers - the kernel doesn't create separate HID devices for paired mice. Instead, it only exposes the receiver itself, which means we need to match on the receiver ID and use `DeviceIndex=1` to communicate with the mouse.

Here's a proposed patch that skips the receiver check for device files that explicitly set `DeviceIndex`, since that indicates intentional receiver-based matching:

```diff
diff --git a/test/receiver-check.py b/test/receiver-check.py
index 1234567..abcdefg 100755
--- a/test/receiver-check.py
+++ b/test/receiver-check.py
@@ -60,6 +60,13 @@ def parse_data_file(path):
     data.read(path)

     matches = data["Device"]["DeviceMatch"]
+
+    # If the device file specifies a DeviceIndex, it's intentionally matching
+    # on a receiver ID to communicate with a paired device (e.g., Lightspeed
+    # receivers that don't expose paired devices as separate HID devices)
+    if data.has_option("Driver/hidpp20", "DeviceIndex"):
+        return []
+
     return matches.split(";")
```

This approach automatically allows any device file that uses `DeviceIndex` to include receiver IDs, since that's the legitimate use case where receiver matching is required.

I'm happy to include this in the PR or submit it separately - let me know your preference.

staticssleever668:
Hi! Thanks for the PR.

> The c547 Lightspeed receiver doesn't expose paired devices as separate HID devices in the kernel - it only shows the receiver itself. 

I think this was implemented in one of the last Linux releases. :)
#1728

And I'm sure if specifying `DeviceIndex` is a good idea. As far as my understanding of hidpp20 libratbag driver goes, `DeviceIndex` depends on what devices are connected to the system as a given moment, which can be problematic with several devices connected at once.

PR #1822, opened 2026-03-10
Add Attack Shark driver and Attack Shark X11 Device profile

kmavrov:
# Attack shark

### Special thanks

First i want to thank to @HarukaYamamoto0 for the amazing work done on the full reverse engineering of the Attack Shark driver.

Check out his implementation here: https://github.com/HarukaYamamoto0/attack-shark-x11-driver

### Overview

This PR adds a new driver for the Attack Shark X11 gaming mouse, a symmetrical wireless mouse sold under VID `0x1D57` with two product IDs — `0xFA55` (wired) and `0xFA60` (2.4 GHz wireless adapter). A single driver entry handles both variants, detecting wired vs. wireless at probe time from the USB product ID and adjusting packet sizes accordingly.

The driver was reverse-engineered from an existing open-source TypeScript implementation for Linux.

---

### Supported features

|     |     |
| --- | --- |
| Feature | Detail |
| DPI stages | 6 stages, range 50–22 000 DPI |
| Polling rate | 125 / 250 / 500 / 1000 Hz |
| Buttons | 8 fully remappable buttons |
| RGB LED | Off, Static, Breathing, Cycle (rainbow) |
| Angle snapping | On/Off |
| Debounce | 8 levels (4–50 ms) |

The device has no hardware read-back protocol, so the profile is declared write-only (`RATBAG_PROFILE_CAP_WRITE_ONLY`).

---

### Protocol

All configuration is sent via USB HID `SET_FEATURE` requests to Interface 2. There are four report types:

|     |     |     |
| --- | --- | --- |
| Report | Purpose | Size |
| `0x04` | DPI stages + angle snap | 52 B (wired) / 56 B (wireless) |
| `0x05` | LED, debounce, sleep timers | 13 B (wired) / 15 B (wireless) |
| `0x06` | Polling rate | 9 B |
| `0x08` | Button mapping | 59 B |

Each report uses its own checksum algorithm. A mandatory 250 ms inter-transfer delay is required to prevent firmware lockup.

DPI encoding uses a 411-entry lookup table with `high_flag` and `stage_mask` fields to disambiguate byte values shared between the 10 100–12 000 and 20 100–22 000 DPI ranges.

---

### Files changed

|     |     |
| --- | --- |
| File | Change |
| `src/driver-attackshark.c` | New driver |
| `data/devices/attack-shark-x11.device` | Device descriptor (both PIDs) |
| `meson.build` | Added driver source to build |
| `src/libratbag.c` | Registered `attackshark_driver` |
| `src/libratbag-private.h` | Added `extern struct ratbag_driver attackshark_driver` |
| `src/libratbag-data.c` | Added `ATTACKSHARK` to `enum driver` and `driver_map[]` |

### DIFF
--- a/data/devices/attack-shark-x11.device
+++ b/data/devices/attack-shark-x11.device
@@ -1,0 +1,6 @@
+[Device]
+Name=Attack Shark X11
+# Wired (USB cable) and 2.4 GHz wireless adapter share the same driver.
+DeviceMatch=usb:1d57:fa55;usb:1d57:fa60
+DeviceType=mouse
+Driver=attackshark
--- a/meson.build
+++ b/meson.build
@@ -186,6 +186,7 @@ src_libratbag = [
 	'src/driver-sinowealth.c',
 	'src/driver-sinowealth.h',
 	'src/driver-sinowealth-nubwo.c',
+	'src/driver-attackshark.c',
 	'src/driver-marsgaming/driver-marsgaming.c',
 	'src/driver-marsgaming/marsgaming-buttons.c',
 	'src/driver-marsgaming/marsgaming-command.c',
--- a/src/driver-attackshark.c
+++ b/src/driver-attackshark.c
@@ -670,7 +670,7 @@ attackshark_build_prefs_packet(uint8_t *buf, size_t bufsz,
 	/* Fixed deep sleep of 10 minutes and sleep timer of 0.5 min.
 	 * These are not yet exposed via a ratbag API; use sensible defaults. */
 	const unsigned int deep_sleep_min = 10;
-	const unsigned int sleep_min_half = 1;  /* 0.5 min → value 1 */
+	const unsigned int sleep_min_half = 4;  /* 0.5 min → value 1 */
 
 	uint8_t led_speed = attackshark_ms_to_led_speed(drv->led_ms);
 	uint8_t hw_speed  = (uint8_t)(6 - led_speed);
--- a/src/driver-attackshark.c
+++ b/src/driver-attackshark.c
@@ -1,0 +1,1170 @@
+/*
+ * Copyright © 2025 AttackTS contributors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a
+ * copy of this software and associated documentation files (the "Software"),
+ * to deal in the Software without restriction, including without limitation
+ * the rights to use, copy, modify, merge, publish, distribute, sublicense,
+ * and/or sell copies of the Software, and to permit persons to whom the
+ * Software is furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice (including the next
+ * paragraph) shall be included in all copies or substantial portions of the
+ * Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.  IN NO EVENT SHALL
+ * THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+ * FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+ * DEALINGS IN THE SOFTWARE.
+ */
+
+/*
+ * Driver for the Attack Shark X11 gaming mouse (wired and wireless 2.4GHz).
+ *
+ * Protocol reverse-engineered from the AttackTS TypeScript driver:
+ *   https://github.com/ilievsoft4/AttackTS
+ *
+ * The device exposes 4 HID interfaces. Configuration is done through
+ * Interface 2 (hidraw, vendor-specific) via HID SET_FEATURE reports.
+ * The driver uses a write-only model (no read-back from device RAM).
+ *
+ * VID: 0x1d57 (Xenta / Beken)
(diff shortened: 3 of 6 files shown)


HarukaYamamoto0:
Hey, I'm really happy to see this PR, thank you so much for taking the time to port the AttackTS protocol to a proper libratbag driver. I've been testing it and most things are working great. I'll do a more thorough test tonight, but here's what I've found so far:

**Working ✓**
- Device detection and probe
- DPI stages
- Polling rate
- LED modes

**Needs fixing before merge**

**`ATTACKSHARK_TRANSFER_DELAY_US` is too low**, the current value `60000u` (60ms) causes the mouse to lock up when a full commit is sent (up to 4 packets in sequence). The firmware needs ~250ms between packets, so `250000u` would be the safe minimum.

**`as_debounces` is incomplete** — the hardware accepts the full even range from 4 to 50ms (step 2). The current array exposes only 8 values, which unnecessarily limits what users can set. The full list is: 4, 6, 8, 10, 12, 14, 16, 18, 20, 22, 24, 26, 28, 30, 32, 34, 36, 38, 40, 42, 44, 46, 48, 50.

**Copyright attribution** — the current header says `AttackTS contributors` which is a bit vague. it would be clearer to have both contributions explicitly named so anyone reading the file knows exactly who did what:
```
 * Copyright © 2026 kmavrov (libratbag C driver)
 * Copyright © 2026 HarukaYamamoto0 (reverse engineering)
```

**Protocol source link** — the referenced repo `https://github.com/ilievsoft4/AttackTS` is private, so anyone reviewing this PR won't be able to access it.

Again, really appreciate the work you put into this!

I'll definitely do more testing tonight to make sure everything is good, since I have the physical mouse I can keep testing and comparing the packets.

kmavrov:
The debounce is limited to 8 values only because of a restriction comming from https://github.com/libratbag/libratbag/blob/874c01732a7d3c074baefd8055c0d3efe8c9a935/src/libratbag-private.h#L275
Where the debounces are predefined to 8 and i don't want to mess with core files, since this is far away from my daily programming

As for the contributions - made some changes, but this is my first bigger attempt for a contribution so any additional feedback is highly appreciated.

HarukaYamamoto0:
## Testing Notes

I tested this PR against my physical X11 and everything is working correctly with Piper. Here's what I validated:

- DPI Protocol (Report 0x04) ✅
- User Preferences Protocol (Report 0x05) ✅
- Polling Rate Protocol (Report 0x06) ✅


### Transfer delay

The 250ms value in `ATTACKSHARK_TRANSFER_DELAY_US` is a confirmed hard minimum, I tested lower values and anything below that causes the mouse to stop responding entirely. The original configuration panel also exhibits this same ~250ms delay between writes, so this is a firmware constraint, not a driver choice.

### Hardcoded sleep values

Lines 673-674 `deep_sleep_min = 10` and `sleep_min_half = 4` (equivalent to 2 minutes). Since these aren't exposed via the ratbag API and will stay static, 2 minutes feels more reasonable as a default from a daily usage perspective — 1 minute is too aggressive.

### Known limitation: write-only LED state

Because the device is write-only, the LED state resets to the driver default (off) whenever the daemon restarts. This is an inherent protocol limitation and not fixable.

### Macros

I'd recommend keeping macro support out of this PR. The protocol has two distinct macro types, simple key mapping (Report 0x08) and custom macros (Report 0x09), and they work very differently. The custom macro path involves a 4-packet sequence with non-trivial delay encoding, which is complex enough to deserve its own PR and review cycle. Mixing it here would make this harder to review and increase regression risk. The current scope is solid as-is.

kmavrov:
> ### Hardcoded sleep values
> Lines 673-674 `deep_sleep_min = 10` and `sleep_min_half = 4` (equivalent to 2 minutes). Since these aren't exposed via the ratbag API and will stay static, 2 minutes feels more reasonable as a default from a daily usage perspective — 1 minute is too aggressive.

I have updated the sleep time also, please verify.

HarukaYamamoto0:
Sleep time verified, working correctly.

One more thing, could you remove `RATBAG_BUTTON_ACTION_TYPE_KEY` and `RATBAG_BUTTON_ACTION_TYPE_MACRO` from the `ratbag_button_enable_action_type` calls in `attackshark_probe`? Custom macro support is intentionally out of scope for this PR due to the complexity of the protocol (two distinct macro types, one of which requires a 4-packet sequence with non-trivial delay encoding). Keeping KEY and MACRO enabled in the probe creates a misleading contract, the action types appear supported but the commit path doesn't fully handle them.

Also worth noting: Piper may still display the macro option in its UI regardless of the declared action types, but the ratbag layer correctly rejects it at the D-Bus level. This is a Piper UI issue, not a driver bug, just mentioning it so nobody opens an issue thinking it's a regression.

HarukaYamamoto0:
Another reason to keep custom macros out permanently: the `0x08` packet that assigns a macro slot to a button requires knowing the complete current button mapping. On a write-only device, that state is lost on every daemon restart, so any macro assignment would risk silently corrupting the other button mappings. There's no clean solution for this without read-back support, which this device doesn't have.

The official panel solves this by maintaining the internal state of the user's preferences within the application, and every time the panel was opened it immediately applied the user's preferences to the mouse, thus synchronizing it. But from what I've seen, it's not possible to do this with libratbag or Piper.

HarukaYamamoto0:
I forgot to mention, but in the driver I made in TS, I solved this by optionally requesting the button mapping builder; that resolved the problem.: https://github.com/HarukaYamamoto0/attack-shark-x11-driver/blob/f0bc5d32a27bfb9604ff7e455d74fef81fad70ef/example.ts#L45

HarukaYamamoto0:
Yeah, unfortunately this is as far as we can get here. It’s due to a limitation of libratbag itself, which doesn’t provide a mechanism to persist settings. Even though the other fields work, they’ll always end up resetting because the mouse is write-only and expects the control panel to know the exact current state.

I’m looking into learning Rust programming language so I can rewrite the repository attack-shark-x11-driver, and use something like napi-rs to bridge it with Node.js, making it possible to build control panels with Tauri, Electron, etc.

Issue #1823, opened 2026-03-19
Logitech G733 (046d:0afe) matched by hidpp20 but exposes 0 LEDs after ERR_LOGITECH_INTERNAL (05)


jalovisko:
**Information**

- `ratbagd` version (`ratbagd --version`): 0.18
- Distribution: Arch Linux
- Kernel version (ex. `uname -srmo`): `Linux 6.19.8-arch1-1 x86_64 GNU/Linux`

Device info (if applicable):
```
$ ratbagctl "bellowing-paca" info
bellowing-paca - Logitech G733 Gaming Headset
             Model: usb:046d:0afe:0
       Device Type: Mouse
 Number of Buttons: 0
    Number of Leds: 0
Number of Profiles: 1
Profile 0: (active)
  Name: n/a
```

**Describe the bug**
Logitech G733 (USB ID `046d:0afe`) is detected and matched by `hidpp20`, but exposes `Number of Leds: 0` even though the device supports RGB (2 zones).

From verbose logs, the device reports the HID++ feature `HIDPP_PAGE_COLOR_LED_EFFECTS (0x8070)` and `libratbag` logs `device has color effects`, but initialization fails with `ERR_LOGITECH_INTERNAL (05)`. After this failure, no LEDs are exposed.

The same device is listed in `logitech-g733.device` with:
```
DeviceMatch=usb:046d:0ab5;usb:046d:0afe
```

So it appears support for `0afe` is incomplete or behaves differently from `0ab5`.

**To Reproduce**
1. Start `ratbagd`
2. Run
    ```bash
   ratbagctl list
   ratbagctl "bellowing-paca" info
   ```
3. Observe `Number of Leds: 0`

**Logs**
Please start the daemon with verbosity and reproduce the issue.

First make sure it isn't running already.
```
$ ratbagd --verbose=raw
Starting ratbagd version 0.18 (API version 2)
Initializing libratbag
ratbag debug: New device: Logitech G733 Gaming Headset
ratbag debug: Using data directory '/usr/share/libratbag'
ratbag debug: device assigned driver hidpp20
ratbag debug: hidraw info: bus 0x03 vendor 0x46d product 0xafe
ratbag debug: Logitech G733 Gaming Headset is device '/dev/hidraw7'.
ratbag debug: Parsing HID report descriptor
ratbag debug: Parsing HID report descriptor
ratbag debug: - HID report ID 02
ratbag debug: - HID report ID 11
ratbag debug: - HID report ID 06
[...]

PR #1824, opened 2026-03-19
Add quirk for Logitech G733 to handle LED zone count

jalovisko:
The G733 wired variant (USB ID `046d:0afe`) returns `ERR_LOGITECH_INTERNAL` when queried with `GET_INFO` (feature 0x8070, fn0), causing LED initialization to abort and exposing 0 LEDs despite the hardware having 2 RGB zones.

Added `HIDPP20_QUIRK_G733` which falls back to a hardcoded zone count of 2 when `GET_INFO` fails. Individual zone queries (`GET_ZONE_INFO`) succeed normally, so both LED zones are exposed and controllable. The fallback only activates on failure, leaving the wireless variant (046d:0ab5) unaffected.

Fixes the issue reported with `ratbagctl` info showing `Number of Leds: 0` on the G733 wired headset (see #1823).

Tested on my G733 by running the following and observing the colours change:
```
ratbagctl "bellowing-paca" led 0 set mode on color ff0000                                                         
ratbagctl "bellowing-paca" led 1 set mode on color 0000ff
```

### DIFF
--- a/data/devices/logitech-g733.device
+++ b/data/devices/logitech-g733.device
@@ -6,3 +6,4 @@ Driver=hidpp20
 
 [Driver/hidpp20]
 Buttons=0
+Quirk=G733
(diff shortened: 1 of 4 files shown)


staticssleever668 (approved):

staticssleever668 (changes requested):
```
../src/hidpp20.c: In function ‘hidpp20_onboard_profiles_write_profile’:
../src/hidpp20.c:2992:25: error: ‘ledinfo’ undeclared (first use in this function)
 2992 |                         ledinfo.zone_count = 2;
      |                         ^~~~~~~
```

jalovisko:
My bad, @staticssleever668, I introduced the fallback logic in the wrong place. Should be good now.

PR #1861, opened 2026-06-18
ratbagd: line buffer stdout/stderr

mvnetbiz:
When stdout is a TTY, printing is line buffered, but when running under a service manager and stdout is a pipe/file printing is fully buffered by default, so you don't see any logging until several KB is printed, making debugging confusing. Always set line buffered on stdout/stderr to fix this.

### DIFF
--- a/ratbagd/ratbagd.c
+++ b/ratbagd/ratbagd.c
@@ -625,6 +625,9 @@ int main(int argc, char *argv[])
 	setrlimit(RLIMIT_CORE, &corelimit);
 #endif
 
+	setlinebuf(stdout);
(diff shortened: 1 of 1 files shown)


mvnetbiz:
With this change, when watching with `journalctl -u ratbagd -f` and starting the daemon, I immediately see `hidraw4: "Logitech PRO X", 5 profiles`. Without it, the message is never seen, and once ratbagd exits from the idle timer, the message is permanently lost.

staticssleever668:
I can reproduce this with `sudo ./build/ratbagd | cat`. Do you have examples of other projects having to do this?

PR #1799, opened 2026-01-11
hidpp20: add quirk to fix active profile index on some devices

krombo-kode:
Addresses index offset errors caused by Logitech devices returning offset indexes for active profile query.

resolves #1323, resolves #1606
resolves libratbag/piper#1083

(G502-Hero, G502-Hero-Wireless, G502-Proteus-Spectrum, G604)

Did not apply to any other devices, as they are not available to me for testing.

PR #1863, opened 2026-06-22
feat: multi quirk g502xplus

perfectra1n:
The G502 X Plus needs multiple quirks, so I put together this `quirks` system, using the bitmask idea previously mentioned. Please let me know if you hate it, and I can take it out and go down to a single line. 

This PR implements the changes needed to have the mouse cooperate with DPI changes.

Commit 1- hidpp20: support multiple quirks per device:
- Turns the existing enum hidpp20_quirk from plain values into bit flags.
- Changes the stored field from "one quirk" (enum … quirk) to "a set of quirks" (uint32_t quirks).
- Rewrites the ~10 quirk == X checks to quirks & X.
- Lets device files specify a Quirks=A;B list (legacy single Quirk= still works).
- Renames the getter get_quirk → get_quirks; iterates flags in the debug log; permits the Quirks key in the test.
- No quirk behavior is added or changed here - it's purely "let a device carry more than one of the quirks that already exist."

Commit 2 - data: enable INDEX_OFFSET on the G502 X PLUS (the fix):
- Changes the device file from Quirk=G502X_PLUS to Quirks=G502X_PLUS;INDEX_OFFSET.

The two quirks being combined, both pre-existing:
- G502X_PLUS (added with the X PLUS support): use LED slot index 1 for the single LED, and zero the onboard custom-animation field on write.
- INDEX_OFFSET (added in #1799): decrement the device-reported active-profile index by 1, for mice that report it 1-indexed.

### DIFF
--- a/src/libratbag-data.c
+++ b/src/libratbag-data.c
@@ -67,7 +67,7 @@ enum driver {
 
 struct data_hidpp20 {
 	int index;
-	enum hidpp20_quirk quirk;
+	uint32_t quirks; /* bitmask of enum hidpp20_quirk */
 	int led_count;
 	int report_rate;
 	int button_count;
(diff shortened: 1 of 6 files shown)


src/libratbag-data.c
```diff
@@ -218,18 +232,19 @@ init_data_hidpp20(struct ratbag *ratbag,
...
-		else if(streq(str, "INDEX_OFFSET"))
-			data->hidpp20.quirk = HIDPP20_QUIRK_INDEX_OFFSET;
+		data->hidpp20.quirks |= hidpp20_quirk_from_string(str);
+		g_free(str);
 	}
+
+	gsize quirks_count = 0;
+	_cleanup_(g_strfreevp) char **quirks = g_key_file_get_string_list(keyfile, group, "Quirks", &quirks_count, NULL);
+	for (gsize i = 0; quirks && i < quirks_count; i++)
+		data->hidpp20.quirks |= hidpp20_quirk_from_string(quirks[i]);
```
staticssleever668: No need for a separate 'Quirks' that does the same thing. Just make 'Quirk' take a list.

perfectra1n: Done :)

perfectra1n:
Since the comment was anchored on the keyfile-parsing hunk, I'm assuming that was the only `Quirks` you wanted cleaned up - and that you were okay with `Quirks` in the other changes

staticssleever668 (approved):
Thanks!

Issue #1206, opened 2021-10-07
Maintainers needed


whot:
This issue is a note about libratbag's maintainership status.

Right now, none of the maintainers have sufficient (or any) time to work on libratbag. Which means all but the simplest issues will go unanswered. Pull requests will not be reviewed for extended periods of time.

This is unlikely to change unless we get more maintainers into the fold.

You can help out by reviewing patches/pull requests and helping users to debug the issues you're seeing.

You can help out by expanding the wiki to explain how to debug issues.

Please do so, otherwise libratbag will just sit there largely stale for the foreseeable future.

perigoso:
Definitely want to help, but between all the stuff I want to do I don't know if I can find much time

FFY00:
I have turned off notifications for issues, so please ping me directly if there is anything you want me to look at.

perigoso:
> Not completely related, I know someone at Roccat, I'll ask if they(the company) want to contribute for Roccat devices, I doubt there's interest but you never know.

They brought it up, sent a couple emails, and went basically ignored. about what would be expected, worth the try.

euclip:
I don't have the time or expertise to step up, but would it be possible or helpful to donate money instead?

whot:
IMO not really, the main issue is lack of time, not lack of money. You'd need enough money to employ someone or do some contracting but that's a rather high hurdle. And you'd still need to find someone with time available.

ZXSwire3:
what kind of knowledge is required? I don't think that I have what would be required to help out, but I'd be interested in learning so that I could help out if I can

whot:
depends on what you want to do. For libratbag *maintenance* you need to know C but you don't really need to know how to RE things, that's left to those who work on the drivers. Most of the maintenance is making sure the internal APIs are correct, etc.

For reverse-engineering a driver you need to know C and how to interpret data dumps, most of the RE-ing is observing what the vendor drivers send to the device and then interpreting that and matching it in code. You'll find that the drivers themselves are reasonably simple - the hard bit is knowing what bit does what, not how to set those bits.

Virepri:
If I wanted to investigate into #1241 and/or #710, where would you guys advise starting? Would I need any RE skills? If so, I'm not opposed to learning if you'll point me in the right direction.

amfern:
@Virepri an issue that reproduces intermediately presents an additional uncertainty factor, so the first step would be to narrow down the factors that reproduce the issue.
[...]

PR #1864, opened 2026-06-25
Add Lemokey G1 8K mouse driver

cebarks:
## Summary

- Add a new driver for the **Lemokey G1 8K** wireless gaming mouse (VID:PID `362d:d028`, "Ultra-Link 8K" receiver, PixArt PAW3950 sensor)
- The protocol was reverse-engineered from [launcher.lemokey.com](https://launcher.lemokey.com) and verified on hardware
- Driver communicates with the receiver dongle over a vendor HID interface (usage page `0xFFC1`) using 64-byte and 21-byte reports

### Supported features (read + write)

| Feature | Range | Command |
|---------|-------|---------|
| Polling rate | 250, 500, 1000, 2000, 4000, 8000 Hz | `0x41` |
| DPI | Up to 5 levels, 50–30000 | `0x40` (write), `0x06` (read) |
| Angle snapping | On/off | `0x42` |
| Debounce | 0–12 ms | `0x43` |
| Connection check | Connected/disconnected | `0x03` |
| Firmware version | Read-only | `0x04` |

### Not yet supported

Button remapping and LED control require the NAPE protocol (HID report ID 0), which is not accessible via Linux hidraw on this device. The launcher's web app uses WebHID on macOS where report ID 0 handling differs.

### Files changed

- `src/driver-lemokey.c` — new driver (~614 lines)
- `data/devices/lemokey-g1.device` — device match file
- `meson.build`, `src/libratbag-data.c`, `src/libratbag-private.h`, `src/libratbag.c` — driver registration

## Test plan

- [x] Builds clean with `-Wall -Wextra -Wmissing-prototypes -Wstrict-prototypes` (zero warnings)
- [x] Polling rate read/write verified on hardware
- [x] DPI read/write verified on hardware (5 levels, set/readback confirmed)
- [x] Angle snapping on/off verified on hardware
- [x] Debounce write verified on hardware
- [x] Connection check returns `-ENODEV` when mouse is off
- [x] Firmware version reads "0.1.2" correctly
- [x] Probe gracefully skips non-matching hidraw interfaces
- [x] Tested via ratbagd.devel + busctl

Implemented and reverse engineered with assistance from [Claude Code](https://claude.ai/code)

### DIFF
--- a/data/devices/lemokey-g1.device
+++ b/data/devices/lemokey-g1.device
@@ -1,0 +1,5 @@
+[Device]
+Name=Lemokey G1 8K
+DeviceMatch=usb:362d:d028
+Driver=lemokey
+DeviceType=mouse
--- a/meson.build
+++ b/meson.build
@@ -193,6 +193,7 @@ src_libratbag = [
 	'src/driver-marsgaming/marsgaming-leds.c',
 	'src/driver-marsgaming/marsgaming-probe.c',
 	'src/driver-marsgaming/marsgaming-query.c',
+	'src/driver-lemokey.c',
 	'src/driver-test.c',
 	'src/libratbag.c',
 	'src/libratbag.h',
--- a/src/driver-lemokey.c
+++ b/src/driver-lemokey.c
@@ -1,0 +1,614 @@
+/*
+ * Copyright © 2024-2026 Anten Skrabec
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a
+ * copy of this software and associated documentation files (the "Software"),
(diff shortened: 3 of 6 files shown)


staticssleever668:
We don't have the capacity to review new drivers - #1206. I noticed this has some LLM assisted changes, we have no precedent for adding such changes yet, and I'm not sure that they are acceptable due to unclear copyright status.

cebarks:
Fair enough, the LLM mainly helped with the reverse engineering process if it matters. This was all done on real hardware as well, so it is confirmed working.

PR #1865, opened 2026-07-04
feature: Add support for Glorious Model I

SixFlix:
No one has created a device file for the Model I, so I decided I might try it and quickly created this.
Not sure how well it'll work considering this is the Model O device file but with some information replaced, including 9 buttons instead of 6.
I'm a novice with Git/GitHub and barely understand what I'm doing with libratbag so I'd appreciate any help with implementing support for the Model I/learning how to use PRs and GitHub stuff.

### DIFF
--- a/data/devices/glorious-model-i.device
+++ b/data/devices/glorious-model-i.device
@@ -1,0 +1,11 @@
+[Device]
+Name=Glorious Model I
+DeviceMatch=usb:22d4:1503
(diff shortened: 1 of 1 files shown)


staticssleever668:
Judging by the USB vendor ID - 22d4 - being different from other Sinowealth devices, this driver is unlikely to work as-is. Sorry.

PR #1868, opened 2026-07-06
Add support for Logitech G305 X SUPERLIGHT

jakobhaervig:


staticssleever668 (changes requested):
c547 is the receiver, not the mouse

Issue #1606, opened 2024-04-04
incorrect active profile


roshal:
**subject**

ratbagctl skips the active profile index one forward in a circle even if next profile is disabled

**ratbag**

`ratbagctl --version` 0.17

`ratbagd --version` 0.17

**system**

`uname --kernel-name` Linux

`uname --kernel-release` 6.6.22-1-lts

`uname --machine` x86_64

`uname --operating-system` GNU/Linux

**device**

`ratbagctl G604 info`

```
Logitech G604 
             Model: usb:046d:4085:0
 Number of Buttons: 16
    Number of Leds: 1
Number of Profiles: 5
Profile 0:
  ..
Profile 1: (disabled) (active)
  ..
Profile 2: (disabled)
  ..
Profile 3: (disabled)
  ..
Profile 4: (disabled)
  ..
```

**reproduction**

`ratbagctl G604 profile 0 enable`
`ratbagctl G604 profile 1 enable`
`ratbagctl G604 profile 2 enable`
`ratbagctl G604 profile 3 enable`
`ratbagctl G604 profile 4 enable`

`ratbagctl G604 profile active set 0`
`ratbagctl G604 profile active get` 0
`sudo systemctl restart ratbagd.service`
`ratbagctl G604 profile active get` 1

`ratbagctl G604 profile active set 1`
`ratbagctl G604 profile active get` 1
`sudo systemctl restart ratbagd.service`
`ratbagctl G604 profile active get` 2

`ratbagctl G604 profile active set 2`
`ratbagctl G604 profile active get` 2
`sudo systemctl restart ratbagd.service`
`ratbagctl G604 profile active get` 3

`ratbagctl G604 profile active set 3`
`ratbagctl G604 profile active get` 3
`sudo systemctl restart ratbagd.service`
`ratbagctl G604 profile active get` 4

`ratbagctl G604 profile active set 4`
`ratbagctl G604 profile active get` 4
`sudo systemctl restart ratbagd.service`
`ratbagctl G604 profile active get` 0

**references**

pulls #811 #1295 #1304

issues #1323 #1464

atomGit:
this isn't specific to the G604

```
$ ratbagctl warbling-mara profile active get
2 # <-- wrong - profile 1 is active
$ ratbagctl warbling-mara profile 2 get
Profile 2: (disabled) (active) # <-- how can it be disabled and active?
[...]

PR #1872, opened 2026-07-18
data: enable INDEX_OFFSET on G502 X and G502 X Wireless

Nephyrin:
f9d1574 enabled this for the G502 X PLUS, which is the RGB variant. I can confirm the normal G502 X Wireless has this as well, resulting in profile weirdness.  I do _not_ have a base model G502 X, but comments on issue #1606 as well as logic suggest it very likely suffers from this unless proven otherwise.

### DIFF
--- a/data/devices/logitech-g502-x-wireless.device
+++ b/data/devices/logitech-g502-x-wireless.device
@@ -3,3 +3,6 @@ Name=Logitech G502 X Wireless
 DeviceMatch=usb:046d:c098;usb:046d:409f
 DeviceType=mouse
 Driver=hidpp20
+
+[Driver/hidpp20]
+Quirk=INDEX_OFFSET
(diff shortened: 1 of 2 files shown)


mgth:
Confirming this on a G502 X LS (046d:c098 wired / 046d:409f through the LS dongle). Since 43f76ee added the 409f match, ratbagd reads the active profile 1-indexed over the dongle: at mouse power-on Piper shows the factory-default profile (index 1) as active instead of the actual profile 0 — exactly what f9d1574 fixed for the G502 X PLUS.

Raw HID++ queries right after power-on (without ratbagd) show the mouse really is in onboard mode with its first profile active (correct DPI and report rate), while `getCurrentProfile` returns 1 — so the 1-indexed report is confirmed on this model too. Wired (c098) behaves correctly without the quirk.

With `Quirk=INDEX_OFFSET` applied locally to `logitech-g502-x-wireless.device` (same as this PR), the device probes fine and the active profile reads correctly after power-on.

staticssleever668 (approved):

PR #1877, opened 2026-08-04
Add: SOAI C18 gaming mouse driver

PadreAdamo:
Adds driver-c18.c for the SOAI C18 gaming mouse (VID:PID 12c9:2003) — profiles, polling rate, DPI (all 5 stages), buttons (native clicks, keys, media keys, DPI/profile functions, modifier shortcuts, Combo Key), full macro upload/playback (multi-packet, wait-blocks, repeat modes at the protocol level), and LED (off/static/breathing/cycle). Protocol reverse-engineered from USB captures, documented in full at PadreAdamo/c18_reverseengineer (https://github.com/PadreAdamo/c18_reverseengineer).

### DIFF
--- a/data/devices/soai-c18.device
+++ b/data/devices/soai-c18.device
@@ -1,0 +1,5 @@
+[Device]
+Name=SOAI C18 Gaming Mouse
+DeviceMatch=usb:12c9:2003
+Driver=c18
+DeviceType=mouse
--- a/meson.build
+++ b/meson.build
@@ -171,6 +171,7 @@ src_libratbag = [
 	'src/libratbag-enums.h',
 	'src/libratbag.h',
 	'src/driver-asus.c',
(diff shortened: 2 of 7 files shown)


staticssleever668:
We don't have the capacity to review new drivers - #1206. I noticed this has some LLM assisted changes, we have no precedent for adding such changes yet, and I'm not sure that they are acceptable due to unclear copyright status.

PadreAdamo:
> We don't have the capacity to review new drivers - #1206. I noticed this has some LLM assisted changes, we have no precedent for adding such changes yet, and I'm not sure that they are acceptable due to unclear copyright status.

Thanks for taking the time to look at this and for the honest feedback; no hard feelings at all. I know bandwidth is tight right now. Just trying to contribute in whatever way I know how. Appreciate you.

PR #1794, opened 2026-01-01
Feature: Add support for ASUS Omni wireless receiver device

haytham818:
## Description

This PR adds support for the ASUS Omni wireless receiver (VID:PID 0b05:1ace), 
a single USB dongle that supports multiple wireless mice models. The challenge 
is that all mice share the same USB identifier, making device identification 
impossible using VID:PID alone.

The solution implements dynamic device identification by querying a unique 
12-byte signature from each connected mouse during driver probe.

## Problem Statement

- ASUS Omni receiver creates a single USB device (0b05:1ace) for all mice
- Different mouse models need different driver configurations
- Without proper identification, all mice appear as generic "ASUS ROG OMNI RECEIVER"
- Users cannot configure device-specific features

## Solution

1. **Device Identification**: Query device signature via HID command (0x03 0x12 0x12 0x02)
2. **Signature Matching**: Map 12-byte signature to 10 known mouse models
3. **Dynamic Naming**: Replace generic name with actual device model during probe
4. **Unified Configuration**: Use generic config file with maximum values for all Omni mice

### New Quirk Flag
- `ASUS_QUIRK_OMNI_RECEIVER`: Special handling for Omni receiver protocol

### HID Protocol Handling
- Output reports: Prepend report ID 0x03 to command
- Input reports: Skip first byte (report ID) from response
- Signature location: Offset 5, 12 bytes (ASCII string)

### Supported Devices (10 models)
- ASUS ROG Harpe Ace Aim Lab Edition (B23*, B241*, B2501*) 
- ASUS ROG Harpe Ace Mini (B24*, B25*)
- ASUS ROG Harpe Ace Extreme (T5*)
- ASUS ROG Keris Wireless Aimpoint (R1*, F24*, FB*, R9*)
- ASUS ROG Keris Ace II (024*, 02501*)
- ASUS ROG Keris II Origin (025*) 
- ASUS ROG Strix Impact III Wireless (20*) 
- ASUS ROG Gladius III Aimpoint (R6*, R8*, RC*, R903*, R923*)

### Code Changes
- `src/asus.h`: Add ASUS_QUIRK_OMNI_RECEIVER flag and function declaration
- `src/asus.c`: 
  - Modify `asus_query()` for Omni HID protocol
  - Implement `asus_omni_identify_mouse()` for signature matching
- `src/driver-asus.c`: Call identification in probe, update device name
- `src/libratbag-data.c`: Add quirk parsing

### Configuration
[...]

PR #1880, opened 2026-08-12
asus: add support for ASUS ROG Strix Impact III and the Omni receiver

haytham818:
# Support for ASUS ROG Strix Impact III and the Omni receiver

Adds the wired ROG Strix Impact III (`usb:0b05:1a88`) and the wireless ROG Strix Impact III Wireless, which connects through the ASUS Omni receiver (`usb:0b05:1ace`).

## The Omni receiver problem

Every mouse behind an Omni receiver enumerates with the same usb id (`0b05:1ace`), so DeviceMatch cannot tell models apart. PR #1794 worked around this with one generic configuration and renamed the device during probe. That broke real hardware: the generic file exposed 3 LEDs to the Harpe Ace, which has fewer, and setting one of them turned the LED off.

This PR follows the approach kyokenn suggested in #1794. The receiver gets a minimal bootstrap file that only declares `Quirks=OMNI_RECEIVER`. During probe the driver queries the serial number signature (`03 12 12 02`) and maps its prefix to the model usb id through a small table. For the Impact III Wireless the prefix is `20` (my unit returns `202407310031`) and the id is `0b05:1ad7`. The probe then re-resolves the device data with `ratbag_device_data_new_for_id()`, so each model ends up with its own .device file and its own button, LED and DPI counts. An unknown signature fails the probe with `-ENODEV`, which also drops the receiver's keyboard and media control hidraw nodes.

## New quirks

- `OMNI_RECEIVER`: transport with report ID 0x03 and 64 byte reports
- `LED_V2`: per-zone LED responses with data at offset 4, brightness in 0-100
- `SETTINGS_V2`: settings and DPI responses with data at offset 4, rate/debounce/snapping as single bytes

## Testing

Verified on a ROG Strix Impact III Wireless:

- the probe identifies the mouse on hidraw4 and rejects the other receiver nodes
- all 5 profiles load with 8 buttons, 1 LED and DPI presets 400/800/1600/3200
- report rate (1000 Hz), debounce (12 ms) and LED brightness (127, i.e. 50%) parse correctly
- DPI, report rate, LED brightness and LED color writes round-trip and survive a ratbagd restart
- `meson test`: 7 OK, 0 failed

## Limitations

The wired Impact III device file (Leds=2, DpiRange=100:12000@50) follows the official specs and the same command set as the wireless model. I have no wired unit to test it with. The wired and wireless models share the command set in g-helper, so the risk is low.

Other Omni models (Harpe Ace, Keris, Gladius) are not part of this PR. The prefix table in `asus_omni_identify()` is the only place that needs a new row, and each model then gets its own data-only commit.

DPI colors, battery, lift-off distance and angle adjustment exist on the device but sit outside the libratbag capability model, so they are not exposed.

### DIFF
--- a/src/asus.c
+++ b/src/asus.c
@@ -26,6 +26,7 @@
 #include "asus.h"
 
 #include <assert.h>
+#include <string.h>
 
 #include "libratbag-data.h"
 #include "libratbag-private.h"
@@ -138,15 +139,37 @@ asus_query(struct ratbag_device *device,
 		union asus_request *request, union asus_response *response)
 {
 	int rc;
+	uint32_t quirks = ratbag_device_data_asus_get_quirks(device->data);
+	uint8_t omni_tx_buf[ASUS_PACKET_SIZE];
+	uint8_t omni_rx_buf[ASUS_PACKET_SIZE];
 
-	rc = ratbag_hidraw_output_report(device, request->raw, ASUS_PACKET_SIZE);
-	if (rc < 0)
-		return rc;
+	if (quirks & ASUS_QUIRK_OMNI_RECEIVER) {
+		/* devices behind the Omni receiver use report ID 0x03 and a
+		 * 64-byte report, the payload is shifted by one byte */
+		memset(omni_tx_buf, 0, sizeof(omni_tx_buf));
+		omni_tx_buf[0] = 0x03;
+		memcpy(omni_tx_buf + 1, request->raw, ASUS_PACKET_SIZE - 1);
 
-	memset(response, 0, sizeof(union asus_response));
-	rc = ratbag_hidraw_read_input_report(device, response->raw, ASUS_PACKET_SIZE, NULL);
-	if (rc < 0)
-		return rc;
+		rc = ratbag_hidraw_output_report(device, omni_tx_buf, ASUS_PACKET_SIZE);
+		if (rc < 0)
+			return rc;
+
+		memset(response, 0, sizeof(union asus_response));
+		rc = ratbag_hidraw_read_input_report(device, omni_rx_buf, ASUS_PACKET_SIZE, NULL);
+		if (rc < 0)
+			return rc;
+
+		memcpy(response->raw, omni_rx_buf + 1, ASUS_PACKET_SIZE - 1);
+	} else {
+		rc = ratbag_hidraw_output_report(device, request->raw, ASUS_PACKET_SIZE);
+		if (rc < 0)
+			return rc;
+
+		memset(response, 0, sizeof(union asus_response));
+		rc = ratbag_hidraw_read_input_report(device, response->raw, ASUS_PACKET_SIZE, NULL);
+		if (rc < 0)
+			return rc;
+	}
 
 	/* invalid state, disconnected or sleeping */
 	if (response->data.code == ASUS_STATUS_ERROR) {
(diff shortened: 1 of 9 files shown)


haytham818:
@kyokenn would you appreciate a review ?

src/asus.c
```diff
@@ -511,3 +563,61 @@ asus_set_led(struct ratbag_device *device,
...
+	signature[12] = '\0';
+
+	/* model table, keyed by the serial number prefix. Keep prefixes
+	 * ordered from the longest to the shortest, so that a shorter
+	 * prefix never shadows a longer one. */
+	static const struct asus_omni_model {
+		const char *prefix;
+		uint16_t product_id;
+	} omni_models[] = {
+		{ "20", 0x1ad7 },  /* ROG Strix Impact III Wireless */
```
kyokenn: If it's possible it would be good to be moved into ".device" configuration file. So other people don't have to rebuild ratbag and can just add ".device" file for their mouse. But I don't mind merging it as is.

kyokenn:
I don't have Omni receiver, so I can't test it, but it looks great to me!
Does the Strix Impact III Wireless still works with this PR in wired mode without Omni receiver connected?

staticssleever668:
Nit: what's in the PR description should have been in the commit message instead

haytham818:
@kyokenn The Impact III Wireless has no wired mode. It has no USB port, it only connects through the Omni receiver or Bluetooth.

The `usb:0b05:1a88` device file in this PR is for the wired Impact III, which is a separate model with its own cable. I could not test that one, since I only own the wireless model.

haytham818:
@staticssleever668 Done, the description now lives in the commit messages.

haytham818:
@kyokenn On the OmniSignature suggestion: done. The prefix is now in the OmniSignature field of the .device file, asus_omni_identify() matches the serial against it, and a new model only needs a data commit.

kyokenn:
> The Impact III Wireless has no wired mode. It has no USB port, it only connects through the Omni receiver or Bluetooth.

Ah, I see. I never saw the mouse with batteries from ASUS.

> On the OmniSignature suggestion: done. The prefix is now in the OmniSignature field of the .device file, asus_omni_identify() matches the serial against it, and a new model only needs a data commit.

Awesome, thank you!

staticssleever668:
What does it mean for `ASUS ROG Omni Receiver` to have a separate device file? `usb:0b05:1ad7`, for example, is it usually not an actual USB device, that is, is it not visible in dmesg or lsusb?

haytham818:
@staticssleever668 Right, `0b05:1ad7` never shows up in lsusb or dmesg. The host only ever sees the receiver (0b05:1ace). The mouse's own PID comes from the receiver's pairing table command (`01 A0 00 00`).

The DeviceMatch on the wireless file works as a lookup key only. It never matches during udev enumeration. After the probe identifies the model, it calls ratbag_device_data_new_for_id() with the resolved id to load the model file. That is the flow kyokenn suggested in #1794: resolve the usb id behind the Omni receiver, then let the normal .device matching do its thing.

The bootstrap file is the one that matches the receiver itself.

dms_capture_1786643919348

kyokenn (approved):

staticssleever668 (approved):
I glanced over the changes and they seem good. It's an interesting way to implement receiver support. @kyokenn thank you for the review too!

Issue #1797, opened 2026-01-06
Support for Razer Naga Left-Handed Edition (1532:008d)


tmschorr:
Hello,

I would like to request support for the Razer Naga Left-Handed Edition mouse under Linux via libratbag.

Details:

- VendorID:ProductID: 1532:008d
- Model: Razer Naga Left-Handed Edition
- Type: Gaming mouse, left-handed with 12-button thumb grid
- Current behavior: ratbagctl does not detect this mouse. 'ratbagctl list' returns "No devices available".

Steps I took:

1. Built the latest libratbag from source.
2. Installed and started ratbagd from /usr/local/sbin/ratbagd.
3. Verified that ratbagd is running and ratbagctl works.
4. The mouse is connected and visible in lsusb: `Bus 003 Device 002: ID 1532:008d Razer USA, Ltd Razer Naga Left-Handed Edition`

Expected behavior:

- libratbag should recognize the mouse and allow configuration of buttons, DPI settings, and LED profiles.

Additional information:

- This is the left-handed version, which seems to have a different USB ID from the standard Naga.
- I would be happy to provide logs or testing if needed.

Thank you for considering adding support for this mouse!

Issue #1690, opened 2025-03-06
Support Razer Basilisk V3


cth-odoo:
**Is your feature request related to a problem? Please describe.**
I own a Razer Basilisk V3 (Hyperspeed 35k to be accurate) and this device is currently not supported by Piper.

**Describe the solution you'd like**
Ability to remap the buttons in Piper.

**Describe alternatives you've considered**
I did not find any easy way to remap the buttons on Linux Mint. The workaround I currently use is setting up the mouse on Windows and relying on the configuration that can be stored in the embedded memory of the mouse. This is not very convenient when I want to adjust my setup.

**Additional context**
Thanks for your great work 😃

ahouse73:
New here, but found this in the [Wiki](https://github.com/libratbag/libratbag/wiki/Devices#razer-devices):

> The EULA of the Razer configuration tool prevents us to reverse engineer their protocol. Until we have legal access (or maybe a 
> clean room implementation), we can not support those devices in libratbag.

So probably will not be added soon

Issue #103, opened 2016-10-13
Support for Razer mice


mbuesch:
libratbag should be taught to talk to Razer mice.

Razer uses proprietaty USB messages to perform device configuration.
An implementation of some Razer mouse protocols can be found in razercfg.
https://bues.ch/h/razercfg#device_support
https://github.com/mbuesch/razer

Lyude:
See the bottom of https://libratbag.github.io/device-notes/
Even though you have a functioning implementation, unfortunately (especially since pretty much all of the libratbag devs are from different countries) we can't use any sort of implementation that made use of reverse engineering their windows drivers. Sorry :(

mbuesch:
Yes, I know about that statement and I don't think it is valid.

First of all reverse engineering for the purpose of writing a driver is not illegal in some countries, even if the EULA states that (pretty much every EULA out there does that voodoo statement).
And I don't see how this affects you anyway, if you use the razercfg code as a reference. You will not be doing reverse engineering, if you read the GPLed code of razercfg. Yes, you cannot take the GPLed code and use it under MIT license. I can't re-license it for various reasons either. _But_ you can read the code and use it as a documentation of the hardware protocol to write a new driver.

And last but not least: Razer is _not_ hostile against Open Source projects. In fact they actively supported me with the development of the first Deathadder driver under GPL license.

Lyude:
Thank you a ton for letting us know this. The main reason we're so paranoid about it is because a lot of the work on libratbag is done at Red Hat we have to tread a lot more carefully with legal issues.
This being said however: if Razer's actually worked with you on this personally that changes things a ton. I definitely can't guarantee that this issue won't get reclosed since we need to do some research about this, but for now I'm going to reopen it to leave the discussion open. Really the legal issues have been the only reason we haven't had razer support up until now, and I would love for us to be able to incorporate support for their devices.

bentiss:
As mentioned in issue #172, I think it'll be quicker to just connect to a different dbus daemon that does the communications with the hardware for us.

I know this bug has been opened taking into account razercfg, but would any of you be strongly against using the daemon from razer-drivers which seems to support far more devices?

mbuesch:
>but would any of you be strongly against using the daemon from razer-drivers which seems to support far more devices?

Of course not.
I just wanted to point out the licensing / reverse engineering situation with Razer and my personal experience with it.

whot:
[...]

PR #1888, opened 2026-08-26
razer: add a driver for Razer mice, starting with the DeathAdder V3

knarxy:
Adds a `razer` driver to libratbag and the first device for it, the wired
DeathAdder V3 (`1532:00b2`). A first step towards #103, and a base the other
Razer requests (#1690, #1797) can build on.

Scope: DPI stages and report rate are configurable, buttons are reported but
not remapped, no LEDs. Everything below was verified against the device
(firmware 1.2) rather than taken from OpenRazer on faith.

## Commits

| | |
|---|---|
| `libratbag-data` | the `[Driver/razer]` group, its accessors, `device.example`, `data-parse-test.py` |
| `razer` | the driver, registration, `meson.build` |
| `data` | `razer-deathadder-v3.device` |
| `README` | Razer in the vendor list |

Each one builds on its own. `meson test` is green (7 ok, 0 fail, 3 skipped
for the usual missing bits), `black`/`ruff` at the versions pinned in
`main.yml` pass on the test change.

### 1. The interface check does I/O, which no other driver does

`razer_test_hidraw()` filters on the descriptor and *then* sends a firmware
request to confirm. Every other driver uses a pure descriptor predicate, so
this deserves an explanation.

The control protocol is a 90 byte feature report on report id 0, declared in
the vendor usage page 0xff00 — but that page sits inside the Generic
Desktop/Mouse application collection:

```
05 01 09 02 a1 01                  Generic Desktop, Mouse, Application
  [...]
  06 00 ff 09 02                   usage page 0xff00, usage 0x02
  15 00 25 01 75 08 95 5a b1 01    feature, 0x5a = 90 bytes
c0
```

`ratbag_hidraw_has_vendor_page()` never sees it. The parser records the usage
page of the first application collection per report id only
(`libratbag-hidraw.c`, `case HID_COLLECTION`), so this interface is filed
under Generic Desktop. Filtering on the vendor page binds nothing at all.

`ratbag_hidraw_has_report(device, 0)` does not separate them either: the
device exposes three interfaces, and two of them are keyboard collections,
one of which also has no report id. What does separate them is the
application usage — mouse (0x02) against keyboard (0x06).

The descriptor check has to come first for a practical reason: a request to
the keyboard collections stalls the control transfer with `-EPROTO`, the
device resets, udev re-adds it and ratbagd probes again, endlessly. That is
what the confirming request is guarded by, and it is why it is a guard rather
than the whole test.

If you would rather have a pure descriptor predicate, dropping the request is
a two-line change and I am happy to do it — a wrong interface would then fail
later in `razer_read_settings()` instead.

The parser limitation itself looks worth fixing separately, so that
`has_vendor_page()` works for descriptors of this shape. I have left it alone
here.

### 2. Two deviations from OpenRazer

Both confirmed by writing and reading back:

- **DPI stage ids are 1-based.** The device reports its stages as id 1..5 and
  takes them back that way. OpenRazer writes 0-based ids when setting stages.
- **The last two bytes of a stage record are not reserved.** They hold a third
  copy of the stage's DPI. OpenRazer zeroes them; this driver preserves what
  the device reported for stages it does not touch.

### 3. No lighting, on purpose

The DeathAdder V3 has no addressable zone, only a DPI indicator LED with no
documented protocol. OpenRazer agrees — in `razer_mouse_probe()`:

```c
case USB_DEVICE_ID_RAZER_DEATHADDER_V3:          /* 0x00B2 */
    CREATE_DEVICE_FILE(&hdev->dev, &dev_attr_poll_rate);
    CREATE_DEVICE_FILE(&hdev->dev, &dev_attr_dpi);
    CREATE_DEVICE_FILE(&hdev->dev, &dev_attr_dpi_stages);
    break;
```

Three attributes, none of them lighting, while neighbouring models in the
same switch get `scroll_led_brightness` and the `matrix_effect_*` family.
None of the 14 occurrences of that constant is in a LED or brightness
function.

So the driver reports zero LEDs. Razer models that do have zones need a
device data key and a led count in `ratbag_device_init_profiles()` — no
restructuring.

## Testing

On the device, through `ratbagctl` against a `ratbagd` built from this
branch: reading and setting each of the five stages, separate X/Y
resolutions, switching the active stage, each of the seven report rates from
125 to 8000 Hz, and persistence across a daemon restart. Also confirmed that
the device is exported exactly once per daemon run, i.e. no re-enumeration
loop.

Piper drives it as-is; it has no drawing for this model yet.

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,7 +9,7 @@ features exposed by these mice and abstracts away hardware-specific and
 kernel-specific quirks.
 
 libratbag currently supports devices from Logitech, Etekcity, GSkill,
-Roccat, Steelseries. See [the device
+Razer, Roccat, Steelseries. See [the device
 files](https://github.com/libratbag/libratbag/tree/master/data/devices) for
 a complete list of supported devices.
 
--- a/data/devices/device.example
+++ b/data/devices/device.example
@@ -77,6 +77,27 @@ DpiRange=100:16000@100
 # Device quirks
 # Quirk=DOUBLE_DPI;STRIX_PROFILE
 
+[Driver/razer]
+# Number of buttons. Reported so they show up in the UI, this driver does
+# not remap them.
+Buttons=6
+
+# The number of DPI stages the device stores, 1 to 5.
+DpiStages=5
+
+# DPI range in the form min:max@step
+# Mutually exclusive with DpiList
+DpiRange=100:30000@100
+
+# The list of available resolutions, separated by semicolons.
+# Mutually exclusive with DpiRange
+# DpiList=400;800;1600;3200
+
+# The transaction id groups request and response and is model specific:
+# 31 (0x1f) for the DeathAdder V3 generation, 255 (0xff) for older Razer
+# mice. Optional, defaults to 31.
+# TransactionId=31
+
 # Replace `FEED` in the group name with firmware version of the device.
 # To find the right value, see the output of ratbagd and search a message like:
 # "ratbag error: Device with firmware version FEED is not supported"
--- a/data/devices/razer-deathadder-v3.device
+++ b/data/devices/razer-deathadder-v3.device
@@ -1,0 +1,13 @@
+[Device]
+DeviceMatch=usb:1532:00b2
+DeviceType=mouse
+Driver=razer
+Name=Razer DeathAdder V3
+
+[Driver/razer]
+# reported so they show up in the UI; this driver does not remap them
+Buttons=6
+DpiRange=100:30000@100
+DpiStages=5
+# 31 == 0x1f, verified against the device (firmware v1.2); older Razer mice use 0xff
+TransactionId=31
--- a/meson.build
+++ b/meson.build
@@ -176,6 +176,7 @@ src_libratbag = [
 	'src/driver-hidpp10.c',
 	'src/driver-logitech-g300.c',
 	'src/driver-logitech-g600.c',
+	'src/driver-razer.c',
 	'src/driver-roccat.c',
 	'src/driver-roccat-kone-emp.c',
 	'src/driver-roccat-kone-pure.c',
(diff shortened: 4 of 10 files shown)


knarxy:
One thing I want to put on the table myself rather than leave to be asked: where
the protocol knowledge in this driver comes from. The wiki's "Razer devices"
section says the EULA of Razer's configuration tool is what blocks Razer support
in libratbag, and this PR should not sit here as if that had not been written.

**What was used**

- OpenRazer's published source, as a reference for the report layout and the
  command ids.
- The device itself: requests sent to a DeathAdder V3 I own and the responses
  read back. The probe script I used is read-only; I am happy to attach its
  captures for any claim in this PR.

**What was not used**

Razer's configuration software played no part in this work. It was not
installed, run, decompiled, or traffic-captured — the machine this was
developed on runs Fedora.

**The part I cannot wave away**

OpenRazer is GPL-2.0-or-later, libratbag is MIT, and this PR names OpenRazer as
a reference. So, concretely: no code was copied. The driver is written against
libratbag's own API and in its idiom. What OpenRazer contributed is factual
protocol information — report length, the checksum, the command ids — which I
then verified against the hardware.

That verification is also where it stops agreeing. Two details in this driver
contradict OpenRazer, both because the device said so:

- DPI stage ids are 1-based, not 0-based.
- The last two bytes of a stage record are not reserved; they hold a third copy
  of the stage's DPI.

Those came out of my own captures, not out of reading someone else's code, and
they are why I am confident the rest was verified rather than transcribed.

I am not going to argue law with you, and I am not asking you to take on a risk
you would rather not. Two questions instead:

1. Does the wiki's Razer position still reflect the project's view? It goes back
   to the 2016 discussion in #103, and the project has changed hands since.
2. If the GPL-referenced-in-an-MIT-tree part is the sticking point, is there a
   shape this could take that would be acceptable to you — or is it simply out?

Either answer is fine and I would rather have it early than late. Happy to go
into any protocol detail, and to attach the raw captures behind every number in
this PR.

Issue #1851, opened 2026-05-27
[Feature request] Add support for Fury / Natec Fury Tanto T4 Wireless


kbkozlev:
**Is your feature request related to a problem? Please describe.**

Current `libratbag` does not support the **Fury / Natec Fury Tanto T4 Wireless** mouse.

The mouse is detected by Linux and works as a regular HID mouse, but it is not detected by `libratbag` / `ratbagd`, so it cannot be configured through Piper or `ratbagctl`.

Current output:

```text
$ ratbagctl list
No devices available.
```

**Describe the solution you'd like**

Add support for the **Fury / Natec Fury Tanto T4 Wireless** mouse.

**Describe alternatives you've considered**

At the moment, I do not have access to a Windows machine, so I cannot use the official vendor software or capture/sniff the configuration protocol from it.

The mouse works as a basic HID mouse on Linux, but DPI stages, polling rate, profiles, button mappings, and onboard settings cannot currently be configured through `libratbag`.

**Hardware attributes**

```text
Device Name: Fury Tanto T4 Wireless USB
Also known as: Natec Fury Tanto T4 Wireless
Vendor ID: 0x1d57 (Xenta / Compx)
Product ID: 0x3005
USB Vendor: Xenta
USB Product: Tanto T4 Wireless USB
Manufacturer String: Fury
Product String: Tanto T4 Wireless USB
Sensor: PixArt PAW3311
Max DPI: 12,000 DPI
Polling Rate: up to 1000 Hz
Connection: 2.4 GHz wireless receiver / USB wired / Bluetooth
Buttons: 7
Programmable buttons: 6
Built-in memory: Yes
RGB: No
Driver Candidate: Unknown, possibly similar to other 0x1d57 / Xenta / Compx PAW3311 wireless mice
```

**Additional context**
Linux detects the mouse as:

```text
Bus 001 Device 005: ID 1d57:3005 Xenta Tanto T4 Wireless USB
```

`lsusb -v -d 1d57:3005` output:

```text
Bus 001 Device 005: ID 1d57:3005 Xenta Tanto T4 Wireless USB
Couldn't open device, some information will be missing
Negotiated speed: Full Speed (12Mbps)

Device Descriptor:
  bcdUSB               1.10
  bDeviceClass            0 [unknown]
  bDeviceSubClass         0 [unknown]
  bDeviceProtocol         0
  bMaxPacketSize0        64
  idVendor           0x1d57 Xenta
[...]

PR #1897, opened 2026-09-11
Add support for Fury Tanto T4 Wireless

kbkozlev:
## Summary

Add a new hardware backend for the Fury Tanto T4 Wireless mouse.

The protocol was reverse-engineered using USB captures from the official application while changing individual settings.

## Hardware information

* Optical sensor: PixArt PAW3311
* Wired USB ID: `1d57:3005`
* 2.4 GHz receiver USB ID: `1d57:3006`

## Supported functionality

* Wired mode (`1d57:3005`)
* 2.4 GHz receiver (`1d57:3006`)
* One write-only profile
* Seven DPI stages
* DPI range from 200 to 12000
* Active DPI-stage selection
* Polling rates of 125, 250, 500 and 1000 Hz
* Six configurable buttons
* DPI-cycle button action

## Hardware testing

Both wired and 2.4 GHz modes were tested on physical hardware.

Verified operations:

* Device detection and initialization
* Switching between DPI stages
* Setting DPI to 800, 1600 and 3200
* Changing the polling rate between 500 and 1000 Hz
* Remapping and restoring the side buttons
* Reconnecting and initializing the 2.4 GHz receiver from a cold start

## Limitations

The device does not return its current configuration through `GET_FEATURE`, which times out. The driver therefore exposes captured defaults and marks the profile as write-only.

Changes made using the physical DPI button are applied by the mouse but cannot currently be reflected in `ratbagctl` or Piper.

## Disclosure

For transparency, I used AI-assisted tools while reverse-engineering and implementing this driver. All USB captures, hardware tests, and verification were performed by me.


kbkozlev:
closes #1851

